PR #2, opened 2022-02-07
Added rspec to the project

seanstory:
Closes https://github.com/elastic/enterprise-search-team/issues/1305

Adds the ability to run rspec tests along with jvm tests through maven


.ci/script/build.sh
```diff
@@ -28,6 +28,16 @@ rbenv versions
 rbenv global $JRUBY_VERSION
 ruby --version # just to double-check...
 which -a ruby
+ruby_bin_file=$(which ruby)
+export BUNDLE_CMD="$ruby_bin_file -S bundle"
```
seanstory, line 32: @tarekziade is there a better way to invoke this logic from https://github.com/elastic/ent-search-ci-build/blob/main/lib/entsearchci/generator/ruby.rb without having to use `perform-ci-build`?

Gemfile
```diff
@@ -0,0 +1,13 @@
+
+# Pull gem index from rubygems
+source 'https://rubygems.org'
+
+# Pin the version of bundle we support
+gem 'bundler', '2.2.29'
+
+group :test do
+  gem 'rspec-core', '~> 3.10.1'
+  gem 'rspec-rails', '~> 4.0.2'
```
orhantoy, line 10: Do we need this gem or could RSpec core suffice?

seanstory: good calll, we don't need `rspec-rails`. The `rspec-collections-matchers` I leave though, as that gives us the `expect()` function, and the `rspec_junit_formatter` should allow jenkins to give us a pretty breakdown of our individual test outputs.

seanstory:
Woo! Rspec tests running in CI, and their output is aggregated with other Junit/maven test output!

Screen Shot 2022-02-08 at 3 41 16 PM

hello-world-connector/src/test/ruby/example_test.rb
```diff
@@ -0,0 +1,5 @@
+RSpec.describe 'example' do
+    it "can run a simple test" do
+      expect(1).to eq(1)
+    end
```
orhantoy, line 4: Indentation is first 4 and then 2 at the next level; both should probably be 2.

hello-world-connector/src/test/ruby/example_test.rb
```diff
@@ -0,0 +1,5 @@
...
+end
```
orhantoy, line 5: Final newline is also missing here.

Gemfile
```diff
@@ -0,0 +1,12 @@
...
+source 'https://rubygems.org'
+
+# Pin the version of bundle we support
+gem 'bundler', '2.2.29'
+
+group :test do
+  gem 'rspec-core', '~> 3.10.1'
+  gem 'rspec-collection_matchers', '~> 1.2.0'
+  gem 'rspec_junit_formatter'
+end
```
orhantoy, line 12: Final newline is also missing here. VSCode and probably other editors insert this by default.

Gemfile
```diff
@@ -0,0 +1,12 @@
+
```
orhantoy, line 1: Let's remove this empty line.

orhantoy (commented):
Looks good; just found some minor stylistic things.

orhantoy (approved):
🚀

PR #3, opened 2022-02-09
move connectors exceptions

seanstory:
# part of https://github.com/elastic/enterprise-search-team/issues/1293

This adds the main connectors exception classes to our connectors-api module


artem-shelkovnikov (approved):
LGTM! Java annotations in ruby files are looking really interesting, makes me think about how the JRuby works internally with those :)

seanstory:
> LGTM! Java annotations in ruby files are looking really interesting, makes me think about how the JRuby works internally with those :)

Fascinating, right? You can read all about it here: https://github.com/jruby/jruby/wiki/GeneratingJavaClasses

PR #5, opened 2022-02-11
boilerplate code for the http service

tarekziade:


artem-shelkovnikov:
Adding a simple HTTP server serving a HelloWorld connector.

To run, use the makefile command:
```shell
make .phony
```

A lot of stuff is in flight, but hey it can be used to actually ingest a bit of stuff to Enterprise Search.

PR #6, opened 2022-02-15
Add ConnectorNotAvailable Error - used by External Connector Framework

artem-shelkovnikov:
Issue for discussion.

The error added in this PR is the error that is specific to the Enterprise Search handling of connectors - it does not get thrown by any connector.

However, `connectors_errors` file contains some errors that are framework specific, as I saw.

Question: should these errors (such as `JobClaimingError`, `JobCannotBeUpdatedError`) be moved back to Enterprise Search along with ConnectorNotAvailableError?

### DIFF
--- a/connectors-api/src/main/ruby/connectors_errors.rb
+++ b/connectors-api/src/main/ruby/connectors_errors.rb
@@ -92,4 +92,4 @@ class JobCannotBeUpdatedError < StandardError; end
 class SecretInvalidError < StandardError; end
 class InvalidIndexingConfigurationError < StandardError; end
 class TokenRefreshFailedError < StandardError; end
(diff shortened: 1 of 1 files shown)


artem-shelkovnikov:
@seanstory thank you for the approval!

What do you think about the part about moving some of error classes back to Enterprise Search?

seanstory:
Yeah, you're probably right. I didn't apply a lot of critical thought, and have just been moving all the errors that we had in that file. I don't think it necessarily hurts anything to have them in the public repo, so I'm not sure I'd advocate for prioritizing moving them back. I could go either way.

artem-shelkovnikov:
I will go ahead and merge this PR here and prepare a separate PR to move some exception classes back to the ent-search repo.

PR #7, opened 2022-02-15
move connectors exceptionhandling

seanstory:
## part of https://github.com/elastic/enterprise-search-team/issues/1294

- moves Connectors::ConnectorsAPI::ExceptionTracking to ConnectorsExceptiontracking
- ensures that multiple layers of dependencies resolve correctly

### DIFF
--- a/connectors-api/src/main/ruby/connectors_exception_tracking.rb
+++ b/connectors-api/src/main/ruby/connectors_exception_tracking.rb
@@ -1,0 +1,41 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+
+# frozen_string_literal: true
+
+require 'swiftype_exception_tracking' unless defined?(Rails)
+require 'bson'
+require_relative './connectors_logger'
+
+java_package 'co.elastic.connectors.api'
+
+class ConnectorsExceptionTracking
+  class << self
+    def capture_message(message, context = {})
+      Swiftype::ExceptionTracking.capture_message(message, context)
+    end
+
+    def capture_exception(exception, context = {})
+      Swiftype::ExceptionTracking.log_exception(exception, :context => context)
+    end
+
+    def log_exception(exception, message = nil)
+      Swiftype::ExceptionTracking.log_exception(exception, message, :logger => ConnectorsLogger.logger)
+    end
+
+    def augment_exception(exception)
+      unless exception.respond_to?(:id)
+        exception.instance_eval do
+          def id
+            @error_id ||= BSON::ObjectId.new.to_s
+          end
+        end
+      end
+    end
+  end
+end
+
--- a/connectors-stubs/src/main/ruby/swiftype_exception_tracking.rb
+++ b/connectors-stubs/src/main/ruby/swiftype_exception_tracking.rb
@@ -1,0 +1,35 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+module Swiftype
+  # empty
+end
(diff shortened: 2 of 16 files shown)


connectors-stubs/src/main/ruby/swiftype_exception_tracking.rb
```diff
@@ -0,0 +1,35 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+module Swiftype
```
artem-shelkovnikov, line 6: Hmmm should we also have `Swiftype` anywhere here?

I'd probably rename it to something else, like `Util::ExceptionTracking`

connectors-api/src/main/ruby/connectors_exception_tracking.rb
```diff
@@ -0,0 +1,41 @@
...
+
+# frozen_string_literal: true
+
+require 'swiftype_exception_tracking' unless defined?(Rails)
+require 'bson'
+require_relative './connectors_logger'
+
+java_package 'co.elastic.connectors.api'
+
+class ConnectorsExceptionTracking
```
artem-shelkovnikov, line 16: Should this whole class be inlined to `Swiftype::ExceptionTracking`?

It seems like this class is only wrapping the `Swiftype::ExceptionTracking` class, while `Swiftype::ExceptionTracking` seem like it's not meant to be used directly.

example-consumer/src/test/java/co/elastic/connectors/example/consumer/ExampleConsumerTest.groovy
```diff
@@ -33,11 +33,11 @@ class ExampleConsumerTest extends Specification {
         then:
         1==1
 
-        when: "active_support"
-        runtime.executeScript("require 'active_support'", 'foo.rb') // not transitively included
+        when: "nokogiri"
+        runtime.executeScript("require 'nokogiri'", 'foo.rb') // not transitively included
```
artem-shelkovnikov, line 37: What is the `foo.rb` file?

artem-shelkovnikov:
Looks good overall, left a couple conceptual questions

connectors-api/src/main/ruby/connectors_exception_tracking.rb
seanstory: Yep, that's exactly the case. We don't want to move the _real_ Swiftype::ExceptionTracking into this repo though. The only reason it's included in `connectors-stubs` is so that this code can still function in ent-search (without `connectors-stubs`) and also run in the MVP for the external framework (with `connectors-stubs`). In the long run, we should do some refactoring and cleaning here, and replace `connectors-stubs` with more fully-fleged implementations. But right now, as we cut scope, those classes can practically no-op as long as they don't cause runtime errors.

connectors-stubs/src/main/ruby/swiftype_exception_tracking.rb
seanstory: I would too. Unfortunately, then the `connectors_exception_tracking.rb` would not function when dropped back into ent-search.

seanstory: For context, here's where this idea came from: https://github.com/elastic/ent-search/pull/6157#issuecomment-1034138558

artem-shelkovnikov: Can we rename it in `ent-search` as well? I think the Swiftype namespace is shrinking and it's a good opportunity to shrink it a bit more.

seanstory: I think that namespacing is somewhat tied to directory structure in ruby? And there's a lot still in `lib/swiftype`. I'm not opposed to it as a general goal, but I don't think we're going to get a greenlight to prioritize that.
Screen Shot 2022-02-17 at 9 51 53 AM

artem-shelkovnikov (approved):

PR #9, opened 2022-02-17
Moved connector constants

seanstory:
not a very exciting PR


connectors-api/src/main/ruby/connectors_constants.rb
```diff
@@ -0,0 +1,15 @@
...
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+require 'java'
+
+java_package 'co.elastic.connectors.api'
+class ConnectorsConstants
+  THUMBNAIL_FIELDS = %w[_thumbnail_80x100 _thumbnail_310x430]
+  SUBEXTRACTOR_RESERVED_FIELDS = %w[_subextracted_as_of _subextracted_version]
```
artem-shelkovnikov, line 12: Hmmm do we need these two here? They seem to be either derivative or somehow very internal.

artem-shelkovnikov (approved):
Not a very exciting approval with a question.

connectors-api/src/main/ruby/connectors_constants.rb
seanstory: it's used in our base/adapter.rb: https://github.com/elastic/ent-search/pull/6188/files#diff-d050fc8f8d7e12f7eef1625836d5fdfd105472bcdbace18667f336e0480264b0R11

We'll eventually want to publish/document all of these fields, so that a customer can leverage them externally. Like if they wanted to do subextraction themselves, they could set these fields and then get clues later about how fresh that subextraction was.

PR #61, opened 2022-03-29
Now running on Windows

tarekziade:
Add support for Windows


artem-shelkovnikov (approved):
LGTM!

A question about the logic - so if I'm on Windows machine, running `install.bat` and `make.bat` will prepare me for connector development by installing `make`, all dependencies for connectors (ruby, rbenv, etc), so that after that I can use windows console to run our make commands, is it correct?

tarekziade:
> LGTM!
> 
> A question about the logic - so if I'm on Windows machine, running `install.bat` and `make.bat` will prepare me for connector development by installing `make`, all dependencies for connectors (ruby, rbenv, etc), so that after that I can use windows console to run our make commands, is it correct?

It's not using the Makefile right now, and `make.bat` just runs the spec using `bundler`

`install.bat` installs `MSYS2`, `MinGW` and `rbenv-win` which uses `RubyInstaller2`

I have made opinionated choices there, to make it easy to create an unattended deployment in our Jenkins Windows worker. I expect real Windows developers out there to use Chocolatey or other tools, but with this patch we offer a simple, embed way to run the project and its tests natively under Windows 10 by reusing how it's done in automation

PR #62, opened 2022-03-29
Test connectors with ent search - Part 2

mchernyavskaya:
## Part 2 of https://github.com/elastic/enterprise-search-team/issues/1504

We want an automated way to ensure that changes to https://github.com/elastic/connectors do not break https://github.com/elastic/ent-search. So instead of manually copying over the gems, we're adding a github link to a specific branch/tag. This means that whenever the branch is updated and the `bundle update` is run on `ent-search`, it will also update the connectors gem folder.

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/ent-search/pull/6372

  - [ent-search PR](https://github.com/elastic/ent-search/pull/6372)


lib/connectors_shared/constants.rb
```diff
@@ -11,4 +11,11 @@ class Constants
     ALLOW_FIELD = '_allow_permissions'.freeze
     DENY_FIELD = '_deny_permissions'.freeze
   end
+
+  class ConfigMetadata
```
tarekziade, line 15: What's the goal of moving from the YAML to here?

mchernyavskaya: There was a problem with path resolution when I tried to do the bundle install for `ent-search` with the github link:
```
[!] There was an error while loading `connectors_sdk.gemspec`: No such file or directory - /Users/marynacherniavska/.rbenv/versions/jruby-9.3.3.0/lib/ruby/gems/shared/bundler/gems/connectors-04a3b5430910/lib/connectors_app/../../config/connectors.yml. Bundler cannot continue.

 #  from /Users/marynacherniavska/.rbenv/versions/jruby-9.3.3.0/lib/ruby/gems/shared/bundler/gems/connectors-04a3b5430910/connectors_sdk.gemspec:1
 #  -------------------------------------------
 >  require_relative 'lib/connectors_app/config'
 #
 #  -------------------------------------------
```
The problem was, `connectors.yml` isn't included in the gemspec for `connectors_sdk` gem, and also we should have had a default `connectors.yml` (not the `connectors.yml.example`) in the project. But instead, we (me and @artem-shelkovnikov) decided it would be better to break off the gemspec dependency on the config file.

artem-shelkovnikov: Indeed. It's also possible to move these 3 variables to a different yaml file but I would probably oppose the idea.

Conceptually what I felt was - our gemspec file is not compatible with one of the default ways to distribute gems (`gem git:` in gemfile), so we need to fix it. Fundamentally, the problem is as described - our build process is "custom" - dependent on a shell script that itself depends on some configuration files, and we cannot make it work with `gem git:` approach, it's limited.

It's also a question to us - do we actually need anything other than VERSION here?

tarekziade: As long as these fields are in a single place in the whole project I think it's fine. But having the version in a static file makes our lives easier when we need to get that version (like how `ent-search` is doing) If you move it to a Ruby file, now you need Ruby to read it, it's harder to use from other places like bash scripts.

tarekziade:
This is our current requirements:

1. have the version and other metadata in a single place
2. make the version easy to read for our automation (jenkins CI scripts) and build process (gemspec), and preferably without requiring Ruby
3. expose the metadata in the HTTP endpoint, since it's used in telemetry and Kibana's UI
4. make it simple for our customers to tweak those metadata if they fork our project

I've initially put all the fields into the YAML file because it was the easiest way to meet all those requirements, and the configuration file *SHOULD* be shipped with the Gem. <-- and I think that fixes your issue?

I think the confusion is around 4.  how will the customers change the values in practice? do we really need to have that change in the gemspec? I am not sure, I think it's useful only for the HTTP endpoint publishing. But then the current
code does work : they deploy and just change the YAML file we have.

mchernyavskaya:
@tarekziade 
> But then the current code does work : they deploy and just change the YAML file we have.
But we don't have a `connectors.yml`. We just have a `connectors.yml.example` - the example file. So, the `connectors.yml` isn't in the repo. Which means that when we build `ent-search` using the github link and not the manually packaged gem, the build fails.

tarekziade:
> @tarekziade
> 
> > But then the current code does work : they deploy and just change the YAML file we have.
> 
> But we don't have a `connectors.yml`. We just have a `connectors.yml.example` - the example file. So, the `connectors.yml` isn't in the repo. Which means that when we build `ent-search` using the github link and not the manually packaged gem, the build fails.

Right, in my initial implementation that file was there and updated by the build process right before creating the gem.
It was renamed later to `connectors.yml.example` and the `make install` command copy it to `connectors.yml`

I think we could just go back to a single `connectors.yml` and remove the example one -- if not, we need to decide how to meet the requirements differently.

lib/connectors_shared/constants.rb
seanstory: I _think_ a gemspec is just a fancy ruby file - so arbitrary ruby code could be used. Could we just make parsing our gemspec actually do the same things as `make build` and then include the `connectors.yml` in the gemspec files?

> As long as these fields are in a single place in the whole project I think it's fine

I disagree - having this in code, and not configuration, and not updated by the build is not fine. It removes a solution key requirement of the Extractors As Dependencies project, which was to be able to clearly tell if a customer had forked our repo, and to help the customer know (from a support perspective) which version of their fork they are running on. Relying on the customer to make certain code changes is too brittle.

seanstory:
> I think we could just go back to a single connectors.yml and remove the example one -- if not, we need to decide how to meet the requirements differently.

It's on me, I made this change without a lot of discussion.

The issue is that `make build` would cause local changes to that file. And I saw enough PRs where people accidentally committed those changes that it seemed appropriate to take a different approach, like how ent-search has a `enterprise-search-local.yml.example`, and your first step after cloning the repo is to copy it.

Perhaps the next step in our solution is to have a `dist/connectors.yml` that our gem uses, that is distinct from the one that our Sinatra app and local development use?

RossAllenBell:
> Perhaps the next step in our solution is to have a `dist/connectors.yml` that our gem uses, that is distinct from the one that our Sinatra app and local development use?

Chiming in with a strong feeling: I agree with this, or something like it. We're using a single config file to support local gem development, local web application development, and a distributed gem. That's causing us to bend over backwards and we probably don't need to.

tarekziade:
> We're using a single config file to support local gem development, local web application development, and a distributed gem

Let's back up a bit, I think we're trying too hard. 

My understanding of Sean's initial intent is to make sure that when a customer runs a modified version, we can detect that it's not ours. Let's keep in mind that as soon as someone out there forks and uses our code in a server, there's not much we can do to fully enforce that they change the version.

As soon as a customer is changing the code he/she becomes a developer like us of this project. If a customer can change our ruby files, they can change the version wherever it is in our code base. They should be able to work with the codebase like how we do. We should have one single way to change the version when we work on the project. 

The other issue is that there's a bit of confusion around the role of the version string. It identifies the app but it is also central to our build process.

We should support the following stories:

- the version in the git repo matches the next release version (like ent-search and kibana does)
- as a developer working locally, when I run Sinatra, it uses the version that is in the current git branch, possibly with a dev marker
- when building the gem, the version in the git branch is used to create the Gem
- when the app is running in production, its version is published in `GET /`

Am I missing a case? 

I am going to reiterate one hard requirement when working with a project like this => we should avoid duplicating the version string across the project - it's a recipe for issues. These stories can be implemented without duplicating the version string.

Another way to solve this would be to stop relying on the version string, and publish a hash of the codebase files when we create our distribution, which will let us know for sure that no files were changed. But again, we should ask ourselves, is all of this worth the extra work?

lib/connectors_shared/constants.rb
tarekziade: >  Relying on the customer to make certain code changes is too brittle.

But if they fork and change the code, I think we can make the assumption they have the developer hat on and can change the version wherever we put that version

mchernyavskaya: Moving these changes back into `yml` due to the considerations above.

seanstory:
> My understanding of Sean's initial intent is to make sure that when a customer runs a modified version, we can detect that it's not ours. 

+1

> Let's keep in mind that as soon as someone out there forks and uses our code in a server, there's not much we can do to fully enforce that they change the version.

To enforce, no. But we can put out a good pattern that a customer would have to willfully break and violate in order to end up in a confusing state, rather than provide them a repo where they will be in a confusing state _unless_ they make specific changes. I'd rather have our default, right after a fork, be prepared to set them up for success. We currently have that in `main`.


> The other issue is that there's a bit of confusion around the role of the version string.

Good point, and @mchernyavskaya raised this above too with:

> And BTW, why do we need both version and git repo + git revision? Wouldn't just version suffice?

Let me see if I can clarify the purposes of these:

- `version` (example: `8.2.0`). This is a way that we can communicate in github.com/elastic/connectors which branch/version of the connectors go with which branch/version of ent-search. A customer is not beholden to keep this version - they may version it totally separately (example: `2022.01.31`), and that's fine. This is just a normal "software version" that hopefully follows semver.

- `repository`. (example: `git@github.com:elastic/ent-search-connectors.git` or `git@github.com:seanstory/connectors.git`). This helps us to know when customers are forking our repo, which implies made or planned modifications. This is helpful to have in our telemetry. It is also valuable in a Support scenario where a customer has an issue, and we can easily identify that they are running a forked version of our code, and not our repo itself.

- `revision`. (example: `1520395028d4f83435f491a9ea5971f02e3ac901`). The git commit for the the running code. This is primarily useful for Support. In the event that the `repository` is still listed as `git@github.com:elastic/ent-search-connectors.git`, but this `revision` does not match one of our tagged releases, we can determine if the code is "ours" or has been modified. It is also helpful for the customer, if, while troubleshooting their own modifications, they need to identify which version of their connector they are actually running.


> Another way to solve this would be to stop relying on the version string, and publish a hash of the codebase files when we create our distribution, which will let us know for sure that no files were changed.

I don't think that would give us anything that `revision` doesn't give us today.

tarekziade:
> I don't think that would give us anything that revision doesn't give us today.

It can be computed automatically when you start the service to reflect reality.

tarekziade:
For the `repository` and `revision` info, you are making the assumption that the customer is running the software in production from a Git clone they did fork. In practice, this may not happen.  

=> If I clone the project without creating a fork, I am breaking these assumptions. the repository will appear as being ours.
=> If I clone the project and I don't change the code, it can appear as being different while it's not
=> If I deploy the code in production without git (like, `.git` is gone) we can't really do much
=> In the future, for new connectors that are not using this repo, how will we provide this info? Maybe a customer will have a connector project that has no repository URL 

I am worried that we are making too many assumptions about how people will deploy our connector service in production.

I don't think that publishing `revision` or `repository` is needed at all. It's an internal detail. What really matters is to know what version we are running with, and if the customer changes it, know it.

So I would remove `revision` or `repository` and make it clear in our documentation that if people are using a modified version of this software in production they should flag it as such.

seanstory:
>=> If I clone the project without creating a fork, I am breaking these assumptions. the repository will appear as being ours.

the `repository` would, but the `revision` would be different if they've made changes. This is good/ok/expected.

> => If I fork the project and I don't change the code, it can appear as being different while it's not

If the code isn't changed, the `revision` and `version` will be the same. This is good/ok/expected. We want to know if people are forking our repo, this is important user data to capture, to know how folks are interacting with the open repo.

> => If I deploy the code in production without git (like, .git is gone) we can't really do much

This seems unlikely to me. But yes, if a user goes to these lengths, the burden is on them to figure out how to populate `repository` and `revision` in the API.

> => In the future, for new connectors that are not using this repo, how will we provide this metadata to telemetry? Maybe a customer will have a connector project that has no repository URL

This also seems unlikely to me, that our repo would be used without them using _our repo_. But yes, at some point, if the customer makes too many changes from our examples, the burden is on them to ensure they haven't broken something.


> So I would remove revision or repository and make it clear in our documentation that if people are using a modified version of this software in production they should flag it as such.

I disagree. I think we should be trying to lessen the burden on the user as much as possible, and give them good defaults where we can. Here, we can, by default, provide reasonable values for `version`, `repository`, and `revision` when they first fork our product, and let them worry about _maintaining_ those as they start making changes. I think this is better than insisting that, from initial fork, they _must_ update these values themselves, manually, or risk us having misleading or incorrect information reported via their API. Let's default to having useful data here.

tarekziade:
> Here, we can, by default, provide reasonable values for version, repository, and revision when they first fork our product, and let them worry about maintaining those as they start making changes.

ok, sounds good. It's something we can tweak later if we feel we should. Thanks for the discussion/feedback

tarekziade (approved):

RossAllenBell:
Trying to catch up on the discussion here. Am I missing anything that occurred between https://github.com/elastic/connectors/pull/62#issuecomment-1087919474 and now? It seems like there were some concerns and then we approved/merged.

Related, I'm noticing `config/connectors.yml` is now checked into Git, which I wasn't expecting but maybe that was a conclusion of the discussion?

RossAllenBell:
Trying to catch up on the discussion here. Am I missing anything that occurred between https://github.com/elastic/connectors/pull/62#issuecomment-1087919474 and now? It seems like there were some concerns and then we approved/merged.

Related, I'm noticing `config/connectors.yml` is now checked into Git, which I wasn't expecting but maybe that was a conclusion of the discussion?

Makefile
```diff
@@ -41,5 +37,10 @@ build-docker:
 run-docker:
 	docker run --rm -it -p 127.0.0.1:9292:9292/tcp connectors
 
+exec_app:
+	cd lib/app; bundle exec rackup config.ru
```
RossAllenBell, line 41: Should this command have been removed from `refresh_config`?.

seanstory:
> Am I missing anything that occurred between [#62 (comment)](https://github.com/elastic/connectors/pull/62#issuecomment-1087919474) and now? It seems like there were some concerns and then we approved/merged.

I was also surprised by this

> Related, I'm noticing config/connectors.yml is now checked into Git, which I wasn't expecting but maybe that was a conclusion of the discussion?

surprised by this too. Would there be objections to me making a change back to having `config/connectors.yml` be ignored, and have a different file that we include in the gemspec - mimicking how `enterprise_search.local.yml`, `enterprise_search.local.yml.example` and `dist/config/enterprise_search.yml` work together in ent-search?

RossAllenBell:
Either way, this PR is marked as `8.2.0` when I think that it should be `8.3.0`. I'm going to update the version labels so we don't accidentally backport this into `8.2`.

PR #66, opened 2022-03-31
Add Confluence Cloud connector

wangch079:
#### Pre-Review Checklist

- [x] New external service dependencies added.

## Related Pull Requests

* https://github.com/elastic/ent-search/pull/6468

### DIFF
--- a/lib/connectors_sdk/base/http_call_wrapper.rb
+++ b/lib/connectors_sdk/base/http_call_wrapper.rb
@@ -23,49 +23,49 @@ module ConnectorsSdk
       end
 
       def document_batch(params)
-        results = []
+        convert_third_party_errors do
+          results = []
 
-        extractor = extractor(params)
+          extractor = extractor(params)
 
-        extractor.yield_document_changes(:break_after_page => true, :modified_since => extractor.config.cursors['modified_since']) do |action, doc, download_args_and_proc|
-          download_obj = nil
-          if download_args_and_proc
-            download_obj = {
-              id: download_args_and_proc[0],
-              name: download_args_and_proc[1],
-              size: download_args_and_proc[2],
-              download_args: download_args_and_proc[3]
+          extractor.yield_document_changes(:break_after_page => true, :modified_since => extractor.config.cursors['modified_since']) do |action, doc, download_args_and_proc|
+            download_obj = nil
+            if download_args_and_proc
+              download_obj = {
+                  id: download_args_and_proc[0],
+                  name: download_args_and_proc[1],
+                  size: download_args_and_proc[2],
+                  download_args: download_args_and_proc[3]
+              }
+            end
+
+            results << {
+                :action => action,
+                :document => doc,
+                :download => download_obj
             }
           end
 
-          results << {
-            :action => action,
-            :document => doc,
-            :download => download_obj
-          }
+          [results, extractor.config.cursors, extractor.completed]
         end
-
-        [results, extractor.config.cursors, extractor.completed]
-      rescue custom_client_error => e
-        raise e.status_code == 401 ? ConnectorsShared::InvalidTokenError : e
       end
 
       def deleted(params)
-        results = []
-        extractor(params).yield_deleted_ids(params[:ids]) do |id|
-          results << id
+        convert_third_party_errors do
+          results = []
+          extractor(params).yield_deleted_ids(params[:ids]) do |id|
+            results << id
+          end
+          results
         end
-        results
-      rescue custom_client_error => e
-        raise e.status_code == 401 ? ConnectorsShared::InvalidTokenError : e
       end
 
       def permissions(params)
-        extractor(params).yield_permissions(params[:user_id]) do |permissions|
-          return permissions
+        convert_third_party_errors do
+          extractor(params).yield_permissions(params[:user_id]) do |permissions|
+            return permissions
+          end
         end
-      rescue custom_client_error => e
-        raise e.status_code == 401 ? ConnectorsShared::InvalidTokenError : e
       end
 
       def authorization_uri(params)
@@ -100,6 +100,12 @@ module ConnectorsSdk
       end
 
       private
+
+      def convert_third_party_errors
+        yield
+      rescue custom_client_error => e
+        raise e.status_code == 401 ? ConnectorsShared::InvalidTokenError : e
+      end
 
       def extractor_class
         raise 'Not implemented for this connector'


Gemfile.lock
```diff
@@ -11,6 +11,7 @@ GEM
     ast (2.4.2)
     attr_extras (6.2.5)
     bson (4.2.2)
+    bson (4.2.2-java)
```
seanstory, line 14: I think we shouldn't be pulling in the `*-java` gems

lib/connectors_shared/middleware/restrict_hostnames.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require 'faraday/middleware'
+
```
seanstory, line 4: I bet this is missing a lot of `require` statements to find classes like `AddressNotAllowed`, `Resolv`, `IPAddr`, etc

lib/connectors_shared/extraction_utils.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
```
RossAllenBell, line 1: I was surprised to see this here, but I'm seeing it's used in the base Confluence adapter. We'll have to decide how to extract it from the crawler. I think the crawler was intentionally designed to depend on a very limited set of gems, and doesn't currently rely on _connectors_sdk_.

I expect this file has no meaningful changes from its origin in Enterprise Search. Let me know if that's not the case.

Don't forget the copyright text.

seanstory (commented):
Seems reasonable to me thus far. Though we should probably move over the unit tests as part of this PR.

RossAllenBell (commented):
A quick pass review for now. It's a good thing we went through this second connector extraction to surface the Authorization and HttpWrapper problems. Going to spend a little bit thinking through those two problems.

lib/connectors_shared/extraction_utils.rb
seanstory: This also surprised me. When I dug into it, I found we've been using the utility like this since 2017. I also found this comment:

https://github.com/elastic/ent-search/blob/045d996b7468c0734e6ce0fa3ef2d74be40e1189/lib/extractor_utils.rb#L1-L5

which made me think we should just have this copied here, AND in crawler, and not have either depend on the other.

Gemfile.lock
RossAllenBell: I think this just points out that if you're bundling using jruby, this is the platform-dependent gem you'll end up with. Similarly with Windows ruby and `nokogiri (1.13.4-x86-mingw32)`. I could be wrong.

RossAllenBell (commented):
Spent some time thinking about the _HttpCallWrapper_ implementations. I had lost track of how much service-type-specific logic was still unabstracted, which this work does a great job of illustrating. Now I'm worried about how much logic is repeated between Connectors and Enterprise Search (ContentSourceService and SecretKeeper). These changes don't make that design worse, but I'm wondering if there's a way to reduce the repetition for the sake of future connector implementations.

Related, I _think_ we could push these _HttpCallWrapper_ subclasses to exist as _REGISTRY_ configuration entries. I'm bringing that up because I think avoiding having to introduce HttpCallWrapper implementation classes will be worth it for subsequent connectors. I'm open to suggestion. My suggestion doesn't make the approach all that much better.

lib/connectors_sdk/confluence_cloud/authorization.rb
```diff
@@ -0,0 +1,64 @@
...
+          response = HTTPClient.new.get(
+            'https://api.atlassian.com/oauth/token/accessible-resources',
+            nil,
+            'Accept' => 'application/json',
+            'Authorization' => "Bearer #{access_token}"
+          )
+          raise 'unable to fetch cloud id' unless HTTP::Status.successful?(response.status)
+          json = JSON.parse(response.body)
+
+          site = json.find { |sites| sites['url'] == 'https://workplace-search.atlassian.net' } || {}
```
RossAllenBell, line 58: I think this needs to refer to the configured third party URL?

lib/connectors_sdk/confluence_cloud/authorization.rb
```diff
@@ -0,0 +1,64 @@
...
+            readonly:content.attachment:confluence
+            search:confluence
+          ]
+        end
+
+        def additional_parameters
+          { :prompt => 'consent', :audience => 'api.atlassian.com' }
+        end
+
+        def fetch_cloud_id(access_token)
```
RossAllenBell, line 48: Most of _Authorization_ logic is just specifying service type specific OAuth2 behavior. This _Cloud ID_ logic strikes me as the only part that's something totally arbitrary to the third party. Is there a reason we couldn't look up the cloud id at extractor runtime? It would mean an extra call or two per sync, but it would make further attempts to abstract _Authorization_ logic easier.

RossAllenBell (commented):
Compared to _HttpCallWrapper_, it looks like _Authorization_ would be much easier to turn into a configuration (potentially in _REGISTRY_) rather than a subclass like it is here. The only odd thing I could see was the reference to Confluence Cloud _Cloud ID_, which I suggest maybe we could remove from this logic.

I'm still concerned that we're duplicating this logic across Enterprise Search and Connectors. Maybe in the long run we could reference this logic in Enterprise Search.

Gemfile.lock
wangch079: I ran `rbenv version` from the root of `connectors` and I got

```
2.6.9 (set by /Users/chenhui.wang/src/github.com/elastic/connectors/.ruby-version)
```

I think I'm not using jruby here?

lib/connectors_shared/extraction_utils.rb
wangch079: Noted. I will keep one copy in `connectors` and leave it unchanged in `ent-search`.

lib/connectors_sdk/base/http_call_wrapper.rb
```diff
@@ -0,0 +1,133 @@
...
+# frozen_string_literal: true
+
+require 'bson'
+
+module ConnectorsSdk
+  module Base
+    class HttpCallWrapper
+      def extractor(params)
+        extractor_class.new(
+          content_source_id: BSON::ObjectId.new,
```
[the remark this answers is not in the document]
wangch079: >I think we can use this if/when Enterprise Search sends a content source id during sync

It's not difficult to pass the real content source ID. I can introduce it in this PR https://github.com/elastic/ent-search/pull/6468

lib/connectors_sdk/base/http_call_wrapper.rb
```diff
@@ -0,0 +1,133 @@
...
+      def refresh(params)
+        authorization.refresh(params)
+      end
+
+      def download(params)
+        extractor(params).download(params[:meta])
+      end
+
+      def source_status(params)
+        client(params).me
```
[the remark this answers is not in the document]
wangch079: Confluence also has this implementation. But I agree it makes more sense to have connector-specific implementation. Will add a method something like `health_check`, and implement it in the subclasses.

lib/connectors_sdk/confluence_cloud/authorization.rb
wangch079: Nice catch

wangch079:
>Though we should probably move over the unit tests as part of this PR.

Will do. Just want to confirm the implementation is OK before I move the tests.

>I expect this file has no meaningful changes from its origin in Enterprise Search. Let me know if that's not the case.

Yes. I will also re-introduce this as a gem and do regression on OOTB Atlassian connectors.

>Don't forget the copyright text.

Will add it soon.

>Is there a reason we couldn't look up the cloud id at extractor runtime? It would mean an extra call or two per sync, but it would make further attempts to abstract Authorization logic easier.

That will make the [external extractor](https://github.com/elastic/ent-search/blob/main/connectors/lib/connectors/content_sources/external/extractor.rb) not connector-agnostic

>Related, I think we could push these HttpCallWrapper subclasses to exist as REGISTRY configuration entries.

>Compared to HttpCallWrapper, it looks like Authorization would be much easier to turn into a configuration (potentially in REGISTRY) rather than a subclass like it is here. The only odd thing I could see was the reference to Confluence Cloud Cloud ID, which I suggest maybe we could remove from this logic.

>I'm still concerned that we're duplicating this logic across Enterprise Search and Connectors. Maybe in the long run we could reference this logic in Enterprise Search.

I think the above is worth a sync meeting

RossAllenBell:
> > Is there a reason we couldn't look up the cloud id at extractor runtime? It would mean an extra call or two per sync, but it would make further attempts to abstract Authorization logic easier.
> 
> That will make the [external extractor](https://github.com/elastic/ent-search/blob/main/connectors/lib/connectors/content_sources/external/extractor.rb) not connector-agnostic

I'm not sure I understand. I think the places that need it could just look it up at runtime since they have an access token. I'm not sure why it was stored in SecretKeeper.

> I think the above is worth a sync meeting

I think that's absolutely reasonable. Whether it's now or after more investigation, could you schedule something? I think it will be important for @serenachou to be there as well.

Gemfile.lock
seanstory: if you restore the `Gemfile.lock` to how it is in `main`, then re-run `make install` from your root, do you still get the `-java` changes? I've had this happen when my connectors checkout is correct, but I run `make install` or `bundle install` from inside my IDE terminal.

wangch079: Yes, I still got thoses `-java` changes.

wangch079:
>Is there a reason we couldn't look up the cloud id at extractor runtime?

>I'm not sure I understand.

When you said `at extractor runtime`, I assume you mean [this](https://github.com/elastic/ent-search/blob/23514503d6913243b6e83b9b2af7e3b1322d52eb/connectors/lib/connectors/content_sources/external/extractor.rb#L7). But the `cloud_id` only applies to Jira/Confluence Cloud, therefore the external extractor has to know the service type to request the `cloud_id`

>could you schedule something?

I found it difficult to schedule anything this week. Could you help elaborate on your proposal? I don't want to drag this to next week.

RossAllenBell:
> When you said `at extractor runtime`, I assume you mean [this](https://github.com/elastic/ent-search/blob/23514503d6913243b6e83b9b2af7e3b1322d52eb/connectors/lib/connectors/content_sources/external/extractor.rb#L7). But the `cloud_id` only applies to Jira/Confluence Cloud, therefore the external extractor has to know the service type to request the `cloud_id`

I'm suggesting we push _cloud_id_ lookup into the Atlassian Cloud extractors (or custom clients) to be performed at sync runtime. That way the extra logic we have to introduce to this repo is reduced at the cost of an extra HTTP call during sync.

> I found it difficult to schedule anything this week. Could you help elaborate on your proposal? I don't want to drag this to next week.

These changes highlight how much logic we were duplicating between this repo and Enterprise Search (specifically in ContentSourceService and SecretKeeper). While these changes don't introduce more types of duplication, they do continue the duplication beyond SharePoint Online. I had two thoughts in reaction to reading through this, and I don't think the thoughts are necessarily mutually exclusive:

First, most of the logic in the _HttpCallWrapper_ and _Authorization_ implementations appears to be little more than configuration properties. To reduce the burden of introducing another connector beyond this, what if we had a configuration structure present in _REGISTRY_ that encapsulated the different properties. For example, something like this could allow the generic _HttpCallWrapper_ to do what it needs to do:

```ruby
REGISTRY.register('confluence_cloud', {
  :class_name => 'ConfluenceCloud',
  :name => 'Confluence Cloud',
  :custom_client_error => ConnectorsSdk::Atlassian::CustomClient::ClientError,
  :client_params => [:base_url, :access_token]
)
```

And second, regardless of how those implementation details live in this repo, can they then be used in Enterprise Search? For example, the OAuth2 _scopes_ specification. I think that would now exist in both projects, right?

I know this makes the Confluence Cloud extraction effort larger, but I think that's preferable so we can truly measure the cost of extracting a connector into this repo, and then also benefit from any part of this pattern we refine early on.

wangch079:
>I'm suggesting we push cloud_id lookup into the Atlassian Cloud extractors (or custom clients) to be performed at sync runtime.

This will apply to all 4 types of jobs and it means one additional request to Atlassian for each job (for each request to `POST /documents` before [this](https://docs.google.com/document/d/18ZYwVniE6ZU2HS_aoKXEO3zWmxkGuTdkQ1Bpc5xkIKA/edit) is done), until https://github.com/elastic/enterprise-search-team/issues/1585 is done. I'm fine with this proposal. cc @elastic/ingestion-team for more visibility.

>First, most of the logic in the HttpCallWrapper and Authorization implementations appears to be little more than configuration properties.

I think I understand what you mean. We will only have one `HttpCallWrapper` and one `Authorization` class, and we instantiate a new instance with configurations, e.g. `service_type` for `HttpCallWrapper`, `scope` for `Authorization`.

>For example, the OAuth2 scopes specification.

`ent-search` can refer to the scope in `ConnectorsSdk::{service_type}::Authorization` directly. I will make the change.

RossAllenBell:
It looks like we share understanding on what I was suggesting. Thanks for hearing me out. I'm absolutely open to alternatives as well. It wouldn't surprise me if you come up with a better idea than my suggestion after spending so much time in that logic.

wangch079:
> > I'm suggesting we push cloud_id lookup into the Atlassian Cloud extractors (or custom clients) to be performed at sync runtime.
> 
> This will apply to all 4 types of jobs and it means one additional request to Atlassian for each job (for each request to `POST /documents` before [this](https://docs.google.com/document/d/18ZYwVniE6ZU2HS_aoKXEO3zWmxkGuTdkQ1Bpc5xkIKA/edit) is done), until [elastic/enterprise-search-team#1585](https://github.com/elastic/enterprise-search-team/issues/1585) is done. I'm fine with this proposal. cc @elastic/ingestion-team for more visibility.
> 
> > First, most of the logic in the HttpCallWrapper and Authorization implementations appears to be little more than configuration properties.
> 
> I think I understand what you mean. We will only have one `HttpCallWrapper` and one `Authorization` class, and we instantiate a new instance with configurations, e.g. `service_type` for `HttpCallWrapper`, `scope` for `Authorization`.

I think I need more opinions from the team for point 1.

I will address the 2nd point in a separate PR.

lib/connectors_sdk/atlassian/custom_client.rb
```diff
@@ -0,0 +1,86 @@
...
+            if status_code == 504
+              ServiceUnavailableError
+            elsif status_code == 400 && response.body.include?('is not ContentConvertible or API available')
+              ContentConvertibleError
+            else
+              ClientError
+            end
+          raise atlassian_error_klass.new(response.env.url.to_s, status_code, response.body)
+        end
+        JSON.parse(response.body)
```
tarekziade, line 82: do you want to catch potential parsing errors here and returning a ClientError or should it be bubbled up as-is?

lib/connectors_sdk/base/http_call_wrapper.rb
```diff
@@ -0,0 +1,129 @@
...
+          results << {
+            :action => action,
+            :document => doc,
+            :download => download_obj
+          }
+        end
+
+        [results, extractor.config.cursors, extractor.completed]
+      rescue custom_client_error => e
+        raise e.status_code == 401 ? ConnectorsShared::InvalidTokenError : e
```
tarekziade, line 50: I realize that you are catching `ConnectorsShared::InvalidTokenError` everywhere. and always treating them the same way by converting them to a 401. I am wondering if this could be done in a single place in this class, that would decorate all functions

tarekziade, line 50: I realize that you are catching `ConnectorsShared::InvalidTokenError` everywhere. and always treating them the same way by converting them to a 401. I am wondering if this could be done in a single place in this class, that would decorate all functions

tarekziade (commented):
LGTM just a few nits. I'll let Sean or Ross approve

lib/connectors_sdk/atlassian/custom_client.rb
wangch079: This is just copied from [ent-search](https://github.com/elastic/ent-search/blob/85888454c342f3dfbdab654adae0e7a49845e1ca/connectors/lib/connectors/content_sources/atlassian/custom_client.rb#L57)

I think it assumes the response body is in JSON format as long as it's a successful. 

I think `JSON::ParserError` is more meaningful than `ClientError` because if that happens, it means Atlassian has done some unexpected changes again.

lib/connectors_sdk/base/http_call_wrapper.rb
wangch079: Yes we can do this. Will update soon

PR #67, opened 2022-04-01
Updated version to 8.3

seanstory:
an `8.2` branch has been created, so `main` should be updated to 8.3

#### Pre-Review Checklist


tarekziade (approved):
looks good. 

D'oh, two places for version already, we should have a single place. (not blocking for this update)

seanstory:
> D'oh, two places for version already, we should have a single place

Don't worry, the "example" file is just an example. The real one gets populated by the `VERSION` file: 
https://github.com/elastic/connectors/blob/df0b3c7fcc0a34696fac84194f2cac2320ddf91a/Makefile#L20

https://github.com/elastic/connectors/blob/df0b3c7fcc0a34696fac84194f2cac2320ddf91a/script/version.sh#L4-L9

PR #68, opened 2022-04-05
Index permissions in document

wangch079:
Related discussion: https://elastic.slack.com/archives/C030MMRDFL3/p1649142679941249

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/ent-search/pull/6421

### DIFF
--- a/lib/connectors_sdk/share_point/http_call_wrapper.rb
+++ b/lib/connectors_sdk/share_point/http_call_wrapper.rb
@@ -23,10 +23,10 @@ module ConnectorsSdk
         # XXX can we cache that class across calls?
         ConnectorsSdk::SharePoint::Extractor.new(
           content_source_id: BSON::ObjectId.new,
-          service_type: 'sharepoint_online',
+          service_type: SERVICE_TYPE,
           authorization_data_proc: proc { { access_token: params[:access_token] } },
           client_proc: proc { ConnectorsSdk::Office365::CustomClient.new(:access_token => params[:access_token], :cursors => cursors) },
-          config: ConnectorsSdk::Office365::Config.new(:cursors => cursors, :drive_ids => 'all'),
+          config: ConnectorsSdk::Office365::Config.new(:cursors => cursors, :drive_ids => 'all', :index_permissions => params[:index_permissions] || false),
           features: features
         )
       end
@@ -101,8 +101,8 @@ module ConnectorsSdk
         'SharePoint'
       end
 
-      def source_status(access_token)
-        client = ConnectorsSdk::Office365::CustomClient.new(:access_token => access_token)
+      def source_status(params)
+        client = ConnectorsSdk::Office365::CustomClient.new(:access_token => params[:access_token])
         client.me
         { :status => 'OK', :statusCode => 200, :message => 'Connected to SharePoint' }
       rescue StandardError => e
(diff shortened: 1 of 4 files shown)


lib/connectors_sdk/share_point/http_call_wrapper.rb
```diff
@@ -23,10 +23,10 @@ def extractor(params)
         # XXX can we cache that class across calls?
         ConnectorsSdk::SharePoint::Extractor.new(
           content_source_id: BSON::ObjectId.new,
-          service_type: 'sharepoint_online',
+          service_type: SERVICE_TYPE,
```
wangch079, line 26: `SERVICE_TYPE` is `sharepoint`, I don't think it make any more sense to use `sharepoint_online` here.

spec/connectors_sdk/sharepoint/http_call_wrapper_spec.rb
```diff
@@ -157,20 +157,20 @@ def mock_endpoint(path, data)
   end
 
   context '.source_status' do
-    let(:access_token) { 'access token' }
+    let(:params) { { :access_token => 'access_token' } }
```
tarekziade, line 160: maybe we could add `:index_permissions` here and verify that `ConnectorsSdk::Office365::Config` gets it

spec/connectors_sdk/sharepoint/http_call_wrapper_spec.rb
```diff
@@ -24,7 +24,10 @@
   end
 
   let(:params) do
-    { 'access_token' => 'something' }
+    {
+      'access_token' => 'something',
+      'index_permissions' => false
```
tarekziade, line 29: `false` is the default value in your code. can we switch this one to `true` and then assert that `ConnectorsSdk::Office365::Config.index_permissions` is `true` ?

tarekziade (approved):
Thanks! Is the `Gemfile.lock` changes intended?

wangch079:
> Thanks! Is the `Gemfile.lock` changes intended?

When I ran `make test` locally, it was changed. Do you see the same in your local branch?

tarekziade:
yeah I think it depends on the Ruby interpreter that is being used. I suspect some folks use JRuby through their IDE (like when @seanstory  committed this one https://github.com/elastic/connectors/commit/24f6a5cf4d4a61ba4f52dcf96125d023040d4dbd)

I think it's fine to push it like this, we should try to make sure we all use the same ruby when running the install

mchernyavskaya (approved):
LGTM

tarekziade:
> > I suspect some folks use JRuby through their IDE (like when @seanstory committed this one [24f6a5c](https://github.com/elastic/connectors/commit/24f6a5cf4d4a61ba4f52dcf96125d023040d4dbd))
> > I think it's fine to push it like this, we should try to make sure we all use the same ruby when running the install
> 
> Yeah, I found at some point that I had an `RBENV_VERSION` env var set, and it was overrriding the `.ruby_version` at the root of the repo. :/ Rbenv _should_ take care of this for us.

To be honest, I have seen this many times in `ent-search` too. I wonder how we can make this more robust.

seanstory:
@wangch079 I don't think this was backported to the 8.2 branch. This is intended for 8.2, per the label, is that correct?

wangch079:
@seanstory , yes it's for 8.2. I have created the backport PR: https://github.com/elastic/connectors/pull/71

PR #69, opened 2022-04-06
Paginate over the contents of a single SharePoint Online folder node

RossAllenBell:
In pursuit of: https://github.com/elastic/enterprise-search-team/issues/1795.

These changes allow paginating over files within a single folder. This requires adding an additional cursor data point.

#### Pre-Review Checklist

- [x] [Manual test steps](../docs/INTERNAL.md#minimal-manual-tests)  are successful


artem-shelkovnikov:
A comment not specific to this PR, but to extractor code - I've been away from developments for Connectors 2.0 for a bit and now it's difficult for me to follow what's going on during the extraction - before extractor was also a bit confusing, but now the difficulty of reading rose higher. Main point of confusion is mutable `cursors` object that changes during the extraction, making mind-mapping the method harder. `cursors` object having a complex structure stored inside (e.g. nesting of data) makes it more difficult to read too.

lib/connectors_sdk/office365/custom_client.rb
```diff
@@ -255,15 +259,30 @@ def get_root_item(drive_id, fields = [])
...
+                   else
+                     endpoint = "drives/#{drive_id}/items/#{item_id}/children"
+                     query_params = transform_fields_to_request_query_params(fields)
+                     request_endpoint(:endpoint => endpoint, :query_params => query_params)
+                   end
 
+        yielded = 0
         loop do
           response.value.each(&block)
           next_link = response['@odata.nextLink']
```
seanstory, line 276: I think you could always just assign `cursors['item_children_next_link'] = response['@odata.netLink']` here, and above always use `next_link = cursors['item_children_next_link']` (line 263) without having to couch these inside of `if break_after_page` checks. We should always use the cursor if there is a cursor, and if there isn't a cursor it means we don't need to use it. Right?

seanstory (dismissed):
Small comment, but I think this lgtm.

lib/connectors_sdk/office365/custom_client.rb
RossAllenBell: I think that'd cause external connector cursors to start showing up in internal connector cursors. With the if-wrapping, it also makes these easy to remove once we no longer need the per-data-source sync pagination logic like this.

seanstory: > I think that'd cause external connector cursors to start showing up in internal connector cursors

Isn't that a good thing? That would make our internal connector syncs better at resuming failed/crashed jobs, right? I definitely think we want a future where internal and external connectors still use the same cursoring (totally fine if that's not feasible today, just looking at this change I thought maybe it was).

RossAllenBell: Ah, these changes were not intended to affect internal connector behavior. While there's some potential value there, I've not thought at all about the effect of including this behavior in internal connector sync logic. Certainly not worth the risk at this particular point, pending the 8.2.0 release.

lib/connectors_sdk/office365/custom_client.rb
```diff
@@ -134,7 +134,10 @@ def list_items(drive_id, fields: [], break_after_page: false)
             yielded += 1
           end
 
-          if break_after_page && yielded >= 100 && stack.any?
+          if break_after_page && yielded >= 100
```
RossAllenBell, line 137: @wangch079 I want to make sure I understood your concern about this logic, so let me know if I missed something here. As we paginate over items in a folder, we return that folder to the stack, relying on `item_children_next_link` to let us pick back up in the middle of that folder as needed.

wangch079: What I want to prevent here is the scenario when it's time to break the page and there's nothing to extract from the current drive.

E.g. that's a drive `foo` with one folder `bar`, within which there are more than 100 documents. When `item_children` is done, we should have `yielded >= 100`, and `stack` is empty. In such a case `[]` will be set to `page_cursor`, which, after I looked depper, will still work: In the next request, it will try to crawl drive `foo` again, but with empty `stack`, it won't enter the while loop and continue with the next drive.

I'm not holding strong opinions now, but if you want to remove `stack.any?`, remember to change this line

https://github.com/elastic/connectors/blob/8e942066309419a552b337e36c4580384f857bcc/lib/connectors_sdk/office365/extractor.rb#L54

to

```
if break_after_page && config.cursors.key?('page_cursor')
```

wangch079:
When we extract documents, we try to break page when we hit 100 docs, but the default page size for Graph API is 200. Do you think we should change the page size to 100? It may not apply to incremental change because we are going to filter the results:

https://github.com/elastic/connectors/blob/8e942066309419a552b337e36c4580384f857bcc/lib/connectors_sdk/office365/custom_client.rb#L166-L167

But for full sync, it actually only break after 200 docs.

lib/connectors_sdk/office365/custom_client.rb
RossAllenBell: Given the scenario of a drive `foo` with one folder `bar` that contains over 100 documents, after `item_children` is done iterating over the items in folder `bar`, `cursors['item_children_next_link']` will be _nil_. That will cause `list_items` to set `cursors['page_cursor']` to `[]`. Because `[].present?` evaluates to _false_, `yield_document_changes` will immediately proceed onto the next drive (or complete if `foo` was the only/last drive).

Am I missing anything else there?

RossAllenBell:
> ... Do you think we should change the page size to 100? ...

I'm optimistic we'll be removing this logic in the next release. I'd prefer to leave it as revertable as possible. I chose a page size of _100_ with little-to-no thought behind it. I expect it'll be fine if we sometimes return ~200 documents.

lib/connectors_sdk/office365/custom_client.rb
wangch079: > Am I missing anything else there?

IMO, we should not store `[]` to `page_cursor`, and the presence of `page_cursor` indicates there's a page break (an empty stack should be considered present). Two negatives make a positive, so it works.

I'm also fine if we keep it this way.

BTW, you should check `item_children_next_link` here to determine whether we should break the page

https://github.com/elastic/connectors/blob/8e942066309419a552b337e36c4580384f857bcc/lib/connectors_sdk/office365/extractor.rb#L54-L56

and here to determine whether this is the end of the extraction

https://github.com/elastic/connectors/blob/8e942066309419a552b337e36c4580384f857bcc/lib/connectors_sdk/office365/extractor.rb#L59-L63

RossAllenBell: @wangch079 Can I get one last review? I've changed the logic to avoid setting `:page_cursor => []`, and I've added a check for `item_children_next_link`. Based on my understanding of the cursor logic, I _think_ the behavior is the same given how _page_cursor_ and _item_children_next_link_ relate to each other, but these changes certainly make the logical expectations more explicit.

lib/connectors_sdk/office365/extractor.rb
```diff
@@ -51,12 +51,12 @@ def yield_document_changes(modified_since: nil, break_after_page: false, &block)
...
           end
 
-          if break_after_page && config.cursors['page_cursor'].present?
+          if break_after_page && (config.cursors['page_cursor'].present? || config.cursors['item_children_next_link'].present?)
             break
           end
         end
 
-        if break_after_page && config.cursors['page_cursor'].blank?
+        if break_after_page && (config.cursors['page_cursor'].blank? && config.cursors['item_children_next_link'].blank?)
```
wangch079, line 59: parenthesis is not needed here?

wangch079 (approved):
LGTM except a small comment.

seanstory (approved):

PR #70, opened 2022-04-07
Return cursors at the end of extraction

wangch079:
#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/ent-search/pull/6441

### DIFF
--- a/lib/connectors_sdk/base/extractor.rb
+++ b/lib/connectors_sdk/base/extractor.rb
@@ -30,7 +30,7 @@ module ConnectorsSdk
         ]
       )
 
-      attr_reader :content_source_id, :config, :features, :original_cursors, :service_type
+      attr_reader :content_source_id, :config, :features, :original_cursors, :service_type, :completed
       attr_accessor :monitor, :client_proc
 
       def initialize(content_source_id:,
@@ -48,6 +48,7 @@ module ConnectorsSdk
         @authorization_data_proc = authorization_data_proc
         @original_cursors = config.cursors.deep_dup
         @monitor = monitor
+        @completed = false
       end
 
       def authorization_data!
--- a/lib/connectors_sdk/office365/custom_client.rb
+++ b/lib/connectors_sdk/office365/custom_client.rb
@@ -134,7 +134,7 @@ module ConnectorsSdk
             yielded += 1
           end
 
-          if break_after_page && yielded >= 100
+          if break_after_page && yielded >= 100 && stack.any?
             cursors['page_cursor'] = stack.dup
             break
           end
--- a/lib/connectors_sdk/office365/custom_client.rb
+++ b/lib/connectors_sdk/office365/custom_client.rb
@@ -50,6 +50,7 @@ module ConnectorsSdk
       def initialize(access_token:, cursors: {}, ensure_fresh_auth: nil)
         @access_token = access_token
         @cursors = cursors || {}
+        @cursors[ConnectorsSdk::Office365::Extractor::DRIVE_IDS_CURSOR_KEY] ||= {}
         super(:ensure_fresh_auth => ensure_fresh_auth)
       end
 
@@ -178,11 +179,11 @@ module ConnectorsSdk
           response = request_json(:url => response['@odata.nextLink'])
         end
 
-        cursors[drive_id] = response['@odata.deltaLink']
+        cursors[ConnectorsSdk::Office365::Extractor::DRIVE_IDS_CURSOR_KEY][drive_id] = response['@odata.deltaLink']
       end
 
       def get_latest_delta_link(drive_id)
-        cursors[drive_id] || exhaustively_get_delta_link(drive_id)
+        cursors[ConnectorsSdk::Office365::Extractor::DRIVE_IDS_CURSOR_KEY][drive_id] || exhaustively_get_delta_link(drive_id)
       end
 
       def exhaustively_get_delta_link(drive_id)
--- a/lib/connectors_sdk/office365/extractor.rb
+++ b/lib/connectors_sdk/office365/extractor.rb
@@ -20,19 +20,9 @@ module ConnectorsSdk
 
           if break_after_page
             current_drive_id = config.cursors['current_drive_id']
-            last_drive_id = config.cursors['last_drive_id']
-
             if current_drive_id.present? && current_drive_id > drive_id # they come alpha sorted
               next
             end
-
-            if last_drive_id.present? && last_drive_id >= drive_id # they come alpha sorted
-              if last_drive_id == drive_id
-                config.cursors.delete('last_drive_id')
-              end
-              next
-            end
-
             config.cursors['current_drive_id'] = drive_id
           end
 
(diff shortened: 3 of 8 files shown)


RossAllenBell (dismissed):
This makes sense to me. I think I'll actually wait for this to get merged before pursing the merge of my other folder cursoring fix.

lib/connectors_sdk/base/extractor.rb
```diff
@@ -30,7 +30,7 @@ class Extractor
         ]
       )
 
-      attr_reader :content_source_id, :config, :features, :original_cursors, :service_type
+      attr_reader :content_source_id, :config, :features, :original_cursors, :service_type, :completed
```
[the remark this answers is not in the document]
wangch079: More and more connectors will use it when they are migrated to external connector repo.

lib/connectors_sdk/office365/custom_client.rb
```diff
@@ -133,7 +134,7 @@ def list_items(drive_id, fields: [], break_after_page: false)
             yielded += 1
           end
 
-          if break_after_page && yielded >= 100
+          if break_after_page && yielded >= 100 && stack.any?
```
wangch079, line 137: @RossAllenBell There will be a side effect:
E.g. if a folder has more than 100 files (without amy folder), after crawling this folder completely, `yielded` will be greater than 100, but there's no more folders in the stack. Instead of breaking the page, we will move to the next drive. This is similar to incremental sync:

https://github.com/elastic/connectors/blob/c2e41de30b804b9df4e555704bd5650963a299b7/lib/connectors_sdk/office365/custom_client.rb#L172-L175

RossAllenBell: Oh, I don't think you need that `&& stack.any?`. The scenario you describe should get handled with: https://github.com/elastic/connectors/pull/69/files#diff-f15a307a042cbd4ac669d0c8c9096c03543bab73c45d8984378a4a8bc251d56aR138.

lib/connectors_sdk/office365/extractor.rb
```diff
@@ -61,14 +51,17 @@ def yield_document_changes(modified_since: nil, break_after_page: false, &block)
...
-            end
+          if break_after_page && config.cursors['page_cursor'].present?
             break
           end
         end
 
+        if break_after_page && config.cursors['page_cursor'].blank?
+          @completed = true
+          config.overwrite_cursors!(retrieve_latest_cursors)
+          log_info("Completed #{modified_since.nil? ? 'full' : 'incremental'} extraction")
```
seanstory, line 62: this just gave me a light-bulb moment - should we just add DEBUG logs for everywhere we're modififying and setting the `cursors`? Maybe that would help us to troubleshoot in the future, and feel good about our changes now.

spec/connectors_sdk/sharepoint/extractor_spec.rb
```diff
@@ -293,42 +293,21 @@ def assert_drive_filtered(drive)
         expect_sites([:id => site_id])
         expect_groups([])
         expect_site_drives(site_id, drives)
+        allow(extractor).to receive(:retrieve_latest_cursors).and_return({ described_class::DRIVE_IDS_CURSOR_KEY => {} })
```
seanstory, line 296: is it worth adding any tests to make sure that `completed` is getting set in the right scenarios? The changes here look like you made existing tests pass, but not like you're testing your changes fixed anything.

lib/connectors_sdk/office365/extractor.rb
```diff
@@ -61,14 +51,17 @@ def yield_document_changes(modified_since: nil, break_after_page: false, &block)
...
-              config.cursors['last_drive_id'] = config.cursors.delete('current_drive_id')
-            end
+          if break_after_page && config.cursors['page_cursor'].present?
             break
           end
         end
 
+        if break_after_page && config.cursors['page_cursor'].blank?
+          @completed = true
+          config.overwrite_cursors!(retrieve_latest_cursors)
```
RossAllenBell, line 61: @wangch079 I think at least one issue with the _full_ sync is with this line. You don't want to record those cursors, which are only used by _incremental_ syncs. It's possible they get passed on to the next full sync, making it behave like an incremental sync?

RossAllenBell:
@wangch079 has previously mentioned some issues with testing these changes that I was trying to reproduce so that we could get closer to merging this PR. I've been doing some testing with this branch and https://github.com/elastic/ent-search/pull/6441. This was the only issue I could find: https://github.com/elastic/connectors/pull/70/files#r847699965. I couldn't reproduce any issues with Incremental syncs.

With Full syncs, I noticed an issue where it started to behave like an Incremental sync, although that was only when I was accidentally running main Enterprise Search that didn't include the `completed?` changes. I think these changes might be fine once they stop persisting cursors for the Full jobs.

lib/connectors_sdk/base/extractor.rb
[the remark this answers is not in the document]
wangch079: @seanstory , yes this is for full/incremental syncs only.

lib/connectors_sdk/office365/extractor.rb
wangch079: It won't. `cursors` will always be `nil` for non-incremental syncs, check [here](https://github.com/elastic/ent-search/blob/7286805237716d62a5cdf0d9c1014d32a71b2ae8/frito_pie/app/models/frito_pie/content_source.rb#L651-L658)

The reason that `cursors` are persisted for `full` sync is because we try to retrieve `cursors` from the last completed job if we can't find the last completed `incremental` job ([ref](https://github.com/elastic/ent-search/blob/7286805237716d62a5cdf0d9c1014d32a71b2ae8/frito_pie/app/models/frito_pie/content_source.rb#L652)):
```
last_job = most_recent_completed_job(Connectors::Job::INCREMENTAL) || most_recent_completed_job
```

We actually persist `cursors` for both [full/incremental](https://github.com/elastic/ent-search/blob/7286805237716d62a5cdf0d9c1014d32a71b2ae8/connectors/lib/connectors/work/index.rb#L65-L68) job and [delete](https://github.com/elastic/ent-search/blob/7286805237716d62a5cdf0d9c1014d32a71b2ae8/connectors/lib/connectors/work/purge.rb#L59-L62) job. We don't do the same for [permissions](https://github.com/elastic/ent-search/blob/7286805237716d62a5cdf0d9c1014d32a71b2ae8/connectors/lib/connectors/work/permissions.rb#L36-L38) job, I guess we just missed it.

lib/connectors_sdk/office365/custom_client.rb
wangch079: I guess what you are trying to do here, is to add back the current crawling folder id to the stack if it's not completed (i.e. `item_children_next_link` is present)

What I would like to prevent here is to prevent the following scenario:
1. we have crawled the drive completely, therefore there's no more folder to crawl, i.e. `stack` is empty
2. we have more than 100 items to yield

In such case, an empty stack will be saved as `page_cursor`

```
if break_after_page && yielded >= 100
  if cursors['item_children_next_link'].present?
    stack << folder_id
  end
  cursors['page_cursor'] = stack.dup
  break
end
```

lib/connectors_sdk/office365/extractor.rb
RossAllenBell: Got it, that makes sense 👍

RossAllenBell (approved):

lib/connectors_sdk/office365/custom_client.rb
RossAllenBell: I think these changes are appropriate within this PR. The additional changes I intend to introduce should handle the scenario you describe.

PR #72, opened 2022-04-12
Clarify our versioning scheme and build process

tarekziade:
Added a `RELEASING.md` doc that captures how we deal with versions and release.


config/connectors.yml
```diff
@@ -1,7 +1,7 @@
 # general metadata
-version: "8.2.0-1648721082"
-repository: "git@github.com:elastic/connectors.git"
-revision: "48e7312286983c0197f7c3027487c4bfd6833090"
+version: "8.3.0.0"
```
seanstory, line 2: it feels weird to me to have these diffs included. Won't this start leading to merge conflicts all the time?

docs/RELEASING.md
```diff
@@ -0,0 +1,38 @@
...
+and the *BUILD* number should be set to **0** the day the release is created
+to be included with the Enterprise Search distribution.
+
+For example, when shipping for `8.1.2`, the version is `8.1.2.0`.
+
+To release Connectors:
+
+- set the VERSION file to the right version
+- make sure all tests and linter pass with `make lint test`
+- run `make release`
```
seanstory, line 20: We probably need additional steps to:
- tag the commit in git
- push the tag to `origin`
- if this is a new MINOR:
  - create a new maintenance branch 
  - update the VERSION on `main`
- create a Github release (idk if this is necessary?)
  - upload the gem to Github's release page (so that we have a source of truth on what the released artifact is)
- immediately update ent-search's gem dependency with the released version

script/version.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 prefix=`cat VERSION` # Retrieve the version prefix
-timestamp=`date +%s` # Calculate the current timestamp
+timestamp=`date +%Y%m%d%H%M%S` # Calculate the current timestamp in ISO8601 format
```
seanstory, line 5:
```suggestion
timestamp=`date +%Y%m%d%TH%M%S` # Calculate the current timestamp in ISO8601 format
```

I don't feel strongly about this, but I think the `T` is a standard separator for "date" and "time" in the ISO formats, and shouldn't break the sorting.

docs/RELEASING.md
tarekziade: thanks, great additions, I forgot about tagging. I guess that can be added in `make release` 
+1 for the branching

For GH releases, I don't know, I have never done this -- It sounds like a good idea though, or in the Gem artifact?

script/version.sh
tarekziade: According to Wikipedia, both are valid. (the full one has dashes) but T sounds fine, the `pre` segment is alpha-sorted

Makefile
```diff
@@ -1,4 +1,5 @@
 YQ ?= "yq"
+.phony: test lint autocorrect api_key update_config build release build_gem install refresh_config build-docker run-docker exec_app
```
RossAllenBell, line 2: TIL https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

config/connectors.yml
RossAllenBell: It was recently removed from `.gitignore` to allow building the gem dependency from git to work in Enterprise Search. I don't think I like this approach, but that's where it came from.

docs/RELEASING.md
seanstory: > For GH releases, I don't know, I have never done this -- It sounds like a good idea though

I hadn't done it before the other week. After the tag was there, it was just a few clicks in here: https://github.com/elastic/connectors/releases. Easy to pick which tag, if the release is a "pre-release" or not, and if you want to add release notes. And then you can upload artifacts (like the gem)

> or in the Gem artifactory service?

This would be my preference, but IDK if we have access to publish to our Elastic Artifactory, or how much work it would be to automate this. In my ideal world, we'd have a Jenkins job that ran the `make release`, created our tags/branches, and published our artifact all automatically.

config/connectors.yml
seanstory: Yeah, before https://github.com/elastic/connectors/pull/62 was merged, this file was in `.gitignore`. I'd prefer we put it back there to avoid these diffs and to avoid accidentally committing an API key.

script/version.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 prefix=`cat VERSION` # Retrieve the version prefix
-timestamp=`date +%s` # Calculate the current timestamp
+timestamp=`date +%Y%m%d%TH%M%S` # Calculate the current timestamp in ISO8601 format
```
RossAllenBell, line 5: Can we use a flat string of numbers for compatibility simplicity? Also I think there's an issue with the date format, and we should standardize to UTC:

```suggestion
timestamp=`date -u +%Y%m%d%H%M%S` # Calculate the current UTC timestamp
```

The ISO8601 format isn't compatible with `Gem::Version` 😞 

```
13:30:37 ~/dev/connectors> date +%Y%m%d%TH%M%S
2022041214:07:12H0712
14:07:12 ~/dev/connectors>
```

```
irb(main):007:0> Gem::Version.new('8.2.0.0-2022041214:07:12H0712')
       14: from /Users/rossbell/dev/ent-search/bin/enterprise-search-internal:15:in `<main>'
       13: from /Users/rossbell/dev/ent-search/shared_togo/lib/shared_togo/cli.rb:112:in `run!'
       12: from /Users/rossbell/dev/ent-search/shared_togo/lib/shared_togo/cli.rb:125:in `run_command'
       11: from /Users/rossbell/dev/ent-search/shared_togo/lib/shared_togo/cli.rb:143:in `run_supported_command'
       10: from /Users/rossbell/dev/ent-search/shared_togo/lib/shared_togo/cli/command.rb:10:in `run_and_exit'
        9: from /Users/rossbell/dev/ent-search/shared_togo/lib/shared_togo/cli/rails_console_command.rb:22:in `run'
        8: from org/jruby/RubyKernel.java:1237:in `catch'
        6: from org/jruby/RubyKernel.java:1507:in `loop'
        5: from org/jruby/RubyKernel.java:1091:in `eval'
        4: from (irb):7:in `evaluate'
        3: from /Users/rossbell/.rbenv/versions/jruby-9.3.3.0/lib/ruby/stdlib/rubygems/version.rb:205:in `new'
        2: from org/jruby/RubyClass.java:907:in `new'
        1: from /Users/rossbell/.rbenv/versions/jruby-9.3.3.0/lib/ruby/stdlib/rubygems/version.rb:214:in `initialize'
ArgumentError (Malformed version number string 8.2.0.0-2022041214:07:12H0712)
irb(main):008:0>
```

However, `20220412140712` would work with _anything_ and is easily human readable.

seanstory: Ah shit. I put the `T` in the wrong place for my suggestion.
```
Seans-MacBook-Pro-2:ent-search seanstory$ date +%Y%m%dT%H%M%S
20220412T142016
Seans-MacBook-Pro-2:ent-search seanstory$ date +%Y%m%d%TH%M%S
2022041214:21:08H2108
Seans-MacBook-Pro-2:ent-search seanstory$ 
```

The former is what I would like, I think. (Also with the `-u`, good call)

RossAllenBell: Got it, it still works with `T` and `Z`. I'll update my recommended change.

docs/RELEASING.md
tarekziade: > This would be my preference, but IDK if we have access to publish to our Elastic Artifactory,

We do, but we could also publish to RubyGems and store the creds in Vault

tarekziade: Ok I changed the code so we push it to RubyGems. I have created an account for ent-search-dev@elastic.co but I did not get the email back yet for registration :(

tarekziade:
note to myself: put the ruby gem password in vaulter and update the doc

RossAllenBell (dismissed):

Makefile
```diff
@@ -1,34 +1,56 @@
...
+	${YQ} e ".revision = \"$(shell git rev-parse HEAD)\"" -i config/connectors.yml
+	${YQ} e ".repository = \"$(shell git config --get remote.origin.url)\"" -i config/connectors.yml
+	${YQ} e ".version = \"$(shell cat VERSION)\"" -i config/connectors.yml
+
+build: update_config_dev build_gem
+
+release: update_config tag build_gem push_gem tag
+
+tag:
+	git tag $(shell cat VERSION)
```
seanstory, line 36: this won't include the `v.` that we use in ent-search. I think that's ok, but wanted to make sure it's intentional? Like we'll end up with `8.2.0.0` instead of `v.8.2.0.0`?

Makefile
```diff
@@ -1,34 +1,56 @@
...
+	${YQ} e ".version = \"$(shell script/version.sh)\"" -i config/connectors.yml
+
+update_config: config/connectors.yml
+	${YQ} e ".revision = \"$(shell git rev-parse HEAD)\"" -i config/connectors.yml
+	${YQ} e ".repository = \"$(shell git config --get remote.origin.url)\"" -i config/connectors.yml
+	${YQ} e ".version = \"$(shell cat VERSION)\"" -i config/connectors.yml
+
+build: update_config_dev build_gem
+
+release: update_config tag build_gem push_gem tag
```
seanstory, line 33: this has `tag` in here twice.

Makefile
```diff
@@ -1,34 +1,56 @@
...
+	git push --tags
+
+build_gem:
 	mkdir -p .gems
 	gem build connectors_sdk.gemspec
 	rm -f .gems/*
 	mv *.gem .gems/
 
+push_gem:
+	gem push .gems/*
```
seanstory, line 46: I don't see `connectors_sdk` in rubygems.org - have you run this already? Any reason to not run this on a pre-release, just to make sure it's all working as expected?

docs/RELEASING.md
```diff
@@ -0,0 +1,49 @@
+# Releasing the Connectors project
```
seanstory, line 1: Want to link this from the README at the bottom, in `### Other guides`?

Makefile
tarekziade: Ah, good point. I guess the `v` helps us make the difference between tags and branches (even if branches don't have the `PATCH` part) WIll add, thanks

Makefile
tarekziade: done https://rubygems.org/gems/connectors_sdk

docs/RELEASING.md
```diff
@@ -0,0 +1,59 @@
...
+
+The Connectors project is packaged as a Gem that is added in the Enterprise Search distribution.
+
+The version scheme we use is **MAJOR.MINOR.PATCH.BUILD** and stored in the `VERSION` file
```
wangch079, line 5: Can link to the `VERSION` file.

Makefile
```diff
@@ -1,35 +1,61 @@
...
-build:
+update_config_dev: config/connectors.yml
+	${YQ} e ".revision = \"$(shell git rev-parse HEAD)\"" -i config/connectors.yml
+	${YQ} e ".repository = \"$(shell git config --get remote.origin.url)\"" -i config/connectors.yml
+	${YQ} e ".version = \"$(shell script/version.sh)\"" -i config/connectors.yml
+
+update_config: config/connectors.yml
+	${YQ} e ".revision = \"$(shell git rev-parse HEAD)\"" -i config/connectors.yml
+	${YQ} e ".repository = \"$(shell git config --get remote.origin.url)\"" -i config/connectors.yml
+	${YQ} e ".version = \"$(shell cat VERSION)\"" -i config/connectors.yml
```
wangch079, line 30: In `ent-search` repo, we have [enterprise-search.yml](https://github.com/elastic/ent-search/blob/main/script/togo/enterprise-search.yml), and you can copy a local yml file (enterprise-search.local.yml) from here. Should we use the same pattern in connectors?

config/connectors.yml.example
```diff
@@ -1,5 +1,5 @@
 # general metadata
-version: "8.3.0-SNAPSHOT"
+version: "${VERSION}-${TIMESTAMP}"
```
wangch079, line 2: I just don't get it where are those values from? Will the `version` be set by 
https://github.com/elastic/connectors/blob/818872f5913abbfee249074c7f7657ba39debc70/Makefile#L25

and 

https://github.com/elastic/connectors/blob/818872f5913abbfee249074c7f7657ba39debc70/Makefile#L30

tarekziade: the `version` value is overwritten by the makefile. we have one for dev previews (generated by the version.sh script)

Makefile
tarekziade: not sure what you are proposing

config/connectors.yml.example
wangch079: Yes, so where is `${VERSION}-${TIMESTAMP}` from? envVar?

Makefile
wangch079: Can we have `config/connectors.yml` instead of `config/connectors.yml.example`?

You can copy a `config/connectors.local.yml` for local development. We can default to `config/connectors.yml` for all configs in Makefile, and use `config/connectors.local.yml` for `release_dev`?

config/connectors.yml.example
tarekziade: That's never used in practice -- could be CHANGEME

artem-shelkovnikov: I think I like CHANGEME or any other placeholder that does not use ${ENVIRONMENT_VARIABLES} to set expectations that we're not compiling the template version, but straight overwrite it

artem-shelkovnikov: Addressed in https://github.com/elastic/connectors/pull/72/commits/97a40b2af4876f12f9cf961aa1edf06348f9bb1b

docs/RELEASING.md
artem-shelkovnikov: Addressed in https://github.com/elastic/connectors/pull/72/commits/cc1c4611031af6ee6c2c2be0581f5fc8c722ef86

Makefile
artem-shelkovnikov: I'm not sure it's something that is needed for today - it's different from how it works in Enterprise Search. `enterprise-search.yml` config file is actually a redistributable config file that is used by docker or actual distribution, and it's possible to actually run Enterprise Search with this file.

For connectors the file cannot be used out-of-the box, cause it needs additional setup - such as api key, for example.

In future when we decide for example to run connectors with docker, we can move closer to this approach (having `config/connectors.yml` instead of `config/connectors.yml.example`, for now I think it's okay to leave it the way it is.

wangch079 (approved):

PR #74, opened 2022-04-13
Fixes to make the external connector work with rack-based managed hosting like Heroku

RossAllenBell:
These are changes I had to get the external connector service running in Heroku, plus a few general fixes I encountered along the way.

Original issue where I pursued Heroku compatibility: https://github.com/elastic/enterprise-search-team/issues/1643.

#### Pre-Review Checklist

- [X] [Manual test steps](../docs/INTERNAL.md#minimal-manual-tests)  are successful


Makefile
```diff
@@ -22,14 +22,13 @@ build:
...
 	rbenv install -s
-	- gem install bundler -v 2.2.33 && rbenv rehash
+	- gem install bundler -v `cat .bundler-version` && rbenv rehash
 	bundle install --jobs 1
 
 refresh_config:
 	${YQ} e ".revision = \"$(shell git rev-parse HEAD)\"" -i config/connectors.yml
 	${YQ} e ".repository = \"$(shell git config --get remote.origin.url)\"" -i config/connectors.yml
 	${YQ} e ".version = \"$(shell script/version.sh)\"" -i config/connectors.yml
-	cd lib/app; bundle exec rackup --host 0.0.0.0 config.ru
```
RossAllenBell, line 32: I'm fairly certain this line was not supposed to be here.

Procfile
```diff
@@ -0,0 +1 @@
+web: cd lib/app; bundle exec rackup config.ru -p ${PORT}
```
RossAllenBell, line 1: I don't know of a clever way to push this into `Makefile` off the top of my head. It didn't feel important enough to spend more time on.

lib/app/app.rb
```diff
@@ -58,7 +58,7 @@ class ConnectorsWebApp < Sinatra::Base
     # XXX to be removed
     return if settings.deactivate_auth
 
-    raise StandardError.new 'You need to set an API key in the config file' if ![:test, :development].include?(settings.environment) && api_key == ConnectorsApp::DEFAULT_PASSWORD
+    raise StandardError.new 'You need to set an API key in the config file' if ![:test, :development].include?(settings.environment) && settings.api_key == ConnectorsApp::DEFAULT_PASSWORD
```
RossAllenBell, line 61: A bug I noticed.

RossAllenBell:
Looking for a temperature check. If folks want to see a lot of changes before we could merge this, I think we'll put it off for now. It's not critical.

lib/app/app.rb
```diff
@@ -30,7 +30,7 @@ class ConnectorsWebApp < Sinatra::Base
     set :raise_errors, false
     set :show_exceptions, false
     set :bind, settings.http['host']
-    set :port, settings.http['port']
+    set :port, [ENV['PORT'], settings.http['port'], '9292'].detect(&:present?)
```
seanstory, line 33: I like this. No reason to do it in this PR, but we should add these env var overrides for all of these, which will make docker config a lot nicer.

Procfile
seanstory: This is the only thing that feels weird to me in this PR. Could this not call `make run` instead? Maybe that directive in the Makefile would need to have a `-p ${PORT:"9292"}`?

seanstory (dismissed):
small comments, but overall, LGTM, and seems like a valuable enhancement.

Gemfile
```diff
@@ -30,14 +33,14 @@ group :test do
...
 # Dependencies for the HTTP service
-gem 'sinatra'
-gem 'sinatra-contrib'
-gem 'rack'
-gem 'forwardable'
-gem 'faraday'
-gem 'httpclient'
-gem 'attr_extras'
-gem 'hashie'
+gem 'sinatra', '~> 2.1.0'
```
tarekziade, line 36: pinning FTW! thanks

Procfile
tarekziade: In the makefile you can write at the top something like:

```
PORT ?= 9292
```
and then 

```
cd lib/app; bundle exec rackup config.ru -p $(PORT)
```

then we you use it, it'll use `9292` or the env variable:

```
PORT=8888 make run
```

RossAllenBell: But then we ignore the port setting in `config/connectors.yml`. It requires parsing YAML to know that value. I couldn't get that working smoothly in Heroku and gave up after ~15mins (`yq` is not available).

Gemfile.lock
```diff
@@ -11,6 +11,7 @@ GEM
     ast (2.4.2)
     attr_extras (6.2.5)
     bson (4.2.2)
+    bson (4.2.2-java)
```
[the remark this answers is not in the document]
RossAllenBell: I don't actually know why bundler did this. It seemed to be in response to specifying MRI ruby in _Gemfile_.

Gemfile
RossAllenBell: Something that's important to note here, these pinned versions were simply pulled from the existing _Gemfile.lock_ values. Some of them _are not_ compatible with Enterprise Search. It's not an issue for now because our gemspec has very limited dependencies, but it could become a problem if we ever need to add any of these to the gemspec.

Gemfile.lock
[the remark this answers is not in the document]
seanstory: Check if you have an `RBENV_VERSION` env var set? I've had that screw up my lock files in the past, as it overrides the local `.ruby_version` file

Procfile
RossAllenBell: Had to rebase with _main_. I believe this is the only real concern remaining. Thoughts on whether it's worth resolving or leaving as is?

seanstory: We can always fix it later.

seanstory (approved):

PR #75, opened 2022-04-14
[DO NOT MERGE] use Puma

tarekziade:
Use Puma so we can have multiple threads


artem-shelkovnikov:
What does this PR change to the connectors?

Having no knowledge about Puma, I see that it allows to add parallelism to Sinatra app - does it work just by changing the `make run` script?

tarekziade:
this should not be merged. it jus shows how to run using Puma

see also https://github.com/tarekziade/asyncpuma

PR #81, opened 2022-04-15
update version to 8.2.0.1

seanstory:
post the 8.2.0.0 release


RossAllenBell (approved):
This is really interesting. Most Elastic projects will bump their _8.2_ branches to reflect `8.2.1` next week after the _8.2.0_ release, but we wont do that until the unified Elastic _8.2.1_ FF. This is because we want to be able to reflect mid-release changes to _8.2.0_ between now and then. I'm mostly just explaining this to myself outloud.

I'll include this in the updates to _RELEASING_.

PR #83, opened 2022-04-20
Stop caching the instance of HttpCallWrapper, which caused it to behave like a singleton class

RossAllenBell:
We discovered that the HttpCallWrapper instance was being shared between requests. These changes ensure it is used as a normal class/object, rather than what is effectively a singleton class. I expect this would resolve some thread-unsafe behavior.

These changes also remove the thread-unsafe behavior, but I think it's still a good idea to remove the singleton class behavior as well.

#### Pre-Review Checklist
- [X] Covered the changes with automated tests

## Release Note

Fixed an issue where two requests to the `/documents` endpoint could interfere with each other in a thread-unsafe manner.

- [X] [Manual test steps](https://github.com/elastic/connectors/blob/main/docs/INTERNAL.md#minimal-manual-tests)  are successful

### DIFF
--- a/lib/app/app.rb
+++ b/lib/app/app.rb
@@ -54,6 +54,8 @@ class ConnectorsWebApp < Sinatra::Base
   end
 
   before do
+    @connector = settings.connector_class.new
+
     Time.zone = ActiveSupport::TimeZone.new('UTC')
     # XXX to be removed
     return if settings.deactivate_auth
@@ -81,16 +83,15 @@ class ConnectorsWebApp < Sinatra::Base
   end
 
   post '/status' do
-    connector = settings.connector_class.new
-    source_status = connector.source_status(body_params)
+    source_status = @connector.source_status(body_params)
     json(
-      :extractor => { :name => connector.name },
+      :extractor => { :name => @connector.name },
       :contentProvider => source_status
     )
   end
 
   post '/documents' do
-    results, cursors, completed = settings.connector_class.new.document_batch(body_params)
+    results, cursors, completed = @connector.document_batch(body_params)
 
     json(
       :results => results,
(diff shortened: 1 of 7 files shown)


seanstory:
Disclaimer: I don't know anything about Sinatra.

I _think_ that this could be accomplished more cleanly with a "before filter". cmd+f that here: http://sinatrarb.com/intro.html

Screen Shot 2022-04-20 at 3 45 56 PM

seanstory (dismissed):
LGTM, though see my note above about us potentially doing this with a `before` so as to not have to have `settings.connector_class.new` all over the app.rb

wangch079:
Agree with @seanstory , the before filter seems to be a more clean solution.

lib/app/app.rb
```diff
@@ -54,6 +54,8 @@ class ConnectorsWebApp < Sinatra::Base
   end
 
   before do
+    @connector = settings.connector_class.new
```
wangch079, line 57: I think with the change you made in `http_call_wrapper` (making `extractor` a method variable instead of instance variable), I think we can revert the changes here? Setting an instance of `http_call_wrapper` to `settings.connector` can avoid the instantiation in each request.

RossAllenBell: Yah, I ended up solving the problem twice, once in the _HttpCallWrapper_ implementation to make it thread-safe, and once in this controller code to make it instantiate a new _HttpCallWrapper_ on every request. I'd prefer to continue instantiating a new object on every request. The shared object approach was surprising and I would expect we'll accidentally introduce new regressions again in the future with it.

seanstory (approved):
LGTM, though do we need a test that would catch any thread safety regressions we might have? These are a pain to write and often cast a really small net, so probably not worth it, but thought I'd toss it out.

RossAllenBell:
> LGTM, though do we need a test that would catch any thread safety regressions we might have? These are a pain to write and often cast a really small net, so probably not worth it, but thought I'd toss it out.

I looked into it for a few minutes earlier and came to the exact same conclusion.

PR #85, opened 2022-04-25
 Connector support for API Token Auth POC

mchernyavskaya:
#### Pre-Review Checklist


config/connectors.yml
```diff
@@ -0,0 +1,11 @@
+# general metadata
```
tarekziade, line 1: do we want to commit this file? (maybe we should add it in `.gitignore`

lib/connectors_sdk/base/authorization.rb
```diff
@@ -70,6 +70,10 @@ def oauth_scope
           raise 'Not implemented for this connector'
         end
 
+        def basic_auth_token(params)
+          raise 'Not implemented for this connector'
```
tarekziade, line 74: should we raise `NotImplementedError` ?

lib/connectors_sdk/confluence_cloud/extractor.rb
```diff
@@ -52,7 +52,8 @@ def yield_permissions(source_user_id)
       def download(item)
         content = item[:content]
         parent_id = content.dig('container', 'id')
-        client.download("#{client.base_url}/wiki/rest/api/content/#{parent_id}/child/attachment/#{content['id']}/download").body
+        wiki_path = client.base_url.downcase.include?('/wiki') ? '' : '/wiki'
```
tarekziade, line 55: what's the goal of this change? it seems unrelated to the Auth POC

spec/connectors_sdk/confluence/custom_client_spec.rb
```diff
@@ -90,6 +101,37 @@
     expect(content_search_request).to have_been_requested
   end
 
+  it 'applies correct middleware for token auth' do
```
tarekziade, line 104: awesome!

tarekziade (commented):
Design question

We're adding everywhere `basic_auth_token` and now each class has to deal with `access_token` and `basic_auth_token`

I am wondering if we should not move the complexity away from options passing.
e.g. define a new `AuthenticationToken` class along with a type (basic, bearer)  that can be used to forge our requests and provide all helpers around the actual raw token. I am worried that we are carrying around auth details that could be encapsulated.

lib/connectors_sdk/confluence_cloud/extractor.rb
mchernyavskaya: It is indeed related. For the Confluence Cloud, the base API URL is different for OAuth and API Token Auth. And it already includes `wiki` in the second case. So I had to add this to not add it again.

mchernyavskaya:
Closing as not relevant. The BYOC work makes this obsolete.

mchernyavskaya:
partially the change was merged in [this PR](https://github.com/elastic/connectors/pull/118) by Sean

PR #86, opened 2022-05-02
Support custom connectors

RossAllenBell:
Screen Shot 2022-05-05 at 10 06 52 AM

These changes allow the external connector service to report connector definitions to Enterprise Search during configuration. This is particularly useful in that it allows the external connector service to define a connector that Enterprise Search has no knowledge or configuration data for.

Relies on: https://github.com/elastic/ent-search/pull/6504.
Relies on: https://github.com/elastic/kibana/pull/131935.

#### Pre-Review Checklist


lib/connectors_sdk/fake_connector/http_call_wrapper.rb
```diff
@@ -0,0 +1,62 @@
...
+            'label' => 'Third Party API Key'
+          }
+        ]
+      end
+
+      def health_check(_params)
+        true
+      end
+
+      def document_batch(_params)
```
mchernyavskaya, line 37: I would still argue that we should do it in a "proper" way - with a client, if not extractor and adapter - but perhaps this "making the code into a template" should be scoped separately. I don't think it'll be a great effort though, as this FakeConnector needs no config and no adapter, just an extractor and custom_client.

lib/app/app.rb
```diff
@@ -90,6 +90,10 @@ class ConnectorsWebApp < Sinatra::Base
     )
   end
 
+  get '/describe' do
```
tarekziade, line 93: snice this sends back metadata about the connector, I am wondering if this could not be added in what `GET /`  returns ?

tarekziade (commented):
LGTM, I am just wondering about `/` versus `/describe` -- I think they could be merged into a single endpoint

lib/connectors_sdk/base/http_call_wrapper.rb
```diff
@@ -99,8 +99,24 @@ def service_type
...
+          :name => name,
+          :connection_requires_redirect => connection_requires_redirect,
+          :configurable_fields => configurable_fields
+        }
+      end
+
       private
 
+      def connection_requires_redirect
+        false
```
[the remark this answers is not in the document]
artem-shelkovnikov: Interestingly, no Crest connector uses OAuth and as far as I could understand we'd also love to get rid of OAuth, I think that's a good default value

lib/connectors_sdk/base/http_call_wrapper.rb
```diff
@@ -99,8 +99,24 @@ def service_type
         self.class::SERVICE_TYPE
       end
 
+      def describe
+        {
+          :name => name,
+          :connection_requires_redirect => connection_requires_redirect,
+          :configurable_fields => configurable_fields
```
artem-shelkovnikov, line 106: What are configurable fields? Is it the connector settings?

lib/connectors_sdk/base/http_call_wrapper.rb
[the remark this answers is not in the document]
RossAllenBell: I think of the OAuth handshake as a special case, even if the majority of our internal connectors use it. Mostly because there's no _alternative_ flow to the OAuth handshake, we simply don't do it if it's not needed. Leaving it as defaulted to _false_ is a more natural default as connectors should opt into the redirect behavior.

And as Artem alludes to, I expect it to be uncommon for _new_ connectors.

lib/connectors_sdk/fake_connector/http_call_wrapper.rb
RossAllenBell: I think it would be nice to introduce at least an adapter for this connector. But yes, I think that should all be part of a subsequent PR.

lib/app/app.rb
RossAllenBell: I think that would work 👍

lib/connectors_sdk/base/http_call_wrapper.rb
RossAllenBell: These are the fields the Enterprise Search UI should present to the user when configuring the connector.

lib/connectors_sdk/confluence_cloud/http_call_wrapper.rb
```diff
@@ -17,12 +17,29 @@ module ConfluenceCloud
...
-      def service_type
-        SERVICE_TYPE
+      def connection_requires_redirect
+        true
+      end
+
+      def configurable_fields
+        [
+          {
+            'key' => 'base_url',
```
wangch079, line 31: Do you think we should prefix config fields from external connector with `external_connector_`? e.g. `external_connector_base_url`

wangch079 (approved):
Leave a comment, overall LGTM.

BTW, what is the purpose of adding a fake connector?

mchernyavskaya (dismissed):

lib/connectors_sdk/confluence_cloud/http_call_wrapper.rb
RossAllenBell: Not a bad idea, although the value it brings is avoiding collision with other configuration fields, which happen to be `external_connector_url` and `external_connector_api_key`, which of course already contain the _external_connector__ prefix. What I _really_ wanted to do was introduce a new hash level for these fields, but that would've increased a lot of scope. We have a lot of code that assumes those values all live in a single, flat hash. At this point I suspect it's not worth the effort.

RossAllenBell:
> BTW, what is the purpose of adding a fake connector?

I wanted to illustrate how this worked in a more stripped down implementation. Once I put something simple in place, I realized it actually has a lot of value as an example template. It could certainly benefit from additional content (ex: adding a real Extractor/Adapter).

lib/app/app.rb
```diff
@@ -78,14 +76,17 @@ class ConnectorsWebApp < Sinatra::Base
       :connectors_version => settings.version,
       :connectors_repository => settings.repository,
       :connectors_revision => settings.revision,
-      :connector_name => ActiveSupport::Inflector.camelize(settings.http['connector'])
+      :connector_name => settings.connector_name,
+      :display_name => @connector.display_name,
+      :configurable_fields => @connector.configurable_fields,
```
RossAllenBell, line 81: Unfortunately this is the one change that makes this PR backwards incompatible without: https://github.com/elastic/ent-search/pull/6504. The nested hash breaks the content source details UI. I'll need to wait until both the Enterprise Search and Kibana PRs are ready to merge.

wangch079:
> Once I put something simple in place, I realized it actually has a lot of value as an example template.

It will be better renamed as `example_connector` or `sample_connector`?

mchernyavskaya:
@wangch079 I also had a list of proposals [here](https://elastic.slack.com/archives/C01795T48LQ/p1651561894211329?thread_ts=1651521993.461929&cid=C01795T48LQ):

* Generic Connector
* Stub/Mock Connector
* Empty Connector (but that's actually not quite right as it does produce data)
* or just Connector Template (again, I am a bit wary here since IMHO a template would just contain `#TODO: implement in places` instead of actual code)

tarekziade:
> I had envisioned we provide an example connector with fully implemented extractor/client/adaptor examples, but now I'm unsure if it's worth the effort. If anybody feels particularly strongly about what the name should be, I can change it, otherwise I was going to leave it as is.

What about a toy implementation of a filesystem directory instead of a fake connector?  
That's a connector we could definitely use in functional tests without having to link a real backend service

RossAllenBell:
> What about a toy implementation of a filesystem directory instead of a fake connector?
That's a connector we could definitely use in functional tests without having to link a real backend service

My only mild concern for a local filesystem example is that we already have a number of ingestion options related to that, and it could risk adding more confusion. That being said, I probably wont increase the scope of this PR in the near term.

mchernyavskaya:
@RossAllenBell is it almost there? =)

RossAllenBell:
Sorry for the re-review request. I needed to manually resolve a merge conflict and I've also renamed _Fake_ -> _Stub_.

PR #87, opened 2022-05-04
Artem/switch to async

artem-shelkovnikov:
#### Pre-Review Checklist


lib/app/app.rb
```diff
@@ -54,8 +57,6 @@ class ConnectorsWebApp < Sinatra::Base
   end
 
   before do
-    @connector = settings.connector_class.new
```
RossAllenBell, line 57: Was there a reason to get rid of this initialization in the before block?

lib/connectors_async/job.rb
```diff
@@ -0,0 +1,76 @@
...
+      up_to.times do
+        break if @data[:documents].empty?
+
+        results << @data[:documents].pop(true)
+      end
+
+      results
+    rescue ThreadError
+      puts 'Attempt to access an empty queue happened, when the queue was not supposed to be empty'
+      []
```
RossAllenBell, line 67: I suspect this wouldn't happen in practice, but if it _did_, you would lose whatever had been popped into `results` unless:
```suggestion
      results
```

lib/connectors_async/job_runner.rb
```diff
@@ -0,0 +1,44 @@
...
+
+require 'concurrent'
+
+require 'connectors_async/job_store'
+require 'connectors_shared/job_status'
+
+module ConnectorsAsync
+  class JobRunner
+    def initialize
+      @pool = Concurrent::ThreadPoolExecutor.new(min_threads: 1, max_threads: 4, max_queue: 0)
```
RossAllenBell, line 15: We'll want to make `max_threads` configurable. I don't have a strong opinion as to which PR that falls under.

lib/app/app.rb
```diff
@@ -83,53 +84,90 @@ class ConnectorsWebApp < Sinatra::Base
...
+    source_status = connector.source_status(body_params)
     json(
-      :extractor => { :name => @connector.name },
+      :extractor => { :name => connector.name },
       :contentProvider => source_status
     )
   end
 
   post '/documents' do
-    results, cursors, completed = @connector.document_batch(body_params)
```
RossAllenBell, line 94: Yah, related to a similar comment in the other PR, I think we need to respect `cursors` at the beginning and end of the sync.

lib/connectors_sdk/share_point/http_call_wrapper.rb
```diff
@@ -20,6 +20,47 @@ def name
         'SharePoint'
       end
 
+      def extractor(params)
```
RossAllenBell, line 23: Is this `extractor` and `extract` method override needed from what's already in https://github.com/elastic/connectors/blob/main/lib/connectors_sdk/base/http_call_wrapper.rb?

Gemfile
```diff
@@ -40,9 +40,11 @@ gem 'rack', '~> 2.2.3'
 gem 'forwardable', '~> 1.3.2'
 gem 'faraday', '~> 1.10.0'
 gem 'faraday_middleware', '= 1.0.0'
+gem 'typhoeus', '~> 1.4.0'
```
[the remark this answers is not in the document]
artem-shelkovnikov: Good catch, dropped in cc223da711fd8f4c8877d3fe40803be1df8ba29c

lib/app/app.rb
artem-shelkovnikov: So it was a chain of thoughts.

JobRunner actually creates a new thread that does the syncing logic. A good idea would be to create an instance of connector inside the thread. Therefore, we don't want to create an instance of connector before running `/documents` endpoint.

Further, if the controller becomes async, it feels like objects with Per-Request lifecycle should be created inside of each individual request, not use a shared `@connector` instance variable that can be overwritten by a different request/thread - I might be wrong here though, I'm not super familiar with how instance variables are handled when multiple requests hit Sinatra server at the same time.

lib/app/app.rb
artem-shelkovnikov: That's a good point.

Big question for me - how much does it matter now since we're switching to a different approach to take.

cc @tarekziade @serenachou

lib/app/app.rb
```diff
@@ -83,53 +84,90 @@ class ConnectorsWebApp < Sinatra::Base
...
+    job_id = body_params[:job_id]
+
+    job = nil
+
+    if job_id
+      job = settings.job_store.fetch_job(job_id)
+    end
+
+    unless job
```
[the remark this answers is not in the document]
artem-shelkovnikov: This is an interesting part of code, consider the following scenario:

Connector crashes and restarts in the middle of sync. Once it's back, Enterprise Search will call it with the jobId that does not exist any more. What should happen?

The way I implemented it now, it'll be a transparent retry for the job, but on the other hand failing a job and forcing it to start again is a valid point too.

lib/connectors_async/job.rb
artem-shelkovnikov: Good point, fixed!

lib/connectors_async/job_runner.rb
artem-shelkovnikov: Good call, I've made an attempt to do so in 85f3d011ef85a0b813e8d39374ffeba26b40ce36

lib/connectors_sdk/share_point/http_call_wrapper.rb
artem-shelkovnikov: Ohhh it was a merge conflict for me, fixed in 315ae4a887529696745af50556e6a846433e64ff

lib/app/app.rb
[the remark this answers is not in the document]
artem-shelkovnikov: Nvm I tested it a bit and I think you are totally right, we should separate those.

It brought me some problems due to the fact that if connector breaks and restarts, Enterprise Search won't send an access token at all, so the sync will be stuck and won't be able to contunue due to 401s. I'll break it into two endpoints.

lib/app/app.rb
RossAllenBell: The way you have it is probably fine. For what it's worth, I do not believe separate requests share `@connector`.

lib/app/app.rb
RossAllenBell: I don't think we can consider the async efforts here complete without preserving `cursors` behavior. We don't need to handle mid-sync cursors like we had introduced for SharePoint Online, but if the final cursors value is not reported back to Enterprise Search after a SharePoint Online incremental sync, I think every incremental sync would start to take about a full sync's amount of time.

tarekziade: yeah I also think we should keep the existing behavior at the API level for now, even if the transient cursor is not really used

lib/connectors_async/job_store.rb
```diff
@@ -0,0 +1,37 @@
...
+      # multiple threads can write to the store
+      @store = Concurrent::Hash.new
+    end
+
+    def create_job
+      job_id = SecureRandom.uuid
+
+      job = ConnectorsAsync::Job.new(job_id)
+
+      @store[job_id] = job
```
RossAllenBell, line 26: Do we ever remove old jobs from `@store`?

RossAllenBell (commented):
Question about HTTP API error flow and the job `@store`.

lib/connectors_async/job_store.rb
seanstory: Yes, I believe that we only keep job history for ~30 days? I'm not quickly finding where that job history deletion logic lives though.

wangch079: Agreed. `job_store` should be able to clean finished/stale jobs. e.g. 
```
@store.delete(job.id) if job.is_finished? || job.stale?
```

In class `Job`
```
def stale?
  Time.now - last_updated > 5.mins
end
```

lib/app/app.rb
```diff
@@ -84,51 +85,92 @@ class ConnectorsWebApp < Sinatra::Base
...
+
+    json(
+      :job_id => job.id,
+      :status => job.status
+    )
+  end
+
   post '/documents' do
-    results, cursors, completed = @connector.document_batch(body_params)
+    job = settings.job_store.fetch_job(body_params[:job_id])
```
wangch079, line 115: Should catch `JobNotFoundError`?

lib/connectors_async.rb
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+require 'connectors_async/job_runner'
```
wangch079, line 7: What is the purpose of this file?

lib/connectors_async/job.rb
```diff
@@ -0,0 +1,88 @@
...
+
+require 'connectors_shared/job_status'
+
+# The class is actually supported for single-threaded usage EXCEPT for :documents field
+# :documents are a Queue that's stated to be safe in a threaded environment
+module ConnectorsAsync
+  class Job
+    class StatusUpdateError < StandardError; end
+    def initialize(job_id)
+      @data = {
```
wangch079, line 15: Should store `content_source_id`, `job_id`, `job_type` as well?

lib/connectors_async/job.rb
```diff
@@ -0,0 +1,88 @@
...
+        :job_id => job_id,
+        :status => ConnectorsShared::JobStatus::CREATED,
+        :documents => Queue.new
+      }
+    end
```
wangch079, line 20: Instead of using a hash `@data` instance variable, why not use variables like `job_id`, `status`, `doc_queue`, etc...?

lib/app/app.rb
```diff
@@ -84,51 +85,92 @@ class ConnectorsWebApp < Sinatra::Base
...
-      :extractor => { :name => @connector.name },
+      :extractor => { :name => connector.name },
       :contentProvider => source_status
     )
   end
 
+  post '/start_sync' do
+    job = settings.job_store.create_job
+
+    settings.job_runner.start_job(
```
wangch079, line 100: Should catch `RejectedExecutionError` here? If the pool is fully utilized and the job is rejected, we should return some meaningful message here?

lib/app/app.rb
```diff
@@ -84,51 +85,92 @@ class ConnectorsWebApp < Sinatra::Base
...
+    source_status = connector.source_status(body_params)
     json(
-      :extractor => { :name => @connector.name },
+      :extractor => { :name => connector.name },
       :contentProvider => source_status
     )
   end
 
+  post '/start_sync' do
+    job = settings.job_store.create_job
```
wangch079, line 98: Should we reject a type if there's the same type of job for the same content source is running?

lib/connectors_async/job.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    def has_cursors?
+      @data.has_key?(:cursors)
+    end
+
+    def cursors
+      @data[:cursors]
+    end
+
+    def update_status(new_status)
```
wangch079, line 46: should validate `new_status` is a valid?

lib/app/app.rb
```diff
@@ -36,6 +37,8 @@ class ConnectorsWebApp < Sinatra::Base
     set :deactivate_auth, settings.http['deactivate_auth']
     set :connector_name, settings.http['connector']
     set :connector_class, ConnectorsSdk::Base::REGISTRY.connector_class(settings.http['connector'])
+    set :job_store, ConnectorsAsync::JobStore.new
+    set :job_runner, ConnectorsAsync::JobRunner.new({ max_threads: settings.worker['max_thread_count'] })
```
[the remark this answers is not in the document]
artem-shelkovnikov: Yeah, every introduction of a new setting will lead to this unfortunately - I don't think there's a good way to avoid problems with it :/

lib/connectors_async/job_store.rb
artem-shelkovnikov: Not yet, I decided not to put the job cleanup in this PR and handle it in a different PR (https://github.com/elastic/enterprise-search-team/issues/1980) - same goes for Token Refresh (https://github.com/elastic/enterprise-search-team/issues/1979)

lib/app/app.rb
artem-shelkovnikov: I'm not sure about this - catching it will make us return something, and the design of the endpoint states that we return fields like "status", "docs" - we'll have to change the design a bit and I'm not sure we need to - this HTTP interface is a bit rudimentary and we won't support it really soon

lib/connectors_async.rb
artem-shelkovnikov: That's just a file to require all connectors_async classes if developer writes `require connectors_async`

lib/app/app.rb
artem-shelkovnikov: We don't currently have a concept of job type, so it's hard to do - connector does not distinguish incremental job from full sync job (we can do it by checking `:modified_since` and `:cursors` when starting job.

There is another point to that - we don't have a way to cancel/restart job. So if in Enterprise Search UI or API you restart jobs, then the jobs in the connector will keep their execution (unfortunately) and won't be interrupted (they need to be recycled after completion or hanging). Therefore for now it's not safe to assume there will be single running job of each type on connector.

lib/connectors_async/job.rb
artem-shelkovnikov: Do you mean check that the variable value is valid or check that the transition to new status is possible?

lib/connectors_async/job.rb
artem-shelkovnikov: it was done with assumption that sometime in future the job has to be stored persistently somewhere, therefore `@data` hash represents future model of data for jobs.

lib/connectors_async/job.rb
artem-shelkovnikov: `job_id` is stored, `job_type` we cannot determine - it's not sent by Enterprise Search and is actually not specific to the job.

As for `content_source_id` - we can, but there was never a need to do so, so I decided to not store it per job until we have a use-case for it.

lib/app/app.rb
```diff
@@ -84,51 +85,92 @@ class ConnectorsWebApp < Sinatra::Base
...
+      :status => job.status
+    )
+  end
+
   post '/documents' do
-    results, cursors, completed = @connector.document_batch(body_params)
+    job = settings.job_store.fetch_job(body_params[:job_id])
+
+    # TODO: is it an ugly way to send cursors?
+    raise job.error if job.is_failed?
```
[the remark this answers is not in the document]
artem-shelkovnikov: Good call, I've fixed it in current state of repo (added `:errors` array returned from connectors into Enterprise Search). Does it look better?

lib/app/app.rb
artem-shelkovnikov: I'm actually not sure about it - our HTTP interface is gonna be frozen in current state and we won't need to bother much about it. Currently it will mark the job as failed with INTERNAL_SERVER_ERROR and the error can be seen in the connector - probably that's sufficient?

RossAllenBell: Based on my understanding, the pool doesn't have a task size limit. I think it will happily queue tasks and never reject a new one.

RossAllenBell (approved):
LGTM, thanks for straightening out the error API.

lib/app/app.rb
```diff
@@ -112,23 +112,26 @@ class ConnectorsWebApp < Sinatra::Base
...
+
+      if job.is_failed?
+        response[:errors] = [ job.error ]
+      else
+        response[:docs] = job.pop_batch
+        response[:cursors] = job.cursors if job.has_cursors?
+      end
+
+      json(response)
+    rescue ConnectorsAsync::JobStore::JobNotFoundError => e
```
tarekziade: you don't use `e`, maybe you can remove it?

lib/connectors_async/job_runner.rb
```diff
@@ -18,10 +18,11 @@ def initialize(max_threads:)
 
     def start_job(job:, connector_class:, cursors:, modified_since:, access_token:)
       @pool.post do
+        Time.zone = ActiveSupport::TimeZone.new('UTC') # bah Time.zone should be init for each thread
```
tarekziade, line 21: I would probably create a `init_thread` function where we can do all the initialization required (even if it's just one for now, for extra clarity

tarekziade (commented):
Looks good. I'd expect tests for these 2 new behavior fixes

lib/app/app.rb
wangch079: >catching it will make us return something

Exactly, we should tell the caller (i.e. ent-search) that we couldn't find the job ID. It's not necessary a 200 response, maybe 404. But here it will return 500 `INTERNAL SERVER ERROR` if we leave it as it is.

lib/connectors_async/job.rb
wangch079: `job_type` and `content_source_id` can be sent by `ent-search`, and can be used for reject job creation, as I mentioned above.

lib/connectors_async/job.rb
wangch079: For that, we can have a `save` method in the future to persist what we think should be persisted in the future, if needed (Not necessarily all fields in `@data` need to be persisted)

lib/connectors_async/job.rb
wangch079: Yes, we can't blindly trust the data from outside world 😄

lib/app/app.rb
artem-shelkovnikov: Yeah, good call - I've fixed this to send 404s now: https://github.com/elastic/connectors/blob/375fa82fe12b94a2876abc68ca7eda8d681723c5/lib/app/app.rb#L130

lib/connectors_async/job.rb
artem-shelkovnikov: Indeed, but since it's not possible to interrupt a running job in the `connectors` repo, it's not possible to enforce this logic without breaking the flow heavily - re-triggering jobs in Enterprise Search. I will think of a way to approach it, but probably there's no good way without major Enterprise Search refactoring (there's no way for now even to send a cancel signal to connector package

lib/connectors_async/job.rb
artem-shelkovnikov: Yeah indeed would be nice to have it in future and I expect it will change the model of jobs and everything around it significantly, so I'm not putting too much thought here for now :)

lib/app/app.rb
artem-shelkovnikov: I think it's yet technically not possible as there's a functionality of job cancellation in Enterprise Search that is unable to communicate the fact that job was cancelled to the connector package, and introducing this logic will be a major refactoring.

Since it's too much work it's not something that we'll probably be aiming at currently, therefore I think we can avoid introducing this logic for now (which is an important logic to have in future imo).

lib/connectors_async/job.rb
artem-shelkovnikov: I've added the validation in 9a70a314d64a9a80bffa92e78442849e02b4ab7c

It brings me to a future discussion about - how much input to the methods do we validate? E.g. do we validate for numbers if we take an argument that is treated like a number? Same with strings? How far should we go with this validation?

tarekziade (dismissed):

tarekziade (approved):

PR #88, opened 2022-05-05
Revert "Add Confluence Cloud connector"

wangch079:
Reverts elastic/connectors#66


artem-shelkovnikov:
Hi @wangch079 , what's the reason behind the revert for Confluence Cloud connector?

serenachou:
@Diana Jourdan ***@***.***> requested the revert due to
concerns about the work to support it. But we're considering whether that
impact @Ross Bell ***@***.***> 's BYO PR if this is reverted. Is
it better to keep it in?

wangch079:
Close it as revert is not needed any more, see discussion [here](https://elastic.slack.com/archives/C030MMRDFL3/p1651738897197369)

PR #91, opened 2022-05-06
Add sites prefix to path

wangch079:
#### Pre-Review Checklist

  - https://github.com/elastic/ent-search/pull/6521

### DIFF
--- a/lib/connectors_sdk/office365/custom_client.rb
+++ b/lib/connectors_sdk/office365/custom_client.rb
@@ -77,12 +77,11 @@ module ConnectorsSdk
         # recently created groups (new Private Team site will be there) to reduce friction and index this site
         # earlier.
         # See: https://github.com/elastic/ent-search/pull/3581
-        share_point_sites = (sites(:fields => %w[id]) + recent_share_point_group_sites(:fields => %[id]))
+        share_point_sites = (sites(:fields => %w[id,name]) + recent_share_point_group_sites(:fields => %w[id,name]))
 
         share_point_sites
-          .map(&:id)
-          .uniq
-          .map { |site_id| site_drives(site_id, :fields => fields) }
+          .uniq(&:id)
+          .map { |site| site_drives(site, :fields => fields) }
           .flatten
           .compact
       end
@@ -104,10 +103,13 @@ module ConnectorsSdk
         request_all(:endpoint => 'sites/', :fields => fields, :additional_query_params => { :search => '', :top => 10 })
       end
 
-      def site_drives(site_id, fields: [])
+      def site_drives(site, fields: [])
         document_libraries(
-          request_all(:endpoint => "sites/#{site_id}/drives/", :fields => fields)
-        )
+          request_all(:endpoint => "sites/#{site.id}/drives/", :fields => fields)
+        ).map do |drive|
+          drive.site_name = site.name
+          drive
+        end
       rescue ClientError => e
         ConnectorsShared::Logger.info("Received response of #{e.status_code} trying to get drive for Site with Id = #{site_id}: #{e.message}")
         nil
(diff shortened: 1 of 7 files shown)


RossAllenBell (dismissed):

RossAllenBell:
Oh, and I think we'll need to release these changes into Enterprise Search.

spec/connectors_sdk/sharepoint/extractor_spec.rb
```diff
@@ -58,7 +58,7 @@
     end
```
seanstory, line 58: Maybe worth adding one new test/expectation about what the `path` of these should be now? Just to protect against any regression.

seanstory (commented):
LGTM, with one comment (and Ross's comment too). @serenachou I pinged Robin with a link to this in the community slack, but any chance we should let him give this a look to confirm it addresses his ask?

wangch079:
> Oh, and I think we'll need to release these changes into Enterprise Search.

Yes will do soon and linked in this PR.

spec/connectors_sdk/sharepoint/extractor_spec.rb
wangch079: It was tested in adapter_spec. But I will add one test in this spec.

lib/connectors_sdk/office365/custom_client.rb
```diff
@@ -75,12 +75,11 @@ def share_point_drives(fields: [])
         # recently created groups (new Private Team site will be there) to reduce friction and index this site
         # earlier.
         # See: https://github.com/elastic/ent-search/pull/3581
-        share_point_sites = (sites(:fields => %w[id]) + recent_share_point_group_sites(:fields => %[id]))
+        share_point_sites = (sites(:fields => %w[id,name]) + recent_share_point_group_sites(:fields => %w[id,name]))
```
seanstory, line 78: :( based off of the failing specs [here](https://github.com/elastic/ent-search/pull/6521), I think this will require you to re-record the document lifecycle specs :(

serenachou:
Go ahead and merge, we don't have to wait for Robin, but if I can get him to respond to me, then I'm happy to show him the PR and get his feedback :)

lib/connectors_sdk/office365/custom_client.rb
wangch079: Yes, as I mentioned here https://github.com/elastic/ent-search/pull/6521#issuecomment-1120679161, will re-record after this is merged, to avoid any further changes.

lib/connectors_sdk/share_point/adapter.rb
```diff
@@ -29,18 +29,36 @@ class FileGraphItem < Office365::Adapter::FileGraphItem
         def self.convert_id_to_es_id(id)
           ConnectorsSdk::SharePoint::Adapter.share_point_id_to_es_id(id)
         end
+
+        private
+
+        def get_path(item)
+          item.site_name.present? ? "/sites/#{item.site_name}#{super}" : super
+        end
```
tarekziade, line 37: I am confused why we have to write three times the same method here. I feel like we could have that in a single place

RossAllenBell: I agree with all of the above.

tarekziade: Well, it still looks like a lot of duplication. This kind of duplication is usually a smell of a design issue. 

Why can't you use `include` btw? (multiple inheritance)

All those classes could include a `SitePrefix` class that implements `get_path` so you remove that duplication and don't have to copy over the same method everywhere.

wangch079: Introduced module `SitePrefix`

seanstory (approved):
lgtm!

seanstory (approved):

PR #96, opened 2022-05-10
Fix Connectors Require Statements

artem-shelkovnikov:
Some of development branches sporadically show broken tests. 

This happens if the order of `*_spec.rb` files changes so that requires happen in different times.

This PR aims to fix it - I've removed shared non-testing specific dependencies from `spec_helper.rb` and ran each `*_spec.rb` file individually, fixing missing `require statements`.

#### Pre-Review Checklist


tarekziade (approved):
Great idea! 

One thing we should also do is add randomness when we run tests, by adding `--order rand` in 

https://github.com/elastic/connectors/blob/main/Makefile#L9

tarekziade:
> @tarekziade While I dislike the randomness in the tests, I believe it's a good way to move forward without too much hassle - seeds of the runs will help find the problem when they happen.

Why do you dislike it? Randomness here is just about running the tests in a different order. It's very good practice.

It's a very good way to make sure we don't have broken setup/teardown. That kind of problem is hidden when you always run the tests in the same order

> 
> I will create another PR that adds `--order rand` for `make test` command.

PR #98, opened 2022-05-12
Add equivalent endpoint

wangch079:
#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/ent-search/pull/6527

### DIFF
--- a/lib/app/app.rb
+++ b/lib/app/app.rb
@@ -175,6 +175,10 @@ class ConnectorsWebApp < Sinatra::Base
     json connector.refresh(body_params)
   end
 
+  post '/secrets' do
+    json @connector.secrets(body_params)
+  end
+
   def body_params
     @body_params ||= JSON.parse(request.body.read, symbolize_names: true).with_indifferent_access
   end
--- a/lib/app/app.rb
+++ b/lib/app/app.rb
@@ -175,8 +175,10 @@ class ConnectorsWebApp < Sinatra::Base
     json connector.refresh(body_params)
   end
 
-  post '/secrets' do
-    json @connector.secrets(body_params)
+  post '/secrets/compare' do
+    connector = settings.connector_class.new
+
+    json connector.compare_secrets(body_params)
   end
 
   def body_params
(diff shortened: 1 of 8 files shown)


lib/app/app.rb
```diff
@@ -131,6 +131,10 @@ class ConnectorsWebApp < Sinatra::Base
     json @connector.refresh(body_params)
   end
 
+  post '/secrets' do
```
wangch079, line 134: The endpoint is to compare 2 secrets (connector specific, e.g. SharePoint will compare whether the 2 access tokens represent the same user) , and is so not RESTful. `POST /secrets` is a terrible REST path, please help propose a more meaningful one. Thanks @elastic/ingestion-team

tarekziade: well I guess the ultimate goal is to verify that you have an up-to-date token for a given user. Maybe `post /verify-identify` ?

wangch079: The use case here is to make sure users don't add two duplicated content sources, so secrets of the 2 content sources are compared, and it's connector specific. E.g. for SharePoint, it checks whether the 2 access tokens represent the same user.
```
client(access_token1).me == client(access_token2).me
```

Maybe rename it to `POST /compare_secrets` or `POST /secrets/compare`?

tarekziade: What happens exactly when the comparison fails? What is the next step in that case? 

I think it means that Kibana+Ent-Search is using another token. My understanding is that for a given user and backend, we should always use the same token on both sides. And the token is an identity. Having the API called `secret` seems to be an internal implementation detail, no?

seanstory: What if, instead, we proxy this endpoint to a single `/me` call, and just have Enterprise Search make 2 calls to this endpoint? Could it be `/identity` or `/identify`? This is translating in sharepoint, as you say, to `/me`. In Confluence, I think it would be `/user/current`. Basically we want to know who the authenticated user is. And we could use this for more than identity comparison - we could display this in the Enterprise Search service account details.

wangch079: @seanstory replied here: https://github.com/elastic/ent-search/pull/6527#discussion_r873883649

This is connector-specific. Calling `/me` only make sense for OAuth access token. a API key may not represent a user.

wangch079: Rename the endpoint to `POST /secrets/compare`.

seanstory (approved):

PR #99, opened 2022-05-12
trying to make permissions method similar to documents and deleted ids

mchernyavskaya:
## No issue exists for this

Small change to make the base call return something, so that it can be used with a `to_a` in specs, same as we do with the other base methods for documents and deleted ids.

I don't expect it to break anything.


artem-shelkovnikov (approved):

artem-shelkovnikov:
It would be great to add a test to surface the problem that led to introducing the change, but change looks good to me!

mchernyavskaya:
> It would be great to add a test to surface the problem that led to introducing the change, but change looks good to me!

I was trying to write a test for a new connector using subject.permissions(user_id).to_a =) to eager-load the yields. And it wasn't working because of no return value.

seanstory (approved):

lib/connectors_sdk/base/extractor.rb
```diff
@@ -115,7 +115,7 @@ def with_auth_tokens_and_retry(&block)
         end
       end
 
-      def yield_document_changes(modified_since: nil)
+      def yield_document_changes(modified_since: nil, break_after_page: false)
```
wangch079, line 118: Why add `break_after_page` here?

With async flow in place, I think @artem-shelkovnikov has reverted all the `break_after_page` changes here: https://github.com/elastic/connectors/pull/101

wangch079 (commented):
You may want to communicate to @RossAllenBell on whether you want to bring this back to `ent-search`
https://elastic.slack.com/archives/C01795T48LQ/p1652816751978499

lib/connectors_sdk/base/extractor.rb
mchernyavskaya: It was a merge oversight. Removed; thanks.

mchernyavskaya:
cc @RossAllenBell - bringing you here because of the comment from @wangch079 above.

wangch079 (approved):
>cc @RossAllenBell - bringing you here because of the comment from @wangch079 above.

@mchernyavskaya , @RossAllenBell had removed the dependency on `connectors_sdk` gem. If you want to bring this change to `ent-search`, you can do so directly by opening another PR in `ent-search` repo.

RossAllenBell:
These changes could go either way, but I certainly don't think they would hurt. Why don't we get into the practice of copying them over to `ent-search` by default, including these changes.

PR #100, opened 2022-05-16
Initial work to enable token refresh with async flow

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/1979

This PR allows to use the async connector package in multitenant mode - having multiple content sources (from different Enterprise Search installations) to work together.

Specifically, it allows to store and get secrets for multiple content sources at the same time.

#### Pre-Review Checklist

  - Corresponding PR in Enterprise Search:


lib/connectors_async/secret_storage.rb
```diff
@@ -0,0 +1,21 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+module ConnectorsAsync
+  class SecretStorage
```
tarekziade, line 8: I am wondering, what will be the difference between secret storage and storage for the connector? 

Asking this because looking at the future, I am kind of wondering if we want to encrypt some values at the *storage* level, or at the *field* level.

In other words, if we were to store configuration fields (or any session data) for a connector, do we want to treat all fields through the same key/value interface and make the encryption story a separate concern?

lib/app/app.rb
```diff
@@ -39,6 +39,7 @@ class ConnectorsWebApp < Sinatra::Base
     set :connector_class, ConnectorsSdk::Base::REGISTRY.connector_class(settings.http['connector'])
     set :job_store, ConnectorsAsync::JobStore.new
     set :job_runner, ConnectorsAsync::JobRunner.new({ max_threads: settings.worker['max_thread_count'] })
+    set :secret_storage, ConnectorsAsync::SecretStorage.new
```
RossAllenBell, line 42: Instead of using Sinatra _settings_, I think I would just change `ConnectorsAsync::SecretStorage` to be a singleton class that other logic can more easily reference without having to pass around references to the secret storage everywhere.

RossAllenBell (dismissed):
The shared auth logic looks correct. I think it'll be less cumbersome to introduce a singleton class.

This makes job-cleanup logic more important, although it'll be a little complex in that multiple jobs can access the same secret entry.

lib/connectors_async/secret_storage.rb
tarekziade: Sounds good! I have added a space time idea around these topics btw :)

lib/app/app.rb
artem-shelkovnikov: For me passing the service in the constructor is a way to control the number of dependencies in the module that's using it, and it comes from the Object Oriented world background, which might not apply here :) 

Does Sinatra has limitation about storage of things in settings? Intuitively, the fact that I use a feature called "settings" to have singletons of services that I use feel wrong indeed.

artem-shelkovnikov: I think it's a good call, but not only for secret_storage, but also for job_store and job_runner. Do you mind if we merge this PR and approach making the mentioned three classes singletons in a different PR?

RossAllenBell (approved):
> Do you mind if we merge this PR and approach making the mentioned three classes singletons in a different PR?

Yep 👍

lib/app/app.rb
wangch079: Singleton class makes sense to me for the 3 classes, and they don't have to be part of Sinatra settings. Agree that we can have it in a separate PR as it has nothing to do with the purpose of this PR

tarekziade: It depends a lot on the lifecycle of the secrets and your API signatures. 

When you start a request, will the secrets need to be updated and have to be fetched dynamically? If not, I think a secret is a static option like another and there's no real need to pass around the secret storage class instance, you can grab the value *before* and inject it in your params.

About OOP ~ I think it's not really about the case being a singleton, but rather about deciding if connectors are smart enough to go and pick secrets themselves. e.g. would it make sense to fully isolate the two concerns and make sure the connector has all it needs to function when you call it? or is this more of behavior where the connector needs to make smart decisions and pick stuff in the secrets

lib/app/app.rb
```diff
@@ -100,9 +101,12 @@ class ConnectorsWebApp < Sinatra::Base
   post '/start_sync' do
     job = settings.job_store.create_job
 
+    settings.secret_storage.store_secret(body_params[:content_source_id], { access_token: body_params[:access_token] })
```
wangch079, line 104: I think we lack a unified concern to validate request params. What if `content_source_id` or `access_token` is missing? I think it will just return 500 Internal Server Error

lib/connectors_async/secret_storage.rb
wangch079: I think this is just in-memory storage? It will make sense to encrypt the secrets if we want to persist them, which might be true for our new ingestion service.

I'm interested to know how other integrations using Agent/Fleet/Package handle data storage(yml? how to store sensitive data?). This is part of my space time topic.

artem-shelkovnikov:
> I think this is just in-memory storage? It will make sense to encrypt the secrets if we want to persist them, which might be true for our new ingestion service.
Yeah it's just a simple and straightforward abstraction to make things run, since our secrets are ephemereal. For the future indeed we need to take care of encryption and such - which I believe can be handled by storing secrets in Elasticsearch, it allows to encrypt data, and it's sharable between different workers.

lib/app/app.rb
artem-shelkovnikov: I personally like Singleton classes that encapsulate pure functions. Other usages of it produce bad surprises based on my experience.

So the JobStore and SecretStorage encapsulate the abstraction of real persistence - disk, database, or anything similar, but actually for now store stuff in-memory. Therefore, currently they are actually static per application run, but in future they can be init per request (which should be okay, they are lightweight abstractions without significant memory footprint).

JobRunner is also static, but it's an abstraction over some form of a async job processor - it can be same machine, different machine, whatever fits. Based on farther implementation of it, it can be switched to per-request with insignificant memory overhead.

artem-shelkovnikov: What I also dislike about Singleton pattern is that it makes sense to make constructors/initializers private in case of singleton for a more robust abstraction, but it makes testing of such classes more difficult and less predictable, especially doing integration testing when a class uses such singleton - we had a bunch of problems testing classes that use AppConfig in Enterprise Search - state of AppConfig was shared within the whole rspec run and tests had to be really careful setting/reverting settings of AppConfig, I find it quite a problem compared to just initializing a lightweight object every time you need to do some work with it.

PR #89, opened 2022-05-05
Blog code: Gitlab connector

mchernyavskaya:
## Related to https://github.com/elastic/enterprise-search-team/issues/1918

This is the code supporting the article about creating custom connectors (aka the `Bring Your Own Connector` feature). The code for now will stay in a draft state; I don't have plans to merge it yet. The code illustrates BYOC creating the example connector for GitLab.

#### Pre-Review Checklist
- [x] Covered the changes with automated tests
- [x] Tested the changes locally
- [x] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)

## For Elastic Internal Use Only
- [ ] Considered corresponding documentation changes to [contribute separately](https://github.com/elastic/enterprise-search-pubs#contribute-docs-changes-for-product-changes)
- [x] [Manual test steps](https://github.com/elastic/connectors/blob/main/docs/INTERNAL.md#minimal-manual-tests)  are successful

PR #104, opened 2022-05-17
Renaming HttpCallWrapper => Connector

mchernyavskaya:
## Related to https://github.com/elastic/enterprise-search-team/issues/1918

We need to rename the `HttpCallWrapper` => `Connector` to keep it in sync with business terminology.
Since the repo is public, we need to be specifically clear in naming to let devs understand what is where.

Other (small) changes:
* renaming connector `name` => `display_name` method to keep consistent with Ross's changes in [BYOC PR](https://github.com/elastic/connectors/pull/86)
* adding a default thread count = number of processors

#### Pre-Review Checklist

## Related Pull Requests

* #86 
* #89 

  - [Ent-search PR](https://github.com/elastic/ent-search/pull/6540)

### DIFF
--- a/lib/app/app.rb
+++ b/lib/app/app.rb
@@ -173,6 +173,6 @@ class ConnectorsWebApp < Sinatra::Base
   end
 
   def connector
-    settings.connector_class.new
+    @_connector ||= settings.connector_class.new
   end
 end
--- a/lib/app/app.rb
+++ b/lib/app/app.rb
@@ -78,7 +78,6 @@ class ConnectorsWebApp < Sinatra::Base
 
     # We only support Basic for now
     error = auth.provided? && auth.scheme != 'basic' ? ConnectorsShared::UNSUPPORTED_AUTH_SCHEME : ConnectorsShared::INVALID_API_KEY
-    p error
     response = { errors: [error.to_h] }.to_json
     halt(error.status_code, { 'Content-Type' => 'application/json' }, response)
   end
--- a/lib/app/app.rb
+++ b/lib/app/app.rb
@@ -11,6 +11,7 @@ require 'active_support/core_ext/time/zones'
 require 'faraday'
 require 'hashie'
 require 'json'
+require 'concurrent'
 
 require 'sinatra'
 require 'sinatra/config_file'
@@ -28,6 +29,8 @@ class ConnectorsWebApp < Sinatra::Base
   register Sinatra::ConfigFile
   config_file ConnectorsApp::CONFIG_FILE
 
+  DEFAULT_THREAD_COUNT = Concurrent.processor_count.freeze
+
   configure do
     set :raise_errors, false
     set :show_exceptions, false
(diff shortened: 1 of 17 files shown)


lib/app/app.rb
```diff
@@ -28,6 +29,8 @@ class ConnectorsWebApp < Sinatra::Base
   register Sinatra::ConfigFile
   config_file ConnectorsApp::CONFIG_FILE
 
+  DEFAULT_THREAD_COUNT = Concurrent.processor_count.freeze
```
RossAllenBell, line 32: This is probably fine, but I'm noticing the _Concurrent_ documentation leaves this method undocumented 🤷

lib/app/app.rb
```diff
@@ -71,6 +78,7 @@ class ConnectorsWebApp < Sinatra::Base
 
     # We only support Basic for now
     error = auth.provided? && auth.scheme != 'basic' ? ConnectorsShared::UNSUPPORTED_AUTH_SCHEME : ConnectorsShared::INVALID_API_KEY
+    p error
```
RossAllenBell, line 81:
```suggestion
```

lib/app/app.rb
```diff
@@ -163,18 +162,18 @@ class ConnectorsWebApp < Sinatra::Base
...
-
     json connector.refresh(body_params)
   end
 
   def body_params
     @body_params ||= JSON.parse(request.body.read, symbolize_names: true).with_indifferent_access
   end
+
+  def connector
+    settings.connector_class.new
```
RossAllenBell, line 177: We'll need to cache this in case it's referenced multiple times, like in `/status`.

```suggestion
    @_connector ||= settings.connector_class.new
```

lib/app/app.rb
tarekziade: Ah, that's interesting. The number of threads you can spin per *process* can be set much higher depending on what you do. Here, the app is I/O bound, so in the same process, you can have a lot of concurrent threads without being blocked by the GIL.  So `DEFAULT_THREAD_COUNT` can probably be set higher.

Python always had preemptive threads, but I've learned that this is something that's been added a bit later in CRuby 

https://bugfactory.io/blog/multithreading-in-the-mri-ruby-interpreter/

So if this still the current model i n `2.6.x` (no reason it's not) you have a pretty efficient context switching when there are dozains of threads running in the same process, as long as they stay I/O bound, which is our case ~ we call a backend through a socket and sit there waiting.

artem-shelkovnikov: Can we make this change (change to the config behaviour) a different PR, so that we can have a wider discussion about it?

mchernyavskaya: @artem-shelkovnikov do we have to? I mean, it's just the default. The number can be set to whatever in config. I just thought it would be a bit more convenient to have a reasonable default instead of crash the app on start when the value in config is unset, as it previously happened.

artem-shelkovnikov (approved):

artem-shelkovnikov (approved):

PR #107, opened 2022-05-19
Add a simple idle logic for the job

artem-shelkovnikov:
This change adds simple throttling logic for the connector that stops ingestion from 3rd party if Enterprise Search haven't picked up the results for a while.

Key criteria is number of not yet ingested items for the job - I've set the number to 500 and it's a bit of an arbitrary number. If job has 500 items in the ingestion queue, then it starts to idle in periods of 10 seconds until Enterprise Search picks the items up.

#### Pre-Review Checklist


RossAllenBell (commented):
I'm not really sure how I didn't notice we weren't controlling for results queue size yet. Have you considered using a [SizedQueue](https://ruby-doc.org/core-2.5.1/SizedQueue.html)? I _think_ it'll give you the behavior you're looking for without having to manage the blocking/sleeping yourself.

wangch079:
> I _think_ it'll give you the behavior you're looking for without having to manage the blocking/sleeping yourself.

From the [doc](https://ruby-doc.org/core-2.5.1/SizedQueue.html#method-i-push):
>If there is no space left in the queue, waits until space becomes available, unless `non_block` is true. If `non_block` is true, the thread isn't suspended, and `ThreadError` is raised.

I don't think suspended thread is what @artem-shelkovnikov wants:
https://github.com/elastic/connectors/blob/11c7d4536c7e8ef7069da95876e43bc818d30b2a/lib/connectors_async/job_runner.rb#L23

lib/connectors_async/job.rb
```diff
@@ -86,5 +86,9 @@ def is_finished?
 
       [ConnectorsShared::JobStatus::FINISHED, ConnectorsShared::JobStatus::FAILED].include?(status)
     end
+
+    def should_throttle?
```
RossAllenBell, line 90: I think my only recommended change would be to change this name. It implies slowing activity, but we want to completely stop activity until later.

How about:
```suggestion
    def should_wait?
```

RossAllenBell (dismissed):
I think the blocking provided by SizedQueue _is_ the type of blocking we'd want in terms of waiting, _but_...

Yes, if a job needs to be capable of timing out without an external thread explicitly interrupting the thread, then we can't rely on SizedQueue's blocking behavior. At that point I think I'd suggest _not_ using SizedQueue. If we're managing queue length and sleeping ourselves, the SizedQueue's blocking behavior becomes exceptional behavior that would only add noise if we needed to debug anything.

lib/connectors_async/job_runner.rb
```diff
@@ -51,6 +63,30 @@ def log(str)
...
+        end
+
+        log("Job #{job.id} woke up")
+      end
+
+      yield
+    end
+
+    def idle(time)
+      Kernel.sleep(time)
```
tarekziade, line 87: Do you need that prefix?

lib/connectors_async/job.rb
artem-shelkovnikov: Good call! Renamed to `should_wait?`

lib/connectors_async/job.rb
```diff
@@ -86,5 +86,9 @@ def is_finished?
 
       [ConnectorsShared::JobStatus::FINISHED, ConnectorsShared::JobStatus::FAILED].include?(status)
     end
+
+    def should_wait?
+      @data[:documents].length > 500
```
seanstory, line 91: Let's pull this `500` out into `ConnectorsShared::Constants` or configuration or something, just to make it easier to find and update in the future.

lib/connectors_async/job_runner.rb
```diff
@@ -51,6 +63,30 @@ def log(str)
 
     private
 
+    def with_throttling(job)
+      attempts = 0
+      if job.should_wait?
+        log("Job #{job.id} is sleeping: Enterprise Search haven't picked up documents for a while.")
```
seanstory, line 69:
```suggestion
        log("Job #{job.id} is sleeping: Enterprise Search hasn't picked up documents for a while.")
```

seanstory (approved):
small comments

lib/connectors_async/job_runner.rb
```diff
@@ -51,6 +63,30 @@ def log(str)
 
     private
 
+    def with_throttling(job)
+      attempts = 0
+      if job.should_wait?
```
wangch079, line 68: Is the `if` only used for logging?

artem-shelkovnikov: Yeah, so far it's only for logging - I think the information the logs provide is quite valuable

lib/connectors_async/job_runner.rb
```diff
@@ -51,6 +62,30 @@ def log(str)
...
+          end
+        end
+
+        log("Job #{job.id} woke up")
+      end
+
+      yield
+    end
+
+    def idle(time)
```
tarekziade, line 85: why the alias ? can't you just use `sleep` ?

spec/connectors_async/job_runner_spec.rb
```diff
@@ -173,6 +176,45 @@
         idle_a_bit
       end
     end
+
```
tarekziade, line 179: I don't see `JobStuckError` in the tests.

tarekziade (commented):
Pretty cool! just two nits

lib/connectors_async/job_runner.rb
artem-shelkovnikov: Oh, I just forgot a logger here :)

spec/connectors_async/job_runner_spec.rb
artem-shelkovnikov: Good call, lemme add a test for it

lib/connectors_async/job_runner.rb
tarekziade: mmm.. it still strikes me as overkill to create a whole method just to call `sleep`

lib/connectors_async/job_runner.rb
```diff
@@ -8,14 +8,22 @@
...
   class JobRunner
+    class JobStuckError < StandardError; end
+
     def initialize(max_threads:)
-      @pool = Concurrent::ThreadPoolExecutor.new(min_threads: 1, max_threads: max_threads, max_queue: 0)
+      @pool = Concurrent::ThreadPoolExecutor.new(
+        min_threads: 1,
+        max_threads: max_threads,
+        max_queue: 0,
+        idletime: ConnectorsShared::Constants::IDLE_SLEEP_TIME + 1 # we +1 just so that thread.sleep manages to finish by the idle timeout
```
RossAllenBell, line 25: Let's make the unit explicit:
```suggestion
        idletime: ConnectorsShared::Constants::IDLE_SLEEP_TIME + 1.seconds # we +1 just so that thread.sleep manages to finish by the idle timeout
```

lib/connectors_shared/constants.rb
```diff
@@ -10,5 +10,15 @@ class Constants
...
+
+    # The following section reads as following:
+    # The job will extract documents until the job queue size will reach
+    # JOB_QUEUE_SIZE_IDLE_THRESHOLD items. After that, the job will attempt to sleep
+    # for IDLE_SLEEP_TIME seconds and check the queue size again. If the queue is still
+    # full, it will sleep for maximum MAX_IDDLE_ATTEMPTS times, and if the queue is still
+    # full, then job will be terminated.
+    JOB_QUEUE_SIZE_IDLE_THRESHOLD = 500 # How many documents the job queue stores until it sleeps
+    IDLE_SLEEP_TIME = 10 # For how long job queue will sleep before checking the queue size again
+    MAX_IDLE_ATTEMPTS = 30 # How many consecutive times job will try to sleep until it's destroyed
```
RossAllenBell, line 22: I think this should be reflected as a unit of time, not a number of attempts. I also think we should go with a high limit just to be safe:
```suggestion
    MAX_IDLE_TIME = 10.minutes # How long a job will try to sleep until it's destroyed
```

I know this will require some minor changes to where this constant is used.

RossAllenBell (approved):
LGTM with some minor changes.

artem-shelkovnikov:
Thank you for review, @RossAllenBell, I will merge this PR and address the comments in the next one!

artem-shelkovnikov:
Follow up PR: https://github.com/elastic/connectors/pull/117

PR #109, opened 2022-05-19
Minimal changes to convert to JRuby

seanstory:
## Relates to https://github.com/elastic/enterprise-search-team/issues/2033

This is related to the recent realization that Crawler uses JRuby, but connectors does not. I wanted to investigate how much work it was to move the repo to JRuby. At this time - not hard.

This is not intended to be merged right now, but is just to illustrate what the delta is.

#### Pre-Review Checklist


Gemfile
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-ruby File.read(File.join(__dir__, '.ruby-version')).strip
```
RossAllenBell, line 3: Was this breaking something?

seanstory: yep. It doesn't like having the `jruby-` prefix. Also, `9.3.0.0` is not a valid ruby version. We'd really need the syntax of:
`ruby '2.6.8', :engine => 'jruby', :engine_version => '9.3.0.0'`, and IDK where we'd get the `2.6.8` from, unless we add another file? Also, ent-search doesn't specify a version, so I figured it wasn't super unsafe to remove it for a POC/draft? But yeah worth talking about - what's the right way to do this, and do we need to specify a ruby version do you think? Are we using specific modern ruby features?

artem-shelkovnikov:
@seanstory do you want to pursue this PR, or should we just drop it?

PR #112, opened 2022-05-20
Add Configuration File Validation

artem-shelkovnikov:
This change adds validation to the configuration files that are used to run the connector. Using the gem `config`, we define a schema for the config file that we're planning to load and then validate the contents of the file against the schema. If validation fails, app is terminated and an error is shown to the user.

#### Pre-Review Checklist

## Release Note

Now configuration files are validated against existing schema and errors are shown if configuration file is invalid.


artem-shelkovnikov:
Example output:
```
artemshelkovnikov@Artems-MBP connectors % make run
"yq" e ".revision = \"e3f58a5996f2c8d525f2d896ad0d4a3c00bcd3f0\"" -i config/connectors.yml
"yq" e ".repository = \"git@github.com:elastic/connectors.git\"" -i config/connectors.yml
"yq" e ".version = \"8.3.0.0-20220520T151127Z\"" -i config/connectors.yml
cd lib/app; bundle _2.3.10_ exec rackup config.ru
Parsing /Users/artemshelkovnikov/git_tree/connectors/lib/connectors_app/../../config/connectors.yml configuration file.
bundler: failed to load command: rackup (/Users/artemshelkovnikov/git_tree/connectors/vendor/bundle/ruby/2.6.0/bin/rackup)
/Users/artemshelkovnikov/git_tree/connectors/vendor/bundle/ruby/2.6.0/gems/config-4.0.0/lib/config/validation/validate.rb:20:in `validate_using!': Config validation failed: (Config::Validation::Error)

  worker: is missing
	from /Users/artemshelkovnikov/git_tree/connectors/vendor/bundle/ruby/2.6.0/gems/config-4.0.0/lib/config/validation/validate.rb:8:in `validate!'
	from /Users/artemshelkovnikov/git_tree/connectors/vendor/bundle/ruby/2.6.0/gems/config-4.0.0/lib/config/options.rb:58:in `reload!'
```

artem-shelkovnikov:
Another example message (I've put string into `http.port`:

```
Parsing /Users/artemshelkovnikov/git_tree/connectors/lib/connectors_app/../../config/connectors.yml configuration file.
bundler: failed to load command: rackup (/Users/artemshelkovnikov/git_tree/connectors/vendor/bundle/ruby/2.6.0/bin/rackup)
/Users/artemshelkovnikov/git_tree/connectors/vendor/bundle/ruby/2.6.0/gems/config-4.0.0/lib/config/validation/validate.rb:20:in `validate_using!': Config validation failed: (Config::Validation::Error)

  http.port: must be an integer
```

artem-shelkovnikov:
@sphilipse tagging you to take a look at the messages too.

For some reason when I was catching the `Config::Validation::Error` and re-throwing a new error with a message that tells which file triggered the error, two errors were thrown, I'll be happy to get help on this one.

For simplicity, I just added a `puts` before the validation that says:

```
Parsing /Users/artemshelkovnikov/git_tree/connectors/lib/connectors_app/../../config/connectors.yml configuration file.
```

lib/app/app.rb
```diff
@@ -16,32 +16,30 @@
...
+  set :connector_name, ConnectorsApp::Config.http['connector']
   set :connector_class, ConnectorsSdk::Base::REGISTRY.connector_class(settings.connector_name)
   set :job_store, ConnectorsAsync::JobStore.new
-  set :job_runner, ConnectorsAsync::JobRunner.new({ max_threads: settings.worker['max_thread_count'] })
+  set :job_runner, ConnectorsAsync::JobRunner.new({ max_threads: ConnectorsApp::Config.worker['max_thread_count'] })
   set :secret_storage, ConnectorsAsync::SecretStorage.new
 
   error do
     e = env['sinatra.error']
+    puts e.message
```
seanstory, line 42: Should this be a `logger.error`? Or is this lingering?

lib/connectors_app/config.rb
```diff
@@ -3,14 +3,42 @@
...
-require 'yaml'
+
+require 'config'
+
+# We look for places in this order:
+# - CONNECTORS_CONFIG environment variable
+# - here: /../../config/connectors.yml
+CONFIG_FILE = ENV['CONNECTORS_CONFIG'] || File.join(__dir__, '../..', 'config', 'connectors.yml')
+
+puts "Parsing #{CONFIG_FILE} configuration file."
```
seanstory, line 14: Should use a logger?

seanstory (commented):
ent-search uses `json-schema`.

https://github.com/elastic/ent-search/blob/97777bea9ddfd395f58bfe5ae3525610001021b7/Gemfile#L72-L73

Any reason in particular to use `config` instead? It seems to pull in a lot of transitive dependencies for just verifying the presence of a few fields.

wangch079 (dismissed):
Overall LGTM, I think it's good to merge after addressing comments from Sean.

Just wondering, where is the [register Config](https://github.com/rubyconfig/config/#installing-on-sinatra)?

Regarding Sean's comment on `json-schema`, I don't think `connectors` (a public repo) has to follow `ent-search` (a private repo). After reading more, I feel `json-schema` is more robust and feature-rich, but `config` serves our needs. I don't hold strong preference here, but would like to hear options from the team.

artem-shelkovnikov:
> Any reason in particular to use config instead? 
No specific reason, it was the first framework I found, it fit the needs with a bit of tinkering, so I took it. I'll take a look at the alternatives now

> It seems to pull in a lot of transitive dependencies for just verifying the presence of a few fields.
The way `dry-` structure their modules make it look like we're dragging in a large thing, but each package if of micro size containing just several small files. While it looks huge in the Gemfile.lock, the dependency is rather small and non-invasive IMO.

I'll take a look at `json-schema` to see how it is.

> Just wondering, where is the [register Config](https://github.com/rubyconfig/config/#installing-on-sinatra)?
I didn't manage to make it work with register Config - it expects the config files in specific folders with specific names, as far as I could understand, for example `config/settings.yml`, `config/settings.local.yml`, `config/settings/development.yml`. Instead I do the load manually here:

https://github.com/elastic/connectors/pull/112/files#diff-28eee624130fc5496c61f7628aa09b0083b8a90c36f76ac53dc4ed6b0a306e08R38

artem-shelkovnikov:
Also looking at the code now with fresh head, I think I barely use the `config` module itself, I mostly use the validation features from `dry-` modules, I might simplify the code and dependencies removing the `config` module and leaving only the validation ones.

artem-shelkovnikov:
About usage of `json-schema` - I tried to use it but it's not very convenient - you have to generate huge verbose schemas that feel more like schemas for Web API, not for the config file.

Also the output for the validation errors is less readable - the delimiters that are used (#/) feel weird because our file is a dot-separated yaml.

I also tried to use `dry-validation` alone, but it still brings a ton of dependencies (whole dry-rb package pretty much).

I'm leaning to just using this module now to have some sort of validation and error output, and then investigate and switch to something if/when it fits us better

lib/connectors_app/config.rb
```diff
@@ -3,14 +3,42 @@
...
+
+Config.setup do |config|
```
seanstory, line 16: Lets make this `::Config`, so that it's really explicit when we use `::Config` vs `ConnectorsApp::Config` in this file.

lib/connectors_app/config.rb
```diff
@@ -3,14 +3,42 @@
...
 
 module ConnectorsApp
-  # We look for places in this order:
-  # - CONNECTORS_CONFIG environment variable
-  # - here: /../../config/connectors.yml
-  CONFIG_FILE = ENV['CONNECTORS_CONFIG'] || File.join(__dir__, '../..', 'config', 'connectors.yml')
   DEFAULT_PASSWORD = 'changeme'
 
-  Config = YAML.load_file(CONFIG_FILE)
+  Config = Settings
```
seanstory, line 43: Where is `Settings` coming from? is that pulled in from the `require 'config'`? 
Same as comment above, using `::Settings` might help make clear that this isn't part of `ConnectorsApp`.

seanstory (dismissed):
Thanks for explaining the rational and answering all my previous questions, @artem-shelkovnikov! LGTM. Two small comments, but feel free to ignore if you disagree at all.

artem-shelkovnikov:
@seanstory can I have your :stamp: again, I've rebased `main` into the branch, added `::` before `Config` and `Settings` variables and have a plan for development of this code in future :)

seanstory (approved):
LGTM! Thanks!

PR #113, opened 2022-05-20
Make it possible to switch to different connector log level

artem-shelkovnikov:
This change allows to change log level of the connector (default is `info` if not specified).

#### Pre-Review Checklist

## Release Note

Now it's possible to change logging level by specifying `log_level: <level>` in configuration file.

### DIFF
--- a/lib/app/initializers/02_log_level.rb
+++ b/lib/app/initializers/02_log_level.rb
@@ -5,6 +5,10 @@
 # you may not use this file except in compliance with the Elastic License.
 #
 
+require 'stubs/app_config'
+require 'connectors_app/config'
+require 'connectors_shared/logger'
+
 logger = AppConfig.connectors_logger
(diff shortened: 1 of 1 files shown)


lib/app/initializers/02_log_level.rb
```diff
@@ -0,0 +1,4 @@
+logger = AppConfig.connectors_logger
```
seanstory, line 1: needs license header and requires

seanstory (approved):
This will be nice to have.

wangch079 (dismissed):

lib/app/initializers/02_log_level.rb
artem-shelkovnikov: Good call, I wonder why pipeline is green when I don't have license headers :hmm:

mchernyavskaya (approved):
LGTM

PR #114, opened 2022-05-20
Add proper logging into JobRunner

artem-shelkovnikov:
Just cleaning up some tech debt - before I was using `puts` to output logging info, this PR changes it to use `ConnectorsShared::Logger` instead.

Ideally, the method `log_with_thread_id` should go away and we should log thread ids when running jobs natively for all code running within threads, but that's work for later.

### DIFF
--- a/lib/connectors_async/job_runner.rb
+++ b/lib/connectors_async/job_runner.rb
@@ -9,6 +9,7 @@ require 'concurrent'
 require 'connectors_app/config'
 require 'connectors_async/job_store'
 require 'connectors_shared/job_status'
+require 'connectors_shared/exception_tracking'
 require 'connectors_shared/logger'
 
 module ConnectorsAsync
@@ -26,7 +27,7 @@ module ConnectorsAsync
         cursors = params[:cursors] ||= {}
         cursors[:modified_since] = params.delete(:modified_since) if params[:modified_since]
 
-        log_with_thread_id("Running the job #{job.id}")
+        log_with_thread_id(:info, "Running the job #{job.id}")
 
         job.update_status(ConnectorsShared::JobStatus::RUNNING)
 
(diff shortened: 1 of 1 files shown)


lib/connectors_async/job_runner.rb
```diff
@@ -37,16 +37,15 @@ def start_job(job:, connector_class:, secret_storage:, params:)
...
-        log("Job #{job.id} failed: #{e.message}")
+        log_with_thread_id("Job #{job.id} failed: #{e.message}")
       end
     end
 
-    def log(str)
-      # TODO: use proper logging
-      puts("[#{Time.now.to_i}] [Thread #{Thread.current.object_id}] #{str}")
+    def log_with_thread_id(str)
+      ConnectorsShared::Logger.debug("[#{Time.now.to_i}] [Thread #{Thread.current.object_id}] #{str}")
```
seanstory, line 48:
```suggestion
      ConnectorsShared::Logger.debug("[Thread #{Thread.current.name}] #{str}")
```

The logger will include the timestamp for you, and `name` will be more meaningful (main thread will be `"main"`)

lib/connectors_async/job_runner.rb
```diff
@@ -37,16 +37,15 @@ def start_job(job:, connector_class:, secret_storage:, params:)
         job.update_status(ConnectorsShared::JobStatus::FINISHED)
         job.update_cursors(new_cursors)
 
-        log("Job #{job.id} has finished successfully")
+        log_with_thread_id("Job #{job.id} has finished successfully")
       rescue StandardError => e
         job.fail(e)
-        log("Job #{job.id} failed: #{e.message}")
+        log_with_thread_id("Job #{job.id} failed: #{e.message}")
```
seanstory, line 43: Not really related to this change, but we should use ExceptionTracking here to get the stack trace.

seanstory (approved):
small comments

lib/connectors_async/job_runner.rb
wangch079, line 48: Do you think we should let the caller specify the log level? The first 2 calls are more appropriate to `info` and apparently, the last one is `error`

artem-shelkovnikov: Good call, I changed the implementation accordingly

mchernyavskaya:
@artem-shelkovnikov there seems still to be 2 suggestions from Sean - are you getting them in? I could approve after.

mchernyavskaya (approved):
LGTM

mchernyavskaya:
@artem-shelkovnikov actually, I'd rather define new helper methods for each level (with thread ID) on `ConnectorsShared::Logger`, but it might be out of scope for the goal here.

mchernyavskaya (approved):

PR #117, opened 2022-05-23
Refactor job cleanup logic for better readability/usability

artem-shelkovnikov:
## Follow up for https://github.com/elastic/connectors/pull/107

Addressed points:

- When duration is specified as a constant, add explicit duration prefix, e.g. `10.seconds`
- Change MAX_IDLE_ATTEMPTS -> MAX_IDLE_TIME and change logic accordingly

### DIFF
--- a/lib/connectors_async/job_runner.rb
+++ b/lib/connectors_async/job_runner.rb
@@ -4,6 +4,7 @@
 # you may not use this file except in compliance with the Elastic License.
 #
 
+require 'active_support/core_ext/numeric/time'
 require 'concurrent'
 
 require 'connectors_app/config'
@@ -22,7 +23,7 @@ module ConnectorsAsync
         min_threads: 1,
         max_threads: max_threads,
         max_queue: 0,
-        idletime: ConnectorsShared::Constants::IDLE_SLEEP_TIME + 1 # we +1 just so that thread.sleep manages to finish by the idle timeout
+        idletime: ConnectorsShared::Constants::IDLE_SLEEP_TIME + 1.second # we +1 just so that thread.sleep manages to finish by the idle timeout
       )
     end
 
@@ -63,16 +64,15 @@ module ConnectorsAsync
     private
 
     def with_throttling(job)
-      attempts = 0
       if job.should_wait?
         log_with_thread_id(:info, "Job #{job.id} is sleeping: Enterprise Search hasn't picked up documents for a while.")
-
+        wait_duration = 0
         while job.should_wait?
-          if attempts < ConnectorsShared::Constants::MAX_IDLE_ATTEMPTS
-            attempts += 1
+          if wait_duration < ConnectorsShared::Constants::MAX_IDLE_TIME
             idle(ConnectorsShared::Constants::IDLE_SLEEP_TIME)
+            wait_duration += ConnectorsShared::Constants::IDLE_SLEEP_TIME
           else
-            raise JobStuckError.new("Enterprise Search failed to collect the data from the queue, waited #{attempts} times for #{ConnectorsShared::Constants::IDLE_SLEEP_TIME} seconds.")
+            raise JobStuckError.new("Enterprise Search failed to collect the data from the queue, waited for #{ConnectorsShared::Constants::MAX_IDLE_TIME} seconds.")
           end
         end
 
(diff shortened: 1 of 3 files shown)


lib/connectors_async/job_runner.rb
```diff
@@ -63,16 +64,15 @@ def log_with_thread_id(level, str)
...
       if job.should_wait?
         log_with_thread_id(:info, "Job #{job.id} is sleeping: Enterprise Search hasn't picked up documents for a while.")
-
+        wait_duration = 0
         while job.should_wait?
-          if attempts < ConnectorsShared::Constants::MAX_IDLE_ATTEMPTS
-            attempts += 1
+          if wait_duration < ConnectorsShared::Constants::MAX_IDLE_TIME
             idle(ConnectorsShared::Constants::IDLE_SLEEP_TIME)
+            wait_duration += ConnectorsShared::Constants::IDLE_SLEEP_TIME
```
artem-shelkovnikov, line 73: Looks a bit awkward, but I believe it's fine - we can actually calculate the idling duration, which will add some milliseconds, but on a high level it should work, unless `sleep` actually sleeps for much more than expected - can happen under heavy load.

Should we address it, team?

lib/connectors_async/job_runner.rb
```diff
@@ -63,16 +64,15 @@ def log_with_thread_id(level, str)
...
     def with_throttling(job)
-      attempts = 0
       if job.should_wait?
         log_with_thread_id(:info, "Job #{job.id} is sleeping: Enterprise Search hasn't picked up documents for a while.")
-
+        wait_duration = 0
         while job.should_wait?
-          if attempts < ConnectorsShared::Constants::MAX_IDLE_ATTEMPTS
-            attempts += 1
+          if wait_duration < ConnectorsShared::Constants::MAX_IDLE_TIME
```
artem-shelkovnikov, line 71: Do we want to be precise about it?

E.g. if setting MAX_IDLE_TIME = 10.seconds and IDLE_SLEEP_TIME = 3.seconds, we will actually sleep for 12.seconds.

Usability-wise it's not a big problem, yet something that might not be immediately expected.

artem-shelkovnikov:
Need to merge https://github.com/elastic/connectors/pull/110 before merging this one

lib/connectors_shared/constants.rb
```diff
@@ -18,7 +20,7 @@ class Constants
     # full, it will sleep for maximum MAX_IDDLE_ATTEMPTS times, and if the queue is still
     # full, then job will be terminated.
     JOB_QUEUE_SIZE_IDLE_THRESHOLD = 500 # How many documents the job queue stores until it sleeps
-    IDLE_SLEEP_TIME = 10 # For how long job queue will sleep before checking the queue size again
-    MAX_IDLE_ATTEMPTS = 30 # How many consecutive times job will try to sleep until it's destroyed
+    IDLE_SLEEP_TIME = 10.seconds # For how long job queue will sleep before checking the queue size again
+    MAX_IDLE_TIME = 300.seconds # How long a job will try to sleep until it's destroyed
```
RossAllenBell, line 24: Make the limit higher:
```suggestion
    MAX_IDLE_TIME = 600.seconds # How long a job will try to sleep until it's destroyed
```

Or:
```suggestion
    MAX_IDLE_TIME = 10.minutes # How long a job will try to sleep until it's destroyed
```

lib/connectors_async/job_runner.rb
RossAllenBell: I think you can just record the idle start time. Then compare `Time.now - idle_start_time` with `ConnectorsShared::Constants::MAX_IDLE_TIME`. Assuming `ConnectorsShared::Constants::IDLE_SLEEP_TIME` in time passed is inaccurate, as you note.

wangch079: `Time.now - idle_start_time` makes more sense.

mchernyavskaya:
@artem-shelkovnikov is it still relevant for `main`? Does it need to be also ported to `8.3`?

artem-shelkovnikov:
@mchernyavskaya I think we don't really need this PR, I'll close it. It was more of a nice-to-have thing

PR #118, opened 2022-05-24
Enable api keys for confluence cloud

seanstory:
Part of spacetime, and to enable https://github.com/seanstory/logstash-input-confluence-cloud
This just copied a bit of work from @mchernyavskaya's work, but this doesn't attempt to expose API Key usage through the web app, but only at the SDK level.

Also fixes the fact that our gemspec didn't include the full list of runtime dependencies in our gemfile, which made using the resulting gem in a non-ent-search environment impossible.

Finally, this creates a gemspec for our "stubs", so that non-ent-search consumers of our SDK can access the stubs.

#### Pre-Review Checklist

- [x] Security-related changes (adding [Confluence Cloud API Key](https://support.atlassian.com/atlassian-account/docs/manage-api-tokens-for-your-atlassian-account/) support)

## Release Note

Added SDK support for Confluence Cloud to use API Keys

- [x] Considered corresponding documentation changes to [contribute separately](https://github.com/elastic/enterprise-search-pubs#contribute-docs-changes-for-product-changes) (N.A)


connectors_sdk.gemspec
```diff
@@ -27,4 +27,21 @@ Gem::Specification.new do |s|
   s.add_dependency 'activesupport'
   s.add_dependency 'bson'
   s.add_dependency 'mime-types'
+  s.add_dependency 'tzinfo-data'
+  s.add_dependency 'nokogiri'
+
+  # Dependencies for the HTTP service
+  s.add_dependency 'sinatra'
```
RossAllenBell, line 34: Does it make sense that the _gem_ relies on Sinatra (and the other deps here)? We shouldn't actually be including any of the web app code in the gem.

lib/connectors_sdk/atlassian/custom_client.rb
```diff
@@ -30,10 +31,11 @@ class ContentConvertibleError < ClientError; end
 
       MEDIA_API_BASE_URL = 'https://api.media.atlassian.com'
 
-      attr_reader :base_url, :access_token
+      attr_reader :base_url, :access_token, :basic_auth_token
 
-      def initialize(base_url:, access_token:, ensure_fresh_auth: nil)
+      def initialize(base_url:, access_token:, basic_auth_token: nil, ensure_fresh_auth: nil)
```
RossAllenBell, line 36:
```suggestion
      def initialize(base_url:, access_token: nil, basic_auth_token: nil, ensure_fresh_auth: nil)
        raise 'Either access_token or basic_auth_token must be provided' unless access_token.present? || basic_auth_token.present?
```

connectors_sdk.gemspec
seanstory: You're probably right. After 3 in a row of, "ugh, we use that too?" I just added everything to unblock myself. But I'll pair this list back down.

seanstory: updated, I was able to remove a few, but not many.

seanstory: I could get rid of signet if we pulled all the `authorization.rb` files into the `connectors_app/` dir. But IDK if we want to go through so much restructuring.

seanstory:
@elastic/ingestion-team can I get a re-review, since Ross is out now?

artem-shelkovnikov (approved):

PR #127, opened 2022-06-22
Adjust namespaces: connectors_sdk -> connectors; connectors_shared -> utility

artem-shelkovnikov:
Just a refactoring based on a Slack thread.

`connectors_sdk` becomes just `connectors` and encapsulates logic that's specific to connectors themselves - extraction, mapping, etc
`connectors_shared` just becomes `utility` namespace containing misc utility classes.

This PR only changes namespaces and class locations, no actual logic updates/class redistribution.


mchernyavskaya (approved):
LGTM pending the build fix

wangch079 (changes requested):
I checked out your branch and searched `connectors_sdk` and `connectors_shared` and there are still some results, mainly in `connectors_sdk.gemspec` and MD file. I think this is also why it failed the CI.

wangch079 (approved):
LGTM, I just made one change, rename dirs `spec/connectors_sdk` and `spec/connectors_shared` to `spec/connectors` and `spec/utility`

PR #128, opened 2022-06-23
Remove the gem building logic from the repository

artem-shelkovnikov:
We don't seem to need to build gems in close future, and if we do, they will be quite different from the current ones.

For simplicity of the process, removing parts that are not needed.


artem-shelkovnikov:
Closing as I won't address it for now

PR #129, opened 2022-06-23
Remove connectors_async and tests

artem-shelkovnikov:
Seems like we don't need anything in connectors_async for now, and if we do it will use different abstractions for async work, therefore cleaning the code up.


wangch079 (approved):
I was about to remove it from my PR. As we are writing directly to Elasticsearch, I don't see a queue is necessary here.

PR #130, opened 2022-06-24
Stub connector

wangch079:
#### Pre-Review Checklist


lib/app/connector.rb
```diff
@@ -0,0 +1,145 @@
...
+      end
+
+      def start_polling_jobs
+        loop do
+          polling_jobs
+        rescue StandardError => e
+          Utility::Logger.error("Error happened during sync. error: #{e.message}")
+          raise
+        ensure
+          sleep(60)
```
tarekziade, line 56: could we use a constant ? `POLL_IDLING` ?

lib/app/connector.rb
```diff
@@ -0,0 +1,145 @@
...
+      def running!
+        raise 'The connector app is already running!' unless running.make_true
+      end
+
+      def start_polling_jobs
+        loop do
+          polling_jobs
+        rescue StandardError => e
+          Utility::Logger.error("Error happened during sync. error: #{e.message}")
+          raise
```
tarekziade, line 54: what happens in that case, do we die?

tarekziade (commented):
Looks promising!

lib/app/connector.rb
wangch079: No, it's supposed to just log the error and go to the next loop. I added this for testing purposes, but forgot to remove it 😓

mchernyavskaya (approved):
LGTM - let's get in in ASAP

lib/app/connector.rb
```diff
@@ -0,0 +1,145 @@
...
+        return false unless connector['_source']['scheduling']['enabled']
+        return true if connector['_source']['sync_now']
+
+        last_synced = connector['_source']['last_synced']
+        return true if last_synced.nil? || last_synced.empty?
+
+        last_synced = Time.parse(last_synced).utc
+        sync_interval = connector['_source']['scheduling']['interval']
+        cron_parser = cron_parser(sync_interval)
+        cron_parser && cron_parser.next(last_synced) < Time.now
```
tarekziade, line 101: Should this be `Time.now.utc` ?

wangch079: The comparison will take timezone into account

PR #132, opened 2022-06-24
[Epic]: Custom Connector Implementations in the Connector Service #2128

mchernyavskaya:
## Closes https://github.com/elastic/enterprise-search-team/issues/2128

Make StubConnector and Gitlab connector functional on the new architecture. These should be used to thoroughly test the service and also to showcase the changes in this release. The other custom connectors stay frozen in this release, and should not be included to avoid confusion.

## Acceptance criteria

- [x] The updated connector framework should reflect in the StubConnector and GitLab implementations
- [x] They should be able to perform sync operations.

#### Pre-Review Checklist


Makefile
```diff
@@ -61,4 +61,7 @@ run-docker:
 exec_app:
 	cd lib/app; bundle _$(shell cat .bundler-version)_ exec ruby console_app.rb
 
+exec_command:
```
tarekziade, line 64: add it in phony

lib/app/console_app.rb
```diff
@@ -9,9 +9,76 @@
...
+      registry.connector(connectors[order], params)
+    end
+
+    def registry
+      @registry = Connectors::Base::REGISTRY
+    end
+
+    puts 'Hello Connectors 3.0!'
+
+    while true
```
tarekziade, line 67: maybe catch Interrupt? (for ctrl+c)

lib/app/console_app.rb
```diff
@@ -9,9 +9,76 @@
...
+      command = read_command
+      if command == :sync
```
tarekziade, line 69: looks like a good candidate for a `case` statement

lib/connectors/base/registry.rb
```diff
@@ -20,6 +20,18 @@ def register(name, klass)
       def connector_class(name)
         @connectors[name]
       end
+
+      def connector(name, params = nil)
+        connector_class = connector_class(name)
+        if connector_class.present?
+          return params.present? ? connector_class.new(params) : connector_class.new
+        end
+        raise "Connector #{name} is not yet registered. You need to register it before use"
```
tarekziade, line 29: you could probably put this raise at the beginning of this function, ending the line with `unless connector_class.present?`

lib/connectors/base/registry.rb
```diff
@@ -20,6 +20,18 @@ def register(name, klass)
       def connector_class(name)
         @connectors[name]
       end
+
+      def connector(name, params = nil)
+        connector_class = connector_class(name)
```
tarekziade, line 25: you are shadowing `connector_class`, use another variable name (`klass` maybe)

lib/connectors/gitlab/adapter.rb
```diff
@@ -17,24 +18,30 @@ class Adapter < Connectors::Base::Adapter
...
-                          nil
-                        else
-                          source_doc[:namespace][:name]
-                        end
-        }
-        if source_doc[:_allow_permissions].present?
-          result[:_allow_permissions] = source_doc[:_allow_permissions]
+        source_doc = source_doc.with_indifferent_access
+        result = {}
+        case type.to_sym
```
tarekziade, line 23: don't use a `case` if you only have two options here

lib/connectors/gitlab/connector.rb
```diff
@@ -5,66 +5,106 @@
...
 
 module Connectors
   module GitLab
     class Connector < Connectors::Base::Connector
       SERVICE_TYPE = 'gitlab'
 
+      def initialize(params)
+        super()
+        @extractor = Connectors::GitLab::Extractor.new
+        @sinks =
```
tarekziade, line 24: can fit in one line

lib/utility/sink.rb
```diff
@@ -0,0 +1,132 @@
...
+        puts "=============================================================="
+      end
+
+      private
+
+      def send(documents)
+        return if documents.empty?
+
+        bulk_request = documents.map do |document|
+          { index:  { _index: index_name, _id: document[:id], data: document } }
```
wangch079, line 119: It doesn't necessarily all be `index`. Bulk support other operations (e.g. `DELETE`), and we do need to delete documents sometimes.

Consider having the `operation` as part of the `document` when calling `ingest`?

Makefile
```diff
@@ -61,4 +61,7 @@ run-docker:
...
+	cd lib/app; bundle _$(shell cat .bundler-version)_ exec ruby console_app.rb $1
```
wangch079, line 65: Trying to understand `exec_command`. Is this the CLI we talked about? So the expectation of the CLI is to provide an `index_name`, select a connector, and start syncing. It bypasses the regular syncing mechanism, and we won't know that a sync has been run from `.elastic-connectors`.

Makefile
```diff
@@ -61,4 +61,7 @@ run-docker:
 exec_app:
 	cd lib/app; bundle _$(shell cat .bundler-version)_ exec ruby console_app.rb
```
wangch079, line 62: I guess it makes sense to have `console_app.rb` dedicated for the CLI. I will also change in my PR to let `exec_app` point to `connector.rb`

Makefile
mchernyavskaya: Done.

lib/app/console_app.rb
mchernyavskaya: done

lib/app/console_app.rb
mchernyavskaya: yeah there were just 2 commands initially, so...
I've redone as case now.

lib/connectors/base/registry.rb
mchernyavskaya: kk

lib/connectors/gitlab/connector.rb
mchernyavskaya: yeah but what if I want more sinks?

mchernyavskaya: kk

lib/utility/sink.rb
mchernyavskaya: I'd rather have a separate method for each operation TBH.

Makefile
mchernyavskaya: It's not used (yet). I was thinking of passing the commands from the command line, then I did it by gets, so this one is kinda extra. We might decide on another approach. I just needed something to run the sync with, some CTA, and this was what came to mind first but might or might not be used later. I guess that for scripting, it does make sense to be able to run the commands with params from command line; for user interactions aka manual mode, `gets` could still come in handy. But those are just different options.

artem-shelkovnikov (approved):
Great PR to move forward!

lib/connectors/gitlab/connector.rb
```diff
@@ -5,66 +5,101 @@
...
 
       def configurable_fields
-        {
-          'api_token' => {
-            'label' => 'API Token',
-            'value' => nil
+        @configurable_fields ||= {
+          :api_token => {
+            :label => 'API Token',
+            :value => App::Config[:gitlab][:api_token]
```
wangch079, line 42: `API token` and `Base URL` should come from connector configuration (i.e. `.elastic-connectors`), instead of YAML configuration.

lib/connectors/gitlab/connector.rb
```diff
@@ -5,66 +5,101 @@
...
+          chunk_size = projects_chunk.size
+          puts("Fetching files for project #{project[:id]} (#{idx + 1} out of #{chunk_size})...")
+          files = files.map { |file| Connectors::GitLab::Adapter.to_es_document(:file, file) }
+          project[:files] = files
+          @sink.ingest_multiple(files)
         end
+      rescue StandardError => e
+        puts(e.message)
+        puts(e.backtrace)
+        raise e
```
wangch079, line 102: I think these should all be encapsulated in `extractor`.

wangch079 (commented):
Sorry for the late review. Just left some comments.

It's OK to merge to unblock others first.

lib/connectors/base/registry.rb
```diff
@@ -0,0 +1,47 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+module Connectors
+  module Base
+    class Factory
```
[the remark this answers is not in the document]
wangch079: Just noticed @artem-shelkovnikov has a PR https://github.com/elastic/connectors/pull/144 to do so.

PR #134, opened 2022-06-27
Use a single function for now

tarekziade:
Let's add a module to deal with time so we always with UTC


artem-shelkovnikov (approved):

lib/utility/time_util.rb
```diff
@@ -0,0 +1,13 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+module Utility
+  def utc_now
```
wangch079, line 10: Should we define this in `Utility::Time`?

tarekziade:
Not needed anymmore with https://github.com/elastic/connectors/pull/136/files#diff-fe9e8ab719790bfd8feafd355d54651477a69f06dcc53fa7437b6944d73c083dR15

PR #135, opened 2022-06-27
Use ExceptionTracking for now to log errors when polling + add info b…

artem-shelkovnikov:
Minor improvement for development process: currently when error happens during `polling_jobs`, sometimes there's no error message, for example I've put wrong config for Elasticsearch connection and got:

```
E, [2022-06-27T17:56:33.769259 #3666] ERROR -- : Error happened during sync. error: 
```

So, there's no error message or stacktrace, I think for development purposes it's nice to log stacktrace too.

### DIFF
--- a/lib/app/connector.rb
+++ b/lib/app/connector.rb
@@ -51,8 +51,9 @@ module App
         loop do
           polling_jobs
         rescue StandardError => e
-          Utility::Logger.error("Error happened during sync. error: #{e.message}")
+          Utility::ExceptionTracking.log_exception(e, 'Sync failed due to unexpected error.')
         ensure
+          Utility::Logger.info("Sleeping for #{POLL_IDLING} seconds")
           sleep(POLL_IDLING)
         end
       end


lib/app/connector.rb
```diff
@@ -51,8 +51,9 @@ def start_polling_jobs
         loop do
           polling_jobs
         rescue StandardError => e
-          Utility::Logger.error("Error happened during sync. error: #{e.message}")
+          Utility::ExceptionTracking.log_exception(e, 'Sync failed due to unexpected error.')
         ensure
+          Utility::Logger.info("Sleeping for #{POLL_IDLING} seconds")
```
tarekziade, line 56: maybe not now, but we'll need a mechanism that stops the program if we have a failure at every call in the loop.
For instance 3 times the same exception in a row or some kind of threshold

tarekziade (approved):

wangch079 (approved):

lib/app/connector.rb
artem-shelkovnikov: Yep, can do that in future: we have https://github.com/elastic/connectors/blob/main/lib/utility/monitor.rb is a class encapsulating this kind of logic

seanstory: I didn't realize that Monitor had made it into connectors. This makes me happy. :)

PR #139, opened 2022-06-29
Add logic to validate config of connector

artem-shelkovnikov:
Working on connector I faced the following problem:

Once I switch from one connector to another sometimes I don't remember to update config and get no or cryptic error messages about it.

Decided that we can wrap logic about connector metadata stored in Elasticsearch in a single class - later it can be extended to also load and update config when needed.


lib/framework/connector_metadata.rb
```diff
@@ -0,0 +1,45 @@
...
+module Framework
+  class IncompatibleConfigurableFieldsError < StandardError
+    def initialize(expected_fields, actual_fields)
+      super("Connector expected configurable fields: #{expected_fields}, actual stored fields: #{actual_fields}")
+    end
+  end
+
+  class ConnectorMetadata
+    def initialize(connector, elasticsearch_response)
+      # TODO: is it a good way to move ahead?
```
seanstory, line 20: lingering TODO?

lib/app/connector.rb
```diff
@@ -74,9 +75,14 @@ def polling_jobs
         claim_job(connector)
 
         SYNC_JOB_POOL.post do
-          connector_klass.new.sync(connector) do |error|
+          executer = connector_klass.new
+          connector_metadata = Framework::ConnectorMetadata.new(executer, connector)
+          puts "metadata is #{connector_metadata}"
```
seanstory, line 80: lingering `puts`?

lib/framework/connector_metadata.rb
```diff
@@ -0,0 +1,45 @@
...
+      validate!
+    end
+
+    def [](index)
+      @elasticsearch_response[:_source][index]
+    end
+
+    def configured_fields
+      puts @elasticsearch_response[:_source]
+      @elasticsearch_response[:_source][:configured_fields]
```
wangch079, line 33:
```suggestion
      @elasticsearch_response[:_source][:configuration]
```

lib/app/connector.rb
```diff
@@ -74,9 +75,14 @@ def polling_jobs
         claim_job(connector)
 
         SYNC_JOB_POOL.post do
-          connector_klass.new.sync(connector) do |error|
+          executer = connector_klass.new
+          connector_metadata = Framework::ConnectorMetadata.new(executer, connector)
```
wangch079, line 79: I think this validation can be done before `claim_job`

lib/framework/connector_metadata.rb
artem-shelkovnikov: Good one, fixed!

artem-shelkovnikov:
Closing in favour of https://github.com/elastic/connectors/pull/143

PR #141, opened 2022-06-30
Add sync job logs

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/2157

This PR
~1. Create index `.elastic-connector-sync-logs` (see schema [here](https://github.com/elastic/enterprise-search-team/issues/2157#issuecomment-1170869941)) at the connector service startup if it does not exist.~
2. Index a new job when claiming the job.
3. Update the job when completing the job.

#### Pre-Review Checklist

### DIFF
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -76,6 +76,7 @@ module Core
         body = {
           :connector_id => connector_package_id,
           :status => Connectors::SyncStatus::IN_PROGRESS,
+          :worker_hostname => Socket.gethostname,
           :created_at => Time.now
         }
         job = client.index(:index => JOB_INDEX, :body => body)
@@ -165,6 +166,7 @@ module Core
             :connector_id => { :type => :keyword },
             :status => { :type => :keyword },
             :error => { :type => :text },
+            :worker_hostname => { :type => :keyword },
             :indexed_document_count => { :type => :integer },
             :deleted_document_count => { :type => :integer },
             :created_at => { :type => :date },
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -71,46 +71,46 @@ module Core
           }
         }
 
-      client.update(:index => CONNECTORS_INDEX, :id => connector_package_id, :body => body)
+        client.update(:index => CONNECTORS_INDEX, :id => connector_package_id, :body => body)
 
-      body = {
-        :connector_id => connector_package_id,
-        :status => Connectors::SyncStatus::IN_PROGRESS,
-        :created_at => Time.now
-      }
-      job = client.index(:index => JOB_INDEX, :body => body)
-
-      Utility::Logger.info("Successfully claimed job for connector #{connector_package_id}")
-      job['_id']
-    end
-
-    def complete_sync(connector_package_id, job_id, status)
-      sync_status = status[:error] ? Connectors::SyncStatus::FAILED : Connectors::SyncStatus::COMPLETED
-
-      body = {
-        :doc => {
-          :last_sync_status => sync_status,
-          :last_sync_error => status[:error],
-          :last_synced => Time.now
+        body = {
+          :connector_id => connector_package_id,
+          :status => Connectors::SyncStatus::IN_PROGRESS,
+          :created_at => Time.now
         }
-      }
+        job = client.index(:index => JOB_INDEX, :body => body)
+
+        Utility::Logger.info("Successfully claimed job for connector #{connector_package_id}")
+        job['_id']
+      end
+
+      def complete_sync(connector_package_id, job_id, status)
+        sync_status = status[:error] ? Connectors::SyncStatus::FAILED : Connectors::SyncStatus::COMPLETED
+
+        body = {
+          :doc => {
+            :last_sync_status => sync_status,
+            :last_sync_error => status[:error],
+            :last_synced => Time.now
+          }
+        }
 
         client.update(:index => CONNECTORS_INDEX, :id => connector_package_id, :body => body)
 
-      body = {
-        :doc => {
-            :status => sync_status,
-            :completed_at => Time.now
-        }.merge(status)
-      }
-      client.update(:index => JOB_INDEX, :id => job_id, :body => body)
+        body = {
+          :doc => {
+              :status => sync_status,
+              :completed_at => Time.now
+          }.merge(status)
+        }
+        client.update(:index => JOB_INDEX, :id => job_id, :body => body)
 
-      if status[:error]
-        Utility::Logger.info("Failed to sync for connector #{connector_package_id} with error #{status[:error]}")
-      else
-        Utility::Logger.info("Successfully synced for connector #{connector_package_id}")
+        if status[:error]
+          Utility::Logger.info("Failed to sync for connector #{connector_package_id} with error #{status[:error]}")
+        else
+          Utility::Logger.info("Successfully synced for connector #{connector_package_id}")
+        end
       end
-    end
 
       def client
         @client ||= Utility::EsClient.new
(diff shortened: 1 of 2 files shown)


wangch079:
Hi @sphilipse , connector service will create the index `.elastic-connector-sync-logs`. Do remember to grant appropriate permissions to this index when generating API key.

artem-shelkovnikov:
Kibana creates `.elastic-connectors` index, but connectors create a job index - can Kibana create both for consistency/simplicity of expectations? cc @sphilipse

wangch079:
>Kibana creates .elastic-connectors index, but connectors create a job index - can Kibana create both for consistency/simplicity of expectations? cc @sphilipse

@sphilipse , I think this makes sense. I will remove `create_job_index_if_absent` if you agree.

lib/connectors/base/connector.rb
```diff
@@ -12,27 +12,38 @@
...
-        yield error
+        yield @status.dup
       end
 
-      def sync(connector); end
+      def sync(connector)
+        @sink = Utility::Sink::CombinedSink.new(
+          [Utility::Sink::ConsoleSink.new,
+           Utility::Sink::ElasticSink.new(connector['index_name'])]
+        )
```
wangch079, line 45: I moved this to `Connectors::Base::Connector`, if any of its subclasses want to add more, it can do so in its own `sync` method

lib/connectors/gitlab/connector.rb
```diff
@@ -79,6 +75,7 @@ def extract_projects
           next_page_link = @extractor.yield_projects_page(next_page_link) do |projects_chunk|
             projects = projects_chunk.map { |p| Connectors::GitLab::Adapter.to_es_document(:project, p) }
             @sink.ingest_multiple(projects)
+            @status[:index_document_count] += projects.count
```
wangch079, line 78: This is not ideal.

We should examine the response of bulk API and see what operations succeed and what fail. Given `@sink` is a list of `sink`s, different `sink`s may get different results. 

We can improve this later
cc @artem-shelkovnikov @mchernyavskaya

lib/framework/sync_job_runner.rb
```diff
@@ -20,7 +20,7 @@ def initialize(expected_fields, actual_fields)
   class SyncJobRunner
     def initialize(connector_settings)
       @connector_settings = connector_settings
-      @connector_instance = Connectors::REGISTRY.connector(@connector_settings.service_type)
+      @connector_instance = Connectors::REGISTRY.connector(App::Config['service_type'])
```
wangch079, line 23: Kibana won't populate `service_type` in connector configs. We need to rely on `service_type` in yaml file.
cc @sphilipse @artem-shelkovnikov @mchernyavskaya

wangch079 (commented):
@mchernyavskaya Do we have any test accounts on GitLab? I have tested it working on `stub_connector` but not on `GitLab` yet.

lib/framework/sync_job_runner.rb
mchernyavskaya: TBH I think [this](https://github.com/elastic/connectors/pull/149/files#diff-8c4301d73e050dc0f0b752430187d390d0609064f769ad6cdf879bda29ffb29a) place is a tiny bit better as we're already using `App::Config` there and don't add extra imports.

lib/connectors/base/connector.rb
```diff
@@ -12,27 +12,38 @@
 module Connectors
   module Base
     class Connector
+      def initialize
+        @status = {
+          :index_document_count => 0,
```
mchernyavskaya, line 17:
```suggestion
          :indexed_document_count => 0,
```

lib/connectors/base/connector.rb
mchernyavskaya: I'm thinking we should add a separate overridable method for this, but we can address this later if needed.

lib/connectors/gitlab/connector.rb
mchernyavskaya:
```suggestion
            @status[:indexed_document_count] += projects.count
```

lib/connectors/gitlab/connector.rb
```diff
@@ -98,6 +95,7 @@ def extract_project_files(projects_chunk)
           files = files.map { |file| Connectors::GitLab::Adapter.to_es_document(:file, file) }
           project[:files] = files
           @sink.ingest_multiple(files)
+          @status[:index_document_count] += files.count
```
mchernyavskaya, line 98:
```suggestion
          @status[:indexed_document_count] += files.count
```

lib/connectors/stub_connector/connector.rb
```diff
@@ -32,10 +32,9 @@ def health_check(_params)
       end
 
       def sync(connector)
-        body = [
-          { index: { _index: connector['index_name'], _id: 1, data: { name: 'stub connector' } } }
-        ]
-        Utility::EsClient.bulk(:body => body)
+        super
+        @sink.ingest({ :id => 1, :name => 'stub connector' })
+        @status[:index_document_count] = 1
```
mchernyavskaya, line 37:
```suggestion
        @status[:indexed_document_count] = 1
```

mchernyavskaya (changes requested):
LGTM pending the

lib/framework/elastic_connector_actions.rb
```diff
@@ -12,6 +12,7 @@
 module Framework
   class ElasticConnectorActions
     CONNECTORS_INDEX = '.elastic-connectors'
+    JOB_INDEX = '.elastic-connectors-sync-logs'
```
sphilipse, line 15: I thought the index was called `.elastic-connectors-sync-jobs`? We can go with this name as well, but need to be consistent.

lib/framework/sync_job_runner.rb
wangch079: I don't use `require` here and it also worked, and I can add it. 

I tend to avoid passing app-wide accessible resources around.

artem-shelkovnikov: I don't think App::Config is an app-wide accessible resource - it can be used by everything in `App` namespace, but every other module - `core`, `utils` or `connectors` should not have direct access to it, at least for now.

I think we need to have a separate discussion about singletons/app wide abstractions and what is provided where, it's a solid and major topic to discuss

lib/framework/elastic_connector_actions.rb
```diff
@@ -61,22 +62,41 @@ def self.claim_job(connector_package_id)
       }
 
       client.update(:index => CONNECTORS_INDEX, :id => connector_package_id, :body => body)
+
+      body = {
+        :connector_id => connector_package_id,
+        :status => Connectors::SyncStatus::IN_PROGRESS,
+        :created_at => Time.now
+      }
+      job = client.index(:index => JOB_INDEX, :body => body)
```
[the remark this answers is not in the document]
wangch079: >Should we drop `force_sync` flag now?

Do you mean `sync_now` in `.elastic-connectors`? It's reset above.

>we create an IN_PROGRESS job and how does the framework decide that it should be picked?

It's the other way round, connector claims a job, then adds an entry (with `in_progress` status) to `JOB_INDEX`

lib/framework/sync_job_runner.rb
wangch079: > but every other module - `core`, `utils` or `connectors` should not have direct access to it, at least for now.

Half of me is convinced, e.g. A `util` should work without dependency on app-level configs. But our `EsClient` reads configs directly (This maybe wrong). I think we should have our 2nd engineers sync soon.

mchernyavskaya:
@wangch079 would you be able to actualise this PR again? Some of the conversations here are probably no longer relevant.

artem-shelkovnikov (approved):

wangch079:
> @wangch079 would you be able to actualise this PR again? Some of the conversations here are probably no longer relevant.

What do you mean by `actualise`? Does it mean to resolve all the conflicts and merge the PR? I have done so. Let me know if it means anything else, we can address it in a separate PR.

PR #142, opened 2022-07-02
Declare a singleton of EsClient + fix usages of it

artem-shelkovnikov:
After a recent commit EsClient became a regular, non-static class that inherits from `ElasticSearch::Client`.

This PR aims to fix the problem that arose from the refactoring - calls to EsClient.get and such fail because they are not proxied to a singleton.

For now I'm also not sure if we want to have EsClient a singleton in future - I'm against it just for the sake of simplicity, better testability and potentially less problems with multithreading (I'm not 100% sure if Elasticsearch::Client is thread safe)

### DIFF
--- a/lib/utility/es_client.rb
+++ b/lib/utility/es_client.rb
@@ -11,6 +11,10 @@ require 'app/config'
 
 module Utility
   class EsClient < Elasticsearch::Client
+    def self.instance
+      @client ||= new
+    end
+
     def initialize
       super(connection_configs)
     end
--- a/lib/app/connector.rb
+++ b/lib/app/connector.rb
@@ -17,8 +17,6 @@ module App
     QUERY_SIZE = 20
     POLL_IDLING = 60
 
-    @client = Utility::EsClient.new
-
     class << self
 
       def start!
--- a/lib/connectors/stub_connector/connector.rb
+++ b/lib/connectors/stub_connector/connector.rb
@@ -36,7 +36,7 @@ module Connectors
         body = [
           { index: { _index: connector['_source']['index_name'], _id: 1, data: { name: 'stub connector' } } }
         ]
-        Utility::EsClient.bulk(:body => body)
+        Utility::EsClient.instance.bulk(:body => body)
       rescue StandardError => e
         Utility.Logger.error("Error happened when syncing #{display_name}. Error: #{e.message}")
         error = e.message
--- a/lib/utility/sink.rb
+++ b/lib/utility/sink.rb
@@ -54,7 +54,7 @@ module Utility
 
       def initialize(_index_name, flush_threshold = 50, flush_interval = 1.minutes)
         super()
-        @client = Utility::EsClient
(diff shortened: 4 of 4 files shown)


lib/utility/es_client.rb
```diff
@@ -11,6 +11,10 @@
 
 module Utility
   class EsClient < Elasticsearch::Client
+    def self.instance
```
wangch079, line 14: This is not really a singleton. You can still call `Utility::EsClient.new`

wangch079 (commented):
I initially made it a singleton, then @mchernyavskaya changed it in her PR.

I think singleton works for our use case now, and I chose it for simplicity, but I'm also fine to have it as a normal instantiable class.

lib/utility/es_client.rb
artem-shelkovnikov: Good point, I actually reverted the code inside `es_client.rb` to be a real singleton - before as you noticed it was allowing you to work with class instances OR singleton, which is a really really confusing way to work with a class.

vidok: If we want to make `EsClient` a singleton I would do it with this https://ruby-doc.org/stdlib-2.6/libdoc/singleton/rdoc/Singleton.html. It will simplify the code a bit and also make sure that `initialize` method is private.

wangch079: TIL there's a Singleton module 👍

vidok:
> I initially made it a singleton, then @mchernyavskaya changed it in her PR.
> 
> I think singleton works for our use case now, and I chose it for simplicity, but I'm also fine to have it as a normal instantiable class.

I this the question we need to answer is: do we want to have a sharable Elasticsearch connection (or a connection pool) in the connectors service runtime or we want to create multiple connections with different configuration or even connect to different ES instances. 

In my opinion, the app should have access to ES clients through a single abstraction like: 
```
  Utility::ESclient.client
```

`client` can be either `elasticsearch-ruby` client instance or a pool of ES clients. 

Also, I don't think we need to override any native ES client methods or proxying them via `method_missing` or delegate. I might be missing something but I don't see any issues that would require to change the default client's behaviour. At the same time, the less we have something on top of the native client, the cheaper is the code maintenance.

vidok:
> @vidok agreed.
> 
> For now there are no concrete plans on pooling/threading/using EsClient for access to multiple Elasticsearch clusters. We might or might not need that in future, therefore I'm up for moving with the simplest way possible.
> 
> In the end I've changes EsClient module to expose a `EsClient.client` method that returns a cached instance of Elasticsearch client. I think this is just simple enough for now and we can change it to something better later.

Yeah, I totally agree with you on leaving `pooling/threading/using` for the future and move forward with a simple solution.

vidok (approved):

wangch079:
It will make sense to have multiple connections when the connector service can ingest data for multiple connectors. But it's good enough for our use case now.

PR #145, opened 2022-07-03
Add basic MongoDB connector

artem-shelkovnikov:
This PR adds a basic MongoDB connector.

Connector accepts the following config:

- host - hostname of MongoDB instance
- database - database on the instance that needs to be synced
- collection - collection in the database that needs to be synced

No auth is yet supported.

Also the code is very basic and does not transform the data received (except renaming `"_id"` to :`id`).


Gemfile
```diff
@@ -48,3 +48,6 @@ gem 'elasticsearch'
 
 # Dependencies for oauth
 gem 'signet', '~> 0.16.0'
+
+# Dependency for mongo connector
+gem 'mongo'
```
artem-shelkovnikov, line 53: TODO: pin

lib/connectors/mongodb/connector.rb
```diff
@@ -0,0 +1,76 @@
...
+#
+
+# frozen_string_literal: true
+
+require 'mongo'
+require 'connectors/base/connector'
+
+module Connectors
+  module MongoDB
+    class Connector < Connectors::Base::Connector
```
seanstory, line 14: I feel pretty weird about a whole connector being added without any tests.

lib/connectors/mongodb/connector.rb
```diff
@@ -0,0 +1,76 @@
...
+
+        config = connector.configuration
+
+        host = config[:host][:value]
+        database = config[:database][:value]
+        collection = config[:collection][:value]
+
+        mongodb_client = create_client(host, database)
+
+        mongodb_client[collection].find.each do |document|
```
seanstory, line 50: This looks like it's bypassing the need for an Extractor. Is this a patter that we're going to be running with going forward?

lib/connectors/mongodb/connector.rb
```diff
@@ -0,0 +1,76 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+require 'mongo'
```
seanstory, line 9: I totally missed this - I know that Tarek wrote a mongo connector for his spacetime, but when did adding a mongo connector become a priority for the product? I've never seen it in the list of potential connectors from PM (like with zoom, s3, assana, etc)

lib/connectors/mongodb/connector.rb
artem-shelkovnikov: I'm gonna add more tests a bit later, as a unified effort - now things get moved around too often to have a stable "framework" definition and distribution of responsibilities.

lib/connectors/mongodb/connector.rb
tarekziade: Having a second connector built as we mature the BYOC protocol helps us a lot. It's a tech preview -- marked as such, it's not hurting anyone. In any case, I know that this backend is on the PM's radar for sure.

lib/connectors/mongodb/connector.rb
seanstory: I disagree. We have a suite of connectors in Workplace Search that we're planning to move over here, right? We don't intend to re-write them, correct? So we will have the `Config`, `Adapter`, `Extractor` pattern to maintain. This is just making two models of code execution for us to maintain, and muddies the water on where logic should live.

> YAGNI

I wish we'd stop throwing this out all over the place. It's not a trump card. The Workplace Search Connectors show that we _are_ going to need this kind of abstraction as we grow. By throwing it away now, we're abandoning lessons-learned and assuming we can do better if we start from scratch. That's an anti-pattern.

serenachou: Just a bit of perspective on the Workplace Search connectors to add to the conversation

>  We have a suite of connectors in Workplace Search that we're planning to move over here, right? We don't intend to re-write them

Yes, but here are some subtle nuances to the yes. 

- The timeline to move them isn't set in stone, will our existing design patterns in Workplace Search be the ones we want to use if we port some over a year from now? 2? What could change in a longer timeline? We are likely going to be pushing for new connectors that branch out our use cases to be served.
- We likely won't port over all of them at the same time  because not all of them are equally worth the work to port over *cough* *cough* Zendesk.  I think our 8.2 release showed me that a lift and shift would have to be done in phases and we would need to challenge their structure to fit new challenges when we connect the dots to Elasticsearch. 
-  A rewrite is likely needed in some portion to be honest. Definitely we would need to rewrite for example Slack and Gmail as we won't be bringing forward concepts like private sources. 

I think the challenge we'll face is to balance being able to maximize how much of the useful connectors stuff we can bring over with the timeline to truly develop our new ingestion catalog identity. I don't expect that @danajuratoni 's project brief on the 8.5 selected connectors will or should request all of the Workplace Search ones, especially since we don't have Oauth support quite sorted yet, nor do we have indexing rules. If we're going to support native connectors, I consider Indexing rules very necessary as the right guardrails to give to customers who like to try and ingest ALL THE THINGs.

lib/connectors/mongodb/connector.rb
serenachou: MongoDB wasn't mentioned in the original project brief, but it's tied to some thinking from Matt as to the types of use cases we'd likely want to support going forward. 

We're actually seeing an interesting customer interest in MongoDB Atlas (https://www.mongodb.com/pricing) as well as some interesting use cases to see how we stack against Couchbase/Dynamo. MongoDB would not be ready in 8.4 to be considered ready as a native connector, but with our technical preview connector framework & service, it's a perfect proving ground to get earlier feedback. I foresee us doing a lot of technical preview connectors like this - early, out of release version releases into the repo, and then as usage and other metrics dictate, a promotion process to native. As part of this we'll want to harden up our definition of done to have a validation checklist to get connectors elevated to the native status, but in the mean time, it's great to sweeten the pot with this connector so PM can shop it around and use it to explore the new market use cases. 

FWIW, Tarek proposed it to the leads, and did get buyin from the PM team at the beginning of the sprint, and since connector service was progressing pretty well (thanks Chenhui!) we felt like we had some capacity to handle adding this as a stretch.

lib/connectors/mongodb/connector.rb
danajuratoni: Adding a short comment here too, since the retro session will happen in 2 weeks.
> We have a suite of connectors in Workplace Search that we're planning to move over here, right? We don't intend to re-write them, correct?

Along Serena's valid points - how much can be reused from the WS connectors and if the abstractions make sense is something that needs to be reviewed by the eng team, but I don't expect we'll be able to migrate them as they are. Not only because of the async model, but also since feature parity is not a goal (rather ease of use, configurability to name a few). e.g. we won't have 3 sync types as in WS.
I think everyone would benefit from close technical collaboration between working groups, so I'm looking forward to hearing suggestions in the retro!

PR #146, opened 2022-07-03
Separate sinks from batching and fix problem with sink eventually dropping some documents

artem-shelkovnikov:
Problem that I faced:
ElasticSink relies on a timer to eventually flush the data into Elasticsearch when threshold of flush is not met. What it leads to is that when sync ends, sink does not always (almost never) flush the last batch of data into Elasticsearch, leading to data loss.

What this PR does:

- Introduces a new Batcher class that moves away the batching logic from the sinks. It's definitely not a final solution, but good enough to move forward and provide a way to either ingest data without batching, if needed, or with batching. We'll need to re-visit it later
- Cleans up Sink classes interfaces, moving methods around and removing not-needed statements.
- Adds very basic error handling to ElasticSink - before it was just sending request without checking the response, now it's possible to see errors that appear during ingestion

What this PR does not do:

- Move files around
- Change namespaces

These ^^^ things can be done in a separate PR.

### DIFF
--- a/lib/connectors/gitlab/connector.rb
+++ b/lib/connectors/gitlab/connector.rb
@@ -25,10 +25,6 @@ module Connectors
           :base_url => configurable_fields[:base_url][:value],
           :api_token => configurable_fields[:api_token][:value]
         )
-        @sink = Utility::Sink::CombinedSink.new(
-          [Utility::Sink::ConsoleSink.new,
-           Utility::Sink::ElasticSink.new(App::Config[:connector_package_id])]
-        )
       end
 
       def display_name
@@ -48,8 +44,15 @@ module Connectors
         }
       end
 
-      def sync_content(_params = {})
-        extract_projects
+      def sync_content(connector)
+        sink = Utility::Sink::CombinedSink.new(
+          [
+            Utility::Sink::ConsoleSink.new,
+            Utility::Sink::ElasticSink.new(connector['_source']['index_name'])
+          ]
+        )
+
+        extract_projects(sink)
       end
 
       def deleted(_params = {})
@@ -70,13 +73,13 @@ module Connectors
         Connectors::GitLab::CustomClient::ClientError
       end
 
-      def extract_projects
+      def extract_projects(sink)
         next_page_link = nil
         loop do
           next_page_link = @extractor.yield_projects_page(next_page_link) do |projects_chunk|
             projects = projects_chunk.map { |p| Connectors::GitLab::Adapter.to_es_document(:project, p) }
-            @sink.ingest_multiple(projects)
-            extract_project_files(projects_chunk)
+            sink.ingest_multiple(projects)
+            extract_project_files(sink, projects_chunk)
           end
           break unless next_page_link.present?
         end
@@ -86,7 +89,7 @@ module Connectors
         raise e
       end
 
-      def extract_project_files(projects_chunk)
+      def extract_project_files(sink, projects_chunk)
         projects_chunk.each_with_index do |project, idx|
           project = project.with_indifferent_access
           files = @extractor.fetch_project_repository_files(project[:id])
@@ -94,7 +97,7 @@ module Connectors
           puts("Fetching files for project #{project[:id]} (#{idx + 1} out of #{chunk_size})...")
           files = files.map { |file| Connectors::GitLab::Adapter.to_es_document(:file, file) }
           project[:files] = files
-          @sink.ingest_multiple(files)
+          sink.ingest_multiple(files)
         end
       rescue StandardError => e
         puts(e.message)
--- a/lib/utility/sink.rb
+++ b/lib/utility/sink.rb
@@ -30,11 +30,7 @@ module Utility
     end
 
     def add_multiple(elements)
-      elements.each do |element|
-        @queue << element
-      end
-
-      flush if ready_to_flush?
+      elements.each { |element| add(element) }
     end
 
     def flush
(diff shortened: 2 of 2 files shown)


lib/utility/sink.rb
```diff
@@ -12,28 +12,57 @@
...
+    end
+
+    def add_multiple(elements)
+      elements.each do |element|
+        @queue << element
+      end
 
-    def print_delim
-      puts '----------------------------------------------------'
+      flush if ready_to_flush?
```
wangch079, line 37: should move this into `each` block?

lib/utility/sink.rb
```diff
@@ -12,28 +12,57 @@
...
-      def ingest(_document)
-        print_header 'Got a single document:'
+    def ready_to_flush?
+      @queue.size >= @flush_threshold
+    end
+  end
+
+  module Sink
+    class Base
+      def with_batching(batch_size = 50)
```
wangch079, line 56: Where is this used?

lib/connectors/gitlab/connector.rb
```diff
@@ -70,13 +73,13 @@ def custom_client_error
         Connectors::GitLab::CustomClient::ClientError
       end
 
-      def extract_projects
+      def extract_projects(sink)
```
wangch079, line 76: Really not a fan of passing `sink` in each method. Can we make it an instance variable?

lib/connectors/gitlab/connector.rb
```diff
@@ -48,8 +44,15 @@ def configurable_fields
         }
       end
 
-      def sync_content(_params = {})
-        extract_projects
+      def sync_content(connector)
+        sink = Utility::Sink::CombinedSink.new(
+          [
+            Utility::Sink::ConsoleSink.new,
```
wangch079, line 50: I'm always curious what is the purpose of `ConsoleSink`? Just log something in the console?

lib/connectors/gitlab/connector.rb
artem-shelkovnikov: Reverted in https://github.com/elastic/connectors/pull/146/commits/7140895376108bac2542cf07097b3c7607683845

artem-shelkovnikov: Ha, actually it's not possible any more now - because ElasticSink requires knowing where to write data, and for now the way connectors are instantiated they don't have access to connector settings - and it makes sense, because we instatiate a connector also for initial setup, where there are no settings or records in DB, thus it's not possible to actually create a real sink in the initializer, so I decided to move the sink creation to the extraction method, and thus it's not an instance variable any more.

Any thoughts on how to do it better?

lib/utility/sink.rb
artem-shelkovnikov: My bad - it's not used in this PR actually, but I use it in my MongoDB PR that's in draft mode.

To use this abstraction in this PR we need to update the code of Gitlab connector, which should probably be done in a separate PR

lib/utility/sink.rb
artem-shelkovnikov: Yeah probably a good idea, need to think more about it.
Just thinking out loud:

The difference will be in behaviour when `ingest_multiple` is used, for example connector attempts to ingest 60 documents, while batching is set up to send data to Elasticsearch in batches of 50.

In currently done way it will do the following: add 60 elements to the queue, then send 50 + 10 elements to Elasticsearch immediately after.

If we do the change proposed by @wangch079, it will send 50 elements to Elasticsearch, then add 10 more and wait.

Intuitively, second case looks better. Any additional thoughts?

lib/utility/sink.rb
artem-shelkovnikov: Which also brings in the question - do we want to give user the flexibility of choosing whether they want to batch the requests or not. I think it's nice to have, but it also increases the complexity of the system - more stuff to document and explain

lib/connectors/gitlab/connector.rb
wangch079: Agree, but it's not necessary that all instance variables are assigned during instantiation, we can do so when calling `sync_content`

lib/utility/sink.rb
mchernyavskaya: I'd say batching isn't a feature, it's a performance-related implementation detail. Since we always prefer to improve performance, we can safely make it a default implementation. For now, I can hardly imagine a case when an override would be needed, but if it happens we'll provide for it.

lib/utility/sink.rb
mchernyavskaya: I have a revolutionary idea (kidding).

```
    def add(element)
      @queue << element

      flush if ready_to_flush?
    end

    def add_multiple(elements)
      elements.each do |element|
        add(element)
      end
    end
```

Why do you add the single-item method if not to use it? =)

mchernyavskaya (changes requested):

lib/utility/sink.rb
artem-shelkovnikov: Yeah I agree, though at current state of code it's probably not possible yet to have a robust transparent batching interface. I'd vote for merging this PR and taking additional time to make this batching interface transparent (probably not much time to do, just a lot of code moving around)

artem-shelkovnikov:
Postponing work here as more changes are needed to make it work nicer - will return back to it soon

artem-shelkovnikov:
Closing cause it's a bit too stale and the changes I want to make will be from scratch

PR #147, opened 2022-07-04
Custom connector implementation - some fixes

mchernyavskaya:
## Follow-up on https://github.com/elastic/enterprise-search-team/issues/2128

Small fixes and additions to custom connector (gitlab).

- [x] the job runner knows when it's the last page
- [x] use `Utility::Logger` instead of puts where needed

#### Pre-Review Checklist

## Related Pull Requests

<!--List any relevant PRs here or remove the section if this is a standalone PR.

* https://github.com/elastic/connectors/pull/132


lib/app/connector.rb
```diff
@@ -10,10 +10,11 @@
 require 'connectors'
 require 'framework'
 require 'utility'
+require 'app/config'
 
 module App
   module Connector
-    POLL_IDLING = 60
+    POLL_IDLING = (App::Config['idle_timeout'] || 60).to_i
```
artem-shelkovnikov, line 17: This setting should be reflected in the config file schema.

lib/app/connector.rb
```diff
@@ -23,13 +24,12 @@ def start!
...
 
-      def create_connector(index_name)
+      def create_connector(index_name, force: false)
         connector_settings = Framework::ConnectorSettings.fetch(App::Config['connector_package_id'])
 
-        if connector_settings.nil?
-          Framework::ElasticConnectorActions.create_connector(index_name)
-
-          connector_settings = Framework::ConnectorSettings.fetch(App::Config['connector_package_id'])
+        if connector_settings.nil? || force
```
artem-shelkovnikov, line 30: I'd break the condition in such a way, that if the create_connector is called, but connector is there it should log that the connector has already been created, so skipping it. If `force` flag is passed, it should also note that connector has already been created, but now we're force overwriting it (maybe also providing what's the diff between existing connector setting and what we're writing for it if possible)

lib/app/console_app.rb
```diff
@@ -46,17 +54,20 @@ def register_connector
       if id.present?
         puts "You already have registered a connector with ID: #{id}. Registering a new connector will not use the existing one."
         puts 'Are you sure you want to continue? (y/n)'
-        return unless gets.chomp.casecmp('y').zero?
+        return false unless gets.chomp.casecmp('y')&.zero?
```
artem-shelkovnikov, line 57: Why `return false` here?

If it's needed, probably other return statements in the method should also return a value

lib/framework/connector_settings.rb
```diff
@@ -12,16 +12,18 @@ module Framework
   class ConnectorSettings
     def self.fetch(connector_package_id)
       es_response = ElasticConnectorActions.load_connector_settings(connector_package_id)
-
+      if es_response.nil? || es_response['found'] == false
+        nil
+      end
```
artem-shelkovnikov, line 17: I think it's great to log here the fact that nothing was found, plus what's the case when `es_response.nil?` is true? Is it even possible?

lib/framework/elastic_connector_actions.rb
```diff
@@ -55,7 +56,9 @@ def self.claim_job(connector_package_id)
         :doc => {
           :sync_now => false,
           :sync_status => Connectors::SyncStatus::IN_PROGRESS,
-          :last_synced => Time.now
+          :scheduling => {
+            :last_synced => Time.now # TODO not sure if this is the right place to set this
+          }
```
artem-shelkovnikov, line 61: How does it affect this PR?  Right now it looks out of scope of this PR

lib/framework/elastic_connector_actions.rb
```diff
@@ -82,7 +87,17 @@ def self.complete_sync(connector_package_id, error)
...
+      @client ||= Utility::EsClient.new
+    end
+
+    # should only be used in CLI
+    def self.ensure_index_exists(index_name)
+      client.indices.create(:index => index_name) unless client.indices.exists?(:index => index_name)
+    end
+
+    # should only be used in CLI
+    def self.ensure_config_index_exists
```
artem-shelkovnikov, line 99:
```suggestion
    def self.ensure_connectors_index_exists
```

lib/framework/sync_job_runner.rb
```diff
@@ -55,14 +55,19 @@ def cron_parser(cronline)
...
     def should_sync?
-      return false unless @connector_settings.scheduling_settings[:enabled]
-      return true if @connector_settings[:sync_now]
+      # sync_now should have priority over cron
+      return true if @connector_settings[:sync_now] == true
+      scheduling_settings = @connector_settings.scheduling_settings
+      return false unless scheduling_settings[:enabled] == true
 
-      last_synced = @connector_settings.scheduling_settings[:last_synced]
+      last_synced = scheduling_settings[:last_synced]
```
artem-shelkovnikov, line 63: I see, probably it was my mistake when porting the code and it should have been `self[:last_synced]`

lib/framework/sync_job_runner.rb
```diff
@@ -55,14 +55,19 @@ def cron_parser(cronline)
...
       return true if last_synced.nil? || last_synced.empty? # first run
 
       last_synced = Time.parse(last_synced) # TODO: unhandled exception
       sync_interval = scheduling_settings['interval']
+      if sync_interval.nil? || sync_interval.empty? # no interval configured
```
artem-shelkovnikov, line 68: I strongly feel this place deserves a log info line

lib/utility/es_client.rb
```diff
@@ -10,26 +10,47 @@
 require 'app/config'
 
 module Utility
```
artem-shelkovnikov, line 12: This change makes EsClient an instance class, instead of singleton, should the class be reverted to its original state?

artem-shelkovnikov (changes requested):

lib/app/connector.rb
mchernyavskaya: It's working this way already. Here's the flow:
```
Please select the command:
    start synchronization (sync)
--> register connector with Elasticsearch (register)
    check the status of a third-party service (status)
    end the program (exit)
You already have registered a connector with ID: vbJiyYEB-Xa53KAxjgk5. Registering a new connector will not use the existing one.
Are you sure you want to continue? (y/n)
y
Please enter index name for data ingestion. Use only letters, underscored and dashes.
two
I, [2022-07-05T08:16:00.577987 #6964]  INFO -- : Successfully registered connector two with ID v7JuzYEB-Xa53KAxNAlC
Please store connector ID in config file and restart the program.
```

lib/app/connector.rb
mchernyavskaya: Makes sense, fixing this.

lib/app/console_app.rb
mchernyavskaya: Good catch. Yeah they should. I missed one case.

lib/framework/connector_settings.rb
mchernyavskaya: I'm not so well acquainted with the ruby lib, so IDK if there's a nil possibility there. I think I'm just checking for it "in case". We can probably cut this.
I don't really want to log the not found case, since it might be quite common that if you register a new connector, the old settings aren't found. I don't want it to look like an error condition.

lib/framework/elastic_connector_actions.rb
mchernyavskaya: If it's not updated, the next cycle picks it up to sync again, thinking "oh there's no sync date, so it's a first sync".

lib/framework/sync_job_runner.rb
mchernyavskaya: It clutters the log very much when the idle interval is short and no interval is configured. We might have it as a debug level perhaps?

lib/utility/es_client.rb
mchernyavskaya: Do we really need a singleton here?

PR #148, opened 2022-07-05
Fixing the make run

mchernyavskaya:
Fixing `make run` to take the service type from app config instead of config in kibana.

#### Pre-Review Checklist
- [ ] Covered the changes with automated tests
- [x] Tested the changes locally

PR #149, opened 2022-07-05
Fix make run

mchernyavskaya:
Fixing `make run`

Ad-hoc/trivial change and does not have a corresponding
issue.

Reading from app config instead of kibana.

replaces https://github.com/elastic/connectors/pull/148/commits

#### Pre-Review Checklist

### DIFF
--- a/lib/app/connector.rb
+++ b/lib/app/connector.rb
@@ -47,7 +47,7 @@ module App
       def create_sync_job_runner
         connector_settings = Framework::ConnectorSettings.fetch(App::Config['connector_package_id'])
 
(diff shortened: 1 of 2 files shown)


wangch079 (commented):
I think I got this fixed in my PR https://github.com/elastic/connectors/pull/141, see my comment [here](https://github.com/elastic/connectors/pull/141/files#r914434843)

You can access `App::Config` directly instead of passing it around.

artem-shelkovnikov (approved):

mchernyavskaya:
No longer relevant.

PR #150, opened 2022-07-06
Create system indices if absent

wangch079:
System indices (currently `.elastic-connectors` and `.elastic-connectors-sync-jobs`) are created at Kibana startup. But connector service can't assume the existence of Kibana, and needs to create those indices if they are absent.

#### Pre-Review Checklist

### DIFF
--- a/lib/app/console_app.rb
+++ b/lib/app/console_app.rb
@@ -42,6 +42,7 @@ module App
       connector_id = App::Config[:connector_package_id]
       config_settings = Core::ConnectorSettings.fetch(connector_id)
       Core::ElasticConnectorActions.ensure_index_exists(config_settings[:index_name])
+      Core::ElasticConnectorActions.ensure_job_index_exists
       App::Connector.start!
     end
 
@@ -52,6 +53,7 @@ module App
       connector_id = App::Config[:connector_package_id]
       config_settings = Core::ConnectorSettings.fetch(connector_id)
       Core::ElasticConnectorActions.ensure_index_exists(config_settings[:index_name])
+      Core::ElasticConnectorActions.ensure_job_index_exists
       Core::ElasticConnectorActions.force_sync(connector_id)
       App::Worker.start!
     end
--- a/lib/app/worker.rb
+++ b/lib/app/worker.rb
@@ -28,6 +28,8 @@ module App
 
       def pre_flight_check
         raise "#{App::Config['service_type']} is not a supported connector" unless Connectors::REGISTRY.registered?(App::Config['service_type'])
+        Framework::ElasticConnectorActions.ensure_connectors_index_exists
+        Framework::ElasticConnectorActions.ensure_job_index_exists
       end
 
       def start_polling_jobs
--- a/lib/app/worker.rb
+++ b/lib/app/worker.rb
@@ -28,8 +28,8 @@ module App
 
       def pre_flight_check
         raise "#{App::Config['service_type']} is not a supported connector" unless Connectors::REGISTRY.registered?(App::Config['service_type'])
-        Framework::ElasticConnectorActions.ensure_connectors_index_exists
-        Framework::ElasticConnectorActions.ensure_job_index_exists
+        Core::ElasticConnectorActions.ensure_connectors_index_exists
+        Core::ElasticConnectorActions.ensure_job_index_exists
       end
 
       def start_polling_jobs
--- a/lib/connectors/base/connector.rb
+++ b/lib/connectors/base/connector.rb
@@ -32,7 +32,12 @@ module Connectors
         yield error
       end
 
-      def sync(connector); end
+      def sync(connector)
+        @sink = Utility::Sink::CombinedSink.new(
+          [Utility::Sink::ConsoleSink.new,
+           Utility::Sink::ElasticSink.new(connector['index_name'])]
+        )
+      end
 
       def source_status(params = {})
         health_check(params)
--- a/lib/connectors/gitlab/connector.rb
+++ b/lib/connectors/gitlab/connector.rb
@@ -46,10 +46,7 @@ module Connectors
       end
 
       def sync(connector_settings)
-        @sink = Utility::Sink::CombinedSink.new(
-          [Utility::Sink::ConsoleSink.new,
-           Utility::Sink::ElasticSink.new(connector_settings[:index_name])]
-        )
+        super
         extract_projects
       end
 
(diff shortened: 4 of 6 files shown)


lib/app/connector.rb
```diff
@@ -28,6 +28,8 @@ def start!
 
       def pre_flight_check
         raise "#{App::Config['service_type']} is not a supported connector" unless Connectors::REGISTRY.registered?(App::Config['service_type'])
+        Framework::ElasticConnectorActions.ensure_connectors_index_exists
```
wangch079, line 31: This won't stop the app from exiting as there won't be any entry in the connector index.

lib/framework/elastic_connector_actions.rb
```diff
@@ -12,88 +12,137 @@
...
+    JOB_INDEX = '.elastic-connectors-sync-jobs'
 
-    def self.force_sync(connector_package_id)
-      body = {
-        :doc => {
-          :scheduling => { :enabled => true },
-          :sync_now => true
+    class << self
+
+      def force_sync(connector_package_id)
```
artem-shelkovnikov, line 19: While I agree with refactoring (`class << self`), I think we should avoid having multiple various changes in a single PR - it should be split to a different PR and reviewed separately.

I see it as a recurring pattern - when PR is small and starts to expand, adding multiple various changes which slows down the review/merge procedure + increase the changes/size of merge conflicts.

lib/framework/elastic_connector_actions.rb
```diff
@@ -12,88 +12,137 @@
...
-    end
+      # should only be used in CLI
+      def ensure_index_exists(index_name, body = {})
+        client.indices.create(:index => index_name, :body => body) unless client.indices.exists?(:index => index_name)
+      end
 
-    # should only be used in CLI
-    def self.ensure_connectors_index_exists
-      ensure_index_exists(CONNECTORS_INDEX)
+      def ensure_alias_exists(alias_name, body = {})
```
artem-shelkovnikov, line 97: What's up with the alias? Why do we have/need it?

artem-shelkovnikov:
@wangch079 can you provide a PR description too, as to what problem is being solved?

lib/framework/elastic_connector_actions.rb
```diff
@@ -12,88 +12,137 @@
...
+        @client ||= Utility::EsClient.new
+      end
 
-    # should only be used in CLI
-    def self.ensure_index_exists(index_name)
-      client.indices.create(:index => index_name) unless client.indices.exists?(:index => index_name)
-    end
+      # should only be used in CLI
+      def ensure_index_exists(index_name, body = {})
+        client.indices.create(:index => index_name, :body => body) unless client.indices.exists?(:index => index_name)
```
seanstory, line 94: Should this check to make sure the response is a 20x? Otherwise `ensure_index_exists` isn't really ensuring anything, it's just attempting to create it.

seanstory (commented):
As a side note, we should be careful about our verbiage. These indices are not "system indices" - that's a special term for Elasticsearch-managed indices with special permissions. Right now, I think they're just normal indices. But they could be hidden indices.

lib/framework/elastic_connector_actions.rb
wangch079: Agree. Will avoid doing so.

lib/framework/elastic_connector_actions.rb
wangch079: For migration purposes. See [here](https://docs.google.com/document/d/1_4VCIDhj-a3EVHIJ_tADm28AtqZdZvo4CW-vrYAWcIQ/edit#bookmark=id.iqzlyufomwsk)

lib/framework/elastic_connector_actions.rb
wangch079: If it's not 20x, I think it will just raise an error, and the app/CLI will exit. We should probably catch it and display more user-friendly message. cc @artem-shelkovnikov @mchernyavskaya

lib/framework/elastic_connector_actions.rb
```diff
@@ -12,88 +12,137 @@
...
-      ensure_index_exists(CONNECTORS_INDEX)
+      def ensure_alias_exists(alias_name, body = {})
+        body[:aliases] = { alias_name => { :is_write_index => true } }
+        client.indices.create(:index => "#{alias_name}_v1", :body => body) unless client.indices.exists?(:index => alias_name)
+      end
+
+      # should only be used in CLI
+      def ensure_connectors_index_exists
+        body = {
+          :mappings => {
```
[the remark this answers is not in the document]
wangch079: Good point

lib/framework/elastic_connector_actions.rb
artem-shelkovnikov: TIL, thanks for the explanation! In this case we might need to add logic to find out which version index we need to create based on what's the last known version, but probably could be done later

wangch079: If the alias/index does not exist, we can assume we need to create `_v1` index. 

I would like to keep this PR simple, as there can be a lot of edge cases, e.g. alias not existing but index existing. I added it as one of the topic to discuss in [Engineers sync](https://docs.google.com/document/d/1Z3t79RpKpML-4vsuFAavXzMqAcw_cEbDqoK7Ov33hVg/edit#)

lib/framework/elastic_connector_actions.rb
wangch079: @artem-shelkovnikov, the connector service doesn't handle edge cases well. I think it's fine as we want to move fast. We can have a dedicated PR later to make the connector service more robust.

artem-shelkovnikov: Good call, 100% agree

artem-shelkovnikov (approved):

lib/framework/elastic_connector_actions.rb
seanstory: For when you get to that client augmentation, I suggest looking at [this code](https://github.com/elastic/ent-search/blob/efeb730e1dda78631b7b8a47ed8ecc35c917f239/lib/swiftype/es/client.rb#L14-L30). I'm currently leveraging this Elasticsearch Client/wrapper for the Elasticsearch Crawler, and really liking how it works. I think it might be a good candidate for us to open so that we don't have to reinvent it from scratch.

PR #152, opened 2022-07-06
Remove old gemspec files and their usages

artem-shelkovnikov:
Just cleaning up old gemspec files and updating `make build_gem` command to build the `connectors_utility` gem.


vidok (approved):

artem-shelkovnikov:
> I think the initial plan is to release the connector app as a gem so ent-search can just call App::Connector.start!.

Indeed we want to do it, but in later releases, for now I think it makes sense to remove everything that's not working/not being used, and then this PR can be reverted to return back the gems when needed.

PR #153, opened 2022-07-06
Rename App::Connector to App::Worker

artem-shelkovnikov:
Worker seems to be an accurate name for the contents of the class.


mchernyavskaya (approved):
LGTM!

PR #155, opened 2022-07-06
Clean up unused Base::Connector methods

artem-shelkovnikov:
Just a housekeeping commit - Base::Connector has a lot of stuff inside that is not used any more, cleaning up.


mchernyavskaya (approved):
High time to do that! The extra code is now confusing.

PR #158, opened 2022-07-07
Fix incorrect access to instance variable

vidok:
This will fix a small typo and some specs to avoid this issue in the future. 
#### Pre-Review Checklist

### DIFF
--- a/connectors_utility.gemspec
+++ b/connectors_utility.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name        = 'connectors_utility'
-  s.version     = '0.5'
+  s.version     = '0.6'
   s.homepage    = 'https://github.com/elastic/connectors'
   s.summary     = 'Gem containing shared Connector Services libraries'
   s.description = ''
--- a/connectors_utility.gemspec
+++ b/connectors_utility.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name        = 'connectors_utility'
-  s.version     = '0.6'
+  s.version     = '0.5'
   s.homepage    = 'https://github.com/elastic/connectors'
   s.summary     = 'Gem containing shared Connector Services libraries'
   s.description = ''
--- a/lib/utility/elasticsearch/index/text_analysis_settings.rb
+++ b/lib/utility/elasticsearch/index/text_analysis_settings.rb
@@ -66,7 +66,7 @@ module Utility
         def initialize(language_code: DEFAULT_LANGUAGE, analysis_icu: false)
           @language_code = language_code.to_sym
 
-          raise UnsupportedLanguageCode unless language_data[language_code]
+          raise UnsupportedLanguageCode unless language_data[@language_code]
 
           @analysis_icu = analysis_icu
           @analysis_settings = icu_settings(analysis_icu)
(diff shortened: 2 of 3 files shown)


connectors_utility.gemspec
```diff
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name        = 'connectors_utility'
-  s.version     = '0.5'
+  s.version     = '0.6'
```
tarekziade, line 3: what is your strategy for this version ?
could you look at https://github.com/elastic/connectors/blob/main/docs/RELEASING.md and expand it to match the new behavior?

vidok: > what is your strategy for this version ?

There is no any strategy because it's a temporary gem and it won't be published to rubygems. 

> could you look at https://github.com/elastic/connectors/blob/main/docs/RELEASING.md and expand it to match the new behavior?

I'd rather align `connectors_utility` version strategy to https://github.com/elastic/connectors/blob/main/docs/RELEASING.md if it's okay

artem-shelkovnikov (approved):
LGTM, shall we address the versioning in a different PR?

vidok:
@tarekziade @seanstory @artem-shelkovnikov I will address there versioning in a new PR

connectors_utility.gemspec
vidok: > Any reason to not publish it? By publishing it, we make it easier for ent-search to consume without having to have build connectors locally (so for someone who does rm -rf ent-search/vendor && ent-search/script/bundle install for some reason)

You have absolutely valid concerns. I just wanted to avoid release process to save us time. We will need to get rid of this gem once we move crawler to connectors. That was my understanding after discussions with your and @tarekziade. I also think it does not have any value for our community. I would be happy to publish to our internal package service (GitHub packages? https://github.com/orgs/elastic/packages)

vidok:
> @tarekziade @seanstory @artem-shelkovnikov I will address the versioning in a new PR

https://github.com/elastic/connectors/pull/168

connectors_utility.gemspec
seanstory: > We will need to get rid of this gem once we move crawler to connectors

I had assumed that we'd keep it and grow it, for consumption by anyone who wanted to? Like I don't think it's a bad idea to have a tool out there that spits out the optimal mappings for App Search.

> I would be happy to publish to our internal package service (GitHub packages? https://github.com/orgs/elastic/packages)

Not a bad idea. Can bundler pick up from packages? I know it can grab from the repo, but I thought it was effectively doing a clone and building from the gemspec for that? @tarekziade might also know how to get us access to the elastic gem server from our internal artifactory?

vidok: Yep, `bundler` can easily download and install gems from GitHub packages: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-rubygems-registry

seanstory: TIL. I'm surprised this isn't linked from the bundler docs!
Looks like that would require some extra setup for ent-search through, to configure an authenticated github user. Not opposed to it, just want to call out the effort.

vidok: > TIL. I'm surprised this isn't linked from the bundler docs!
> Looks like that would require some extra setup for ent-search through, to configure an authenticated github user. Not opposed to it, just want to call out the effort.

Yes, every user (read software-engineer) need generate GitHub token and put it into bundle config. 

> Not opposed to it, just want to call out the effort.
Agree, this will require significant effort.

> Any reason to not publish it?
I'd like to discuss it online if you don't mind because this PR is already merged (Wednesday's sync?)

seanstory: > I'd like to discuss it online if you don't mind because this PR is already merged (Wednesday's sync?)

sounds good - added to the agenda

PR #159, opened 2022-07-07
Change connector interface

artem-shelkovnikov:
A simplification step - connector instead of calling sink directly, can return documents (or in future events about documents), and the framework will use its own internals and settings to push the data into desired sink.

Change in a nutshell:

Before:

Connector decides how, when and where to ingest.

```ruby
def sync(settings)
    GitLabClient.get_projects.each |do| project
        @sink.ingest(project)
    end
end
```

After:

Connector returns the documents to ingest, delegating ingestion itself to the framework

```ruby
def yield_documents(settings)
    GitLabClient.get_projects.each |do| project
        yield project
    end
end
```

Upcoming future:

Connector returns events of sync - whether the document was created, updated or deleted, delegating further index operations to the framework.

```ruby
def yield_documents(settings)
    GitLabClient.get_projects.each |do| project
        yield :upsert project
    end
end
```

### DIFF
--- a/lib/connectors/gitlab/connector.rb
+++ b/lib/connectors/gitlab/connector.rb
@@ -68,12 +68,10 @@ module Connectors
         Connectors::GitLab::CustomClient::ClientError
       end
 
-      def yield_projects
+      def yield_projects(&block)
         next_page_link = nil
         loop do
-          next_page_link = @extractor.yield_projects_page(next_page_link) do |projects_chunk|
-            yield projects_chunk
-          end
+          next_page_link = @extractor.yield_projects_page(next_page_link, &block)
           break unless next_page_link.present?
         end
       end
--- a/lib/connectors/gitlab/connector.rb
+++ b/lib/connectors/gitlab/connector.rb
@@ -81,6 +81,10 @@ module Connectors
       def yield_project_files(projects_chunk)
         projects_chunk.each_with_index do |project, idx|
           project = project.with_indifferent_access
+
+          chunk_size = projects_chunk.size
+          Utility::Logger.info("Fetching files for project #{project[:id]} (#{idx + 1} out of #{chunk_size})...")
+
           files = @extractor.fetch_project_repository_files(project[:id])
 
           yield files
--- a/lib/connectors/gitlab/connector.rb
+++ b/lib/connectors/gitlab/connector.rb
@@ -22,10 +22,9 @@ module Connectors
       def initialize
         super()
         @extractor = Connectors::GitLab::Extractor.new(
-          :base_url => configurable_fields[:base_url][:value],
-          :api_token => configurable_fields[:api_token][:value]
+          :base_url => configurable_fields['base_url'][:value],
+          :api_token => configurable_fields['api_token'][:value]
         )
-        @sink = nil # later set in sync
       end
 
       def display_name
@@ -34,11 +33,11 @@ module Connectors
 
       def configurable_fields
         @configurable_fields ||= {
-          :api_token => {
+          'api_token' => {
             :label => 'API Token',
             :value => App::Config[:gitlab][:api_token]
           },
-          :base_url => {
+          'base_url' => {
             :label => 'Base URL',
             :value => App::Config[:gitlab][:api_base_url] || Connectors::GitLab::DEFAULT_BASE_URL
           }
--- a/lib/connectors/gitlab/connector.rb
+++ b/lib/connectors/gitlab/connector.rb
@@ -33,11 +33,11 @@ module Connectors
 
       def configurable_fields
         @configurable_fields ||= {
-          'api_token' => {
+          :api_token => {
             :label => 'API Token',
             :value => App::Config[:gitlab][:api_token]
           },
-          'base_url' => {
+          :base_url => {
             :label => 'Base URL',
             :value => App::Config[:gitlab][:api_base_url] || Connectors::GitLab::DEFAULT_BASE_URL
           }
--- a/lib/connectors/gitlab/connector.rb
+++ b/lib/connectors/gitlab/connector.rb
@@ -44,18 +44,18 @@ module Connectors
         }
       end
 
-      def sync(connector_settings)
-        super(connector_settings)
+      def yield_documents(_connector_settings)
+        yield_projects do |projects_chunk|
+          projects_chunk.each do |project|
+            yield Connectors::GitLab::Adapter.to_es_document(:project, project)
 
-        extract_projects
-      end
-
-      def deleted(_params = {})
-        []
-      end
-
-      def permissions(_params = {})
-        []
+            yield_project_files(projects_chunk) do |files|
+              files.each do |file|
+                yield Connectors::GitLab::Adapter.to_es_document(:file, file)
+              end
+            end
+          end
+        end
       end
 
       private
@@ -68,36 +68,23 @@ module Connectors
         Connectors::GitLab::CustomClient::ClientError
       end
 
-      def extract_projects
+      def yield_projects
         next_page_link = nil
         loop do
           next_page_link = @extractor.yield_projects_page(next_page_link) do |projects_chunk|
-            projects = projects_chunk.map { |p| Connectors::GitLab::Adapter.to_es_document(:project, p) }
-            @sink.ingest_multiple(projects)
-            extract_project_files(projects_chunk)
+            yield projects_chunk
           end
           break unless next_page_link.present?
         end
-      rescue StandardError => e
-        puts(e.message)
-        puts(e.backtrace)
-        raise e
       end
 
-      def extract_project_files(projects_chunk)
+      def yield_project_files(projects_chunk)
         projects_chunk.each_with_index do |project, idx|
           project = project.with_indifferent_access
           files = @extractor.fetch_project_repository_files(project[:id])
-          chunk_size = projects_chunk.size
-          puts("Fetching files for project #{project[:id]} (#{idx + 1} out of #{chunk_size})...")
-          files = files.map { |file| Connectors::GitLab::Adapter.to_es_document(:file, file) }
-          project[:files] = files
-          @sink.ingest_multiple(files)
+
+          yield files
         end
-      rescue StandardError => e
-        puts(e.message)
-        puts(e.backtrace)
-        raise e
       end
     end
   end
--- a/lib/connectors/gitlab/connector.rb
+++ b/lib/connectors/gitlab/connector.rb
@@ -22,8 +22,8 @@ module Connectors
       def initialize
         super()
         @extractor = Connectors::GitLab::Extractor.new(
-          :base_url => configurable_fields['base_url'][:value],
-          :api_token => configurable_fields['api_token'][:value]
+          :base_url => configurable_fields[:base_url][:value],
+          :api_token => configurable_fields[:api_token][:value]
         )
       end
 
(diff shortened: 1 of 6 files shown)


wangch079:
I have a different view here

> After:
> 
> Connector returns the documents to ingest, delegating ingestion itself to the framework
> 
> ```ruby
> def yield_documents(settings)
>     GitLabClient.get_projects.each |do| project
>         yield project
>     end
> end
> ```

This is exactly the `extractor`'s job. Why do we need to wrap `extractor` with `connector`?

Our current `connector` holds reference to an `extractor` and a `sink` (BTW, we should move the below code to `base`):

https://github.com/elastic/connectors/blob/ff7fad61b4804546f18bea933b1662fd7a3e29b8/lib/connectors/gitlab/connector.rb#L22-L29


I think this makes perfect sense. A `connector`'s job is to 1) extracting data from 3rd-party source via `@extractor` 2) indexing data to Elasticsearch via `@sink`

artem-shelkovnikov:
> This is exactly the extractor's job. Why do we need to wrap extractor with connector?

The thought process here is more about taking extractor and merging it into the connector as there is no particular need to why Sinks should be exposed at all to the user.

So this PR has 2 goals:

1. Simplify `connector` class - do not expose sink to the connector directly. You'll still be able to choose which sinks you want to use via a config option `output: [ elasticsearch, console ]`. I just don't see the benefit of exposing it directly, neither I see the necessity of following the pattern of "extractor" and "adapter" - it works nice for some connectors, for some it's just not making much sense.

2. Work around the flaws of batching approach taken.
In a nutshell, the problem we have is: Sink exposes methods `ingest` and `ingest_multiple` that hide batching inside - if you ingest documents above threshold (let's say you ingest 50 docs), then the docs are sent in a single batch to Elasticsearch.

If the connector ingests 60 documents and bails, last 10 documents are never actually sent to Elasticsearch - something in the framework has to call `sink.flush` after the sync has ended OR there should be a better mechanism to do so. Right now it's just a bug that we have, and my approach (in further PRs after this) would have been to just call `@sink.flush` in the class that creates/manages the sending of the documents to sinks, and this would never be exposed to the creators/users of the connectors.

Also exposing `flush` from the sink classes as a part of public interface exposes the users of the sink to unnecessary complexity.

So to me this change is purely engineering change - reduce connector complexity without reducing the featureset + fix a problem with batching as a concept by moving it outside of connector into higher-level abstraction.

mchernyavskaya:
> Our current `connector` holds reference to an `extractor` and a `sink` (BTW, we should move the below code to `base`)

To me, our class model is a bit overcomplicated. There's 5 classes - connector, extractor, adapter, config, custom client - and all just to get one batch of documents. And now, we're adding the sink, too, so there's actually 6.

While I am for keeping the classes that make most sense as separate units (due to separation of concerns) like, for example, the adapter or the sink, I am for trying to merge together the classes that actually serve the same purpose. So if we can combine connector, extractor and config, or connector + extractor and client + config (and IMHO both are acceptable options), let's by any means do so.

One more observation. While writing a blog post about Connectors 2.0, I found it **really** difficult to explain why all those classes were required and how they interact between themselves. We should think of the external devs, and streamlining the adoption of this tool as best as we can. So to me, this is the step in the right direction.

lib/connectors/gitlab/connector.rb
```diff
@@ -22,10 +22,9 @@ class Connector < Connectors::Base::Connector
       def initialize
         super()
         @extractor = Connectors::GitLab::Extractor.new(
-          :base_url => configurable_fields[:base_url][:value],
-          :api_token => configurable_fields[:api_token][:value]
+          :base_url => configurable_fields['base_url'][:value],
```
mchernyavskaya, line 25: why not symbols?

lib/connectors/gitlab/connector.rb
```diff
@@ -34,29 +33,29 @@ def display_name
 
       def configurable_fields
         @configurable_fields ||= {
-          :api_token => {
+          'api_token' => {
```
mchernyavskaya, line 36: I find symbols a bit easier to work with for constants than just strings. Why do we need to replace them?

lib/connectors/gitlab/connector.rb
artem-shelkovnikov: That was my lousy attempt to fix the broken configurable fields validation, I've attempted to do the fix in https://github.com/elastic/connectors/pull/166 instead and will revert this change

lib/connectors/gitlab/connector.rb
artem-shelkovnikov: See my comment above: https://github.com/elastic/connectors/pull/159#discussion_r916761741

mchernyavskaya (approved):
Looks good. I think this is the step in the right direction.

wangch079:
>To me, our class model is a bit overcomplicated. There's 5 classes - connector, extractor, adapter, config, custom client - and all just to get one batch of documents. And now, we're adding the sink, too, so there's actually 6.

I don't think we need `config` any more, at least for now. I do think `adapter` and `custom_client` (in case there's no official Ruby client provided for the source) make sense. `extractor` and `sink` serve totally different purposes, one to extract data from remote source, and one to index data to Elasticsearch, so we should keep them. I see @artem-shelkovnikov moved `sink` from `connector` to `SyncJobRunner`, and I totally agree, as this is general for all connectors, and it does not need to be repeated in each connector. 

But now I see `connector` as just a wrapper of `extractor`. Do we really need it?

To summarise, we have 5 components in our connector service. 

2 of them are general across all connectors

1. `SyncJobRunner`, to run a sync job, managing 1) connector settings 2) extractor 3) sinks
2. `Sink`, to index data to Elasticsearch

The other 3 are connector specific
1. `Extractor`, to extract data from remote source.
2. `Adapter`, transform raw data to Elasticsearch document
3. `CustomClient`, only needed when there's no official Ruby client for the remote source.

artem-shelkovnikov:
> But now I see connector as just a wrapper of extractor. Do we really need it?

It's a good point, and I think it deserves a good discussion!

So far the idea that I have about it is the following:

`*Connector` classes should be a super simple user-facing interface that has to be implemented so that the connector can work. These classes can use more complex abstractions (like custom clients, adapters), or can be as simple as possible just implementing the logic right in the `*Connector` class, like StubConnector does.

`Adapter` and `CustomClient` classes are not mandatory - you can implement connector without them, or you can choose a different way to approach the tasks that they do, if you need to. For example, for some connectors it's more convenient to unite the data fetching and mapping to output format in one class, e.g.:

```ruby
def yield_users
    data = sql_client.execute("
        SELECT u.Id, u.Name, a.Country as address_country, a.State as address_state
        FROM Users u INNER JOIN Addresses a on u.address_id = a.id
    "
    
    mapped_data = data.map do |row|
       {
          :id => row[:Id],
          :name => row[:Name],
          :address => "#{row[:address_country]} - #{row[:address_state]}"
       }
    end
end
```

What I particularly like here, that it's easy to see both data access and the format that it will be returned in, which is not possible with `Extractor`/`Adapter` separation, and in some cases it's just making life much much easier.

As for `Extractor` class itself, I think its responsibilities will be split between multiple places - part of it already became `Connector` - e.g. `yield_documents` method. Some logic, such as retry/error handling will go either to `SyncJobRunner` or to separate classes, for example `RetryPolicy` class and etc. In the end I don't think `Extractor` abstraction will stay, it will be abolished by other classes.

artem-shelkovnikov:
> But I do think its worth keeping the yield_documents concern separated from as many other concerns as we can.
Absolutely agreed, +1000 on that.

> And we may eventually need a way to recommend how to display documents for each service type. And we may eventually need a place that specifies which features are available for this service type (full text extraction? thumbnailing? delete syncs? incremental syncs? DLP? etc).

Yes, absolutely agreed and we'll have to approach it again - probably salvaging major pieces of existing code. I'm currently also thinking whether it should be an inheritance-style approach, or some sort of a composition, like when you init a connector you specify some features of it in the constructor or separate method as its behaviour modifications:

```ruby
def initialize
    use_retry_policy(Connectors::RetryPolicy::ExponentialBackoff.new)
    use_throttling_policy(Connectors::ThrottlingPolicy::HeaderTimeoutPolicy.new(header: "x-try-again-in")
    use_authentication(Connectors::Authentication::OAuth.new(base_url: "some_url", token_url: "other_url"))
end
```

Or maybe it'll be super different from that, who knows? What's important to me is isolating these features in classes that can be used or can be fully ignored if you don't really need them - or for example when you're implementing a new connector, you probably wanna first focus on the data ingestion, then on the rest.

> In Workplace Search, all of these concerns fell in the ContentSourceService classes. In Connectors 2.0, we put all of this either in their own classes (Authorization) which were joined up in the Connector class, or in the Connector class directly.

That's true, and also there were SecretKeeper classes for secret management - we'll need to revisit these concepts too.

> If we move yield_documents into the Connector and do away with the Extractor class, I think this will end up looking a lot muddier to a potential implementor.

It can, though I think it's best to address it as we add features to the connectors, and it'll naturally lead us to cleaner design, I find it hard to plan such things upfront without having specific use-cases for this.

> (1) we don't need if we drop the delete sync, or the way it was implemented in Workplace Search. (2) we don't strictly need, as app search is much less opinionated about field names. (3) we probably shouldn't lose, but we could just add this logic in Elastic::Utility so that it's available.

I think the features that you've mentioned are something that we need. (2) indeed probably can be dropped, though we still impose some restriction, for example document needs to have an `id` field. (1) is still very usable and can be done in a different part of the framework, though also it's nice to give user direct ability to generate ids for the documents or use our built-in functions.

> Having the ability to run connector_instance.client in a REPL and then start making ad-hoc requests is a massive productivity booster.

I think we don't have REPL now and it's quite a loss - I totally agree that the class structure that we come up with should be usable via REPL without lots of pain, let's plan for it accordingly, I think productivity-related changes are ultra important as they not only improve our performance, but also help connector authors a lot.


I think the concerns that you've specified are very real and should be addressed - maybe in a little bit of a different class structure than it is now, which will shape out as we implement things and reflect on things that are implemented.

seanstory:
>  I'm currently also thinking whether it should be an inheritance-style approach, or some sort of a composition, like when you init a connector you specify some features of it in the constructor or separate method as its behaviour modifications:

I think this is a good train of thought. What if we utilized a builder pattern, so that a default implementation had a default set of composable units, but you could easily override them?

```
connector_instance = ConnectorBuilder.new.build
```

effectively gives you a stub connector, and

```
connector_instance = ConnectorBuilder.new
    .withRetryPolicy(MyCustomRetry.new)
    .withThrottlingPolicy(MyCustomThrottling.new)
    .withAuthentication(Connectors::OAuth.new)
    .withExtractorClass(Connectors::Gitlab)
    .build
```

gives you something much more tailored? We could even represent such a builder in YAML DSL so that you can configure your connector from components as part of your configuration (no code).

> I think we don't have REPL now and it's quite a loss

I'd forgotten this got lost. We used to have `make console` - can we just add that back?

> I think the concerns that you've specified are very real and should be addressed - maybe in a little bit of a different class structure than it is now, which will shape out as we implement things and reflect on things that are implemented.

My preference would be for us to start with the class structure we're familiar with, and to incrementally shift to the different class structure you envision. I have moderate concerns that if we start from scratch and only add in the old pieces when we think we need them:
a) we'll only add them after a customer points out they're missing, which leaves a negative impression for that customer
b) we'll likely forget about some pieces, or re-integrate them incorrectly (introducing new bugs).

The risks seem lower, to me, to start with the existing, vetted, solution. But this is strictly my opinion, and I don't think that the from scratch approach is "wrong."

PR #160, opened 2022-07-07
Add connector configuration initialization

artem-shelkovnikov:
Now when connector starts up and no configuration is there, connector just runs as is.

What happens now instead is - connector starts up and writes its default configuration into Elasticsearch if none is provided.

This way Kibana can show configurable fields to the user, so that they can do the initial connector setup.

### DIFF
--- a/lib/core/connector_settings.rb
+++ b/lib/core/connector_settings.rb
@@ -42,7 +42,6 @@ module Core
     end
 
     def configuration_initialized?
-      return false if configuration.nil?
       configuration.present?
     end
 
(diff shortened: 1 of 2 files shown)


wangch079 (approved):
I remembered I have this implemented before, but maybe it's removed accidentally when refactoring.

BTW, the first line is not necessary here
https://github.com/elastic/connectors/blob/fc90f9524c2ceec6afd38a0fcc63d784f5baa527/lib/core/connector_settings.rb#L44-L47

PR #161, opened 2022-07-07
Refactor sink.rb

artem-shelkovnikov:
Things done in this PR:

- Base class for sinks
- Each sink implementation has its own file
- Moved sinks to `core/output_sink` namespace


lib/core/output_sink/console_sink.rb
```diff
@@ -0,0 +1,45 @@
...
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+require 'core/output_sink'
+require 'utility/logger'
+
+module Core::OutputSink
+  class ElasticSink < Core::OutputSink::BaseSink
```
wangch079, line 13:
```suggestion
  class ConsoleSink < Core::OutputSink::BaseSink
```

lib/core/output_sink/console_sink.rb
```diff
@@ -0,0 +1,45 @@
...
+    def ingest(_document)
+      print_header 'Got a single document:'
+    end
```
wangch079, line 16:
```suggestion
    def ingest(document)
      print_header 'Got a single document:'
      Utility::Logger.info document
    end
```

To be consistent with `ingest_multiple(documents)

wangch079 (approved):
LGTM except for a typo.

Agree with what @seanstory suggested. `Elastic` is the company name, we should rename it. We have `EsClient`, so I think either we rename both to `ElasticsearchClient` and `ElasticsearchSink` or we use `EsClient` or `EsSink`, to be consistent.

lib/core/output_sink/console_sink.rb
artem-shelkovnikov: Good call, fixed!

artem-shelkovnikov:
> Do these have specs?

Not yet, planning to add basic tests to everything as the last set of tasks for connector, now focussing on functionality and ability to throw away/rewrite major pieces of code quickly.

I'm a big fan of tests, so I'll close as much as possible in the last week before release :)

artem-shelkovnikov:
> Agree with what @seanstory suggested. Elastic is the company name, we should rename it. We have EsClient, so I think either we rename both to ElasticsearchClient and ElasticsearchSink or we use EsClient or EsSink, to be consistent.

Agreed, renamed to EsSink to be consistent with EsClient. I like `ElasticsearchX` more, but this can be renamed later.

PR #164, opened 2022-07-08
Fix another place for Core::OutputSink

artem-shelkovnikov:
Fix another merge conflict - Base::Connector does not actually require `core/output_sink` + fix wrong way to instantiate `CombinedSink`.


mchernyavskaya (approved):
LGTM

PR #165, opened 2022-07-08
Move update_connector_field to the correct scope

artem-shelkovnikov:
`update_connector_field` method is outside of `class << self`, thus not available to class methods of ElasticConnectorActions.

This PR fixes this problem by moving `update_connector_field` inside the `class << self` scope.


mchernyavskaya (approved):
LGTM

PR #166, opened 2022-07-08
Make validate_configuration! not fail if configuration is validated in symbols

artem-shelkovnikov:
When just starting up a connector, an error message is displayed:

```
/Users/artemshelkovnikov/git_tree/connectors/lib/core/sync_job_runner.rb:49:in `validate_configuration!': Connector expected configurable fields: [:api_token, :base_url], actual stored fields: ["api_token", "base_url"] (Core::IncompatibleConfigurableFieldsError)
```

It happens due to the fact that it's possible to declare configurable fields both way - as a symbol and as a string. When these values are de-serialized from the database, they are always stringified.

This commit aims to fix the validation, by stringifying the field names when validating them.

### DIFF
--- a/lib/core/sync_job_runner.rb
+++ b/lib/core/sync_job_runner.rb
@@ -43,8 +43,8 @@ module Core
     private
 
     def validate_configuration!
-      expected_fields = @connector_instance.configurable_fields.keys.map { |field| field.to_s }
-      actual_fields = @connector_settings.configuration.keys.map { |field| field.to_s }
+      expected_fields = @connector_instance.configurable_fields.keys.map(&:to_s)
+      actual_fields = @connector_settings.configuration.keys.map(&:to_s)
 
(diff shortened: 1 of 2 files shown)


lib/core/sync_job_runner.rb
```diff
@@ -43,8 +43,8 @@ def execute
     private
 
     def validate_configuration!
-      expected_fields = @connector_instance.configurable_fields.keys
-      actual_fields = @connector_settings.configuration.keys
+      expected_fields = @connector_instance.configurable_fields.keys.map(&:to_s)
```
mchernyavskaya, line 46: Could we maybe do the `.with_indifferent_access` conversion in the `@connector_instance.configurable_fields` base method? Or add the `.with_indifferent_access` here?

artem-shelkovnikov: Also added .with_indifferent_access in https://github.com/elastic/connectors/pull/166/commits/f7b54acca15a595920aae7efafd133d03bb05e9f

PR #168, opened 2022-07-08
Align with connectors gem versioning

vidok:
It makes `connectors_utility` compatible with ES cross-versioning.


artem-shelkovnikov (approved):

mchernyavskaya (approved):
LGTM

PR #169, opened 2022-07-11
Sort fields when doing validation of config

artem-shelkovnikov:
Fixes the problem when validating the config:

Elasticsearch stores configuration without keeping the original sorting of the config fields, therefore it can re-sort them.

Error message that surfaces it:

```
Connector expected configurable fields: ["host", "database", "collection"], actual stored fields: ["database", "host", "collection"] (Core::IncompatibleConfigurableFieldsError)
```

This commit makes sure that we sort the arrays of fields before comparing them against each other.


wangch079 (commented):
I think our design introduced an issue for FE too. Kibana won't know which field to display first.
Instead of sorting by string, we should probably introduce an additional `order` in the `configurable_fields`. cc @sphilipse

artem-shelkovnikov:
cc @danajuratoni @serenachou @DianaJourdan - for now we don't have ordering of configurable fields working by design of hashes and the way we store the data - if the user defines fields like `hostname`, `username`, `password` they will be sorted arbitrarily in the frontend. 

It's not related to this PR, but this is suboptimal UX. Do we want to address it in 8.4?

seanstory:
> this is suboptimal UX

Just my 2cents, I don't think this is a huge deal, as long as the display is consistent and doesn't re-order on refresh or anything. Displaying config fields in alphabetical order seems reasonable to me.

wangch079 (approved):
> Just my 2cents, I don't think this is a huge deal, as long as the display is consistent and doesn't re-order on refresh or anything. Displaying config fields in alphabetical order seems reasonable to me.

@sphilipse had the same [thought](https://elastic.slack.com/archives/C03GAPTQ2JF/p1657634966385279?thread_ts=1657282529.571969&cid=C03GAPTQ2JF). Now I'd vote we proceed with alphabetical order in 8.4.

PR #170, opened 2022-07-11
Make display_name and configurable_fields a part of class instead of instance

artem-shelkovnikov:
Both methods are used as instance methods, while in reality they don't need to have access to the instance of `Connector` class - moreover they can enforce mis-usage of these methods.

Therefore, moving `display_name` and `configurable_fields` to be methods on the instance of class.

### DIFF
--- a/lib/connectors/base/connector.rb
+++ b/lib/connectors/base/connector.rb
@@ -21,8 +21,8 @@ module Connectors
         {}
       end
 
-      def service_type
-        self.class::SERVICE_TYPE
+      def self.service_type
+        raise 'Not implemented for this connector'
       end
 
       def yield_documents(connector_settings); end
--- a/lib/connectors/base/connector.rb
+++ b/lib/connectors/base/connector.rb
@@ -13,6 +13,18 @@ require 'utility/logger'
 module Connectors
   module Base
     class Connector
+      def self.display_name
+        raise 'Not implemented for this connector'
+      end
+
+      def self.configurable_fields
+        {}
+      end
+
+      def service_type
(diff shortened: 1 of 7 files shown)


lib/connectors/base/connector.rb
```diff
@@ -13,25 +13,25 @@
...
       end
 
-      def display_name
-        raise 'Not implemented for this connector'
+      def self.configurable_fields
+        {}
       end
 
       def service_type
         self.class::SERVICE_TYPE
```
wangch079, line 25: This is confusing. Why not make it a static method `service_type` in each connector, and remove the Constant `SERVICE_TYPE`?

artem-shelkovnikov: Makes sense, will do that!

artem-shelkovnikov: Fixed in https://github.com/elastic/connectors/pull/170/commits/2902f4fde1d3e5006887567b8f0b0c10800793a0

wangch079 (approved):
LGTM

lib/connectors/base/connector.rb
seanstory: I'm ambivalent, but just wanted to share...

> Why not make it a static method service_type in each connector, and remove the Constant SERVICE_TYPE?

Adam Harwood was always a fan of using functions over constants. I asked him about this once, and he explained that it's a lot more clear to override a function in a child class than it is to override a constant. I thought this was an interesting pattern, as I hadn't seen it before. 

:homerdisappear:

PR #177, opened 2022-07-13
Implement byoc interface status updates

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/2268

Implements the logic for updating the connector `status` field:

1. When connector first initialises configurable fields, set the status to NEEDS_CONFIGURATION
2. When connector starts a sync it sets the status to CONNECTED
3. When connector fails a sync it sets the status to ERROR

⚠️ The logic above might be inaccurate, let's discuss!


lib/connectors/connector_status.rb
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+module Connectors
+  class ConnectorStatus
```
seanstory, line 10: nit - with `module Connectors`, the class name of `ConnectorStatus` seems redundant. Could it be `Connectors:Status` instead of `Connectors::ConnectorStatus`?

EDIT: ah, I see there's also a `sync_status` in the same dir. Ok, that makes more sense.

seanstory (approved):
lgtm

lib/core/elastic_connector_actions.rb
```diff
@@ -86,6 +88,16 @@ def claim_job(connector_package_id)
         job['_id']
       end
 
+      def update_connector_status(connector_package_id, status)
```
seanstory, line 91: Should there also be a `get_connector_status` function?

lib/core/sync_job_runner.rb
```diff
@@ -33,13 +34,14 @@ def initialize(connector_settings, service_type)
 
     def execute
       unless @connector_settings.configuration_initialized?
-        @connector_settings.update_configuration(@connector_instance.configurable_fields)
+        @connector_settings.initialize_configuration(@connector_instance.configurable_fields)
       end
 
       validate_configuration!
       return unless should_sync?
```
seanstory, line 41: Should this (or `should_sync?`) also check to ensure that the current status is not `needs_configuration`?

seanstory (commented):
had a follow-up thought

lib/core/sync_job_runner.rb
artem-shelkovnikov: Good call, though I need to double check it with @sphilipse.

So connector says that it `needs_configuration`, what will be the status of connector after user updates the config in Kibana?

sphilipse: Status will be `configured`

lib/core/connector_settings.rb
```diff
@@ -50,8 +50,12 @@ def configuration_initialized?
       configuration.present?
     end
 
-    def update_configuration(configuration)
+    def initialize_configuration(configuration)
+      # TODO: actually check it on the level lower, so that desync does not happen if configuration was updated elsewhere
+      raise 'Configuration is already initialized!' if configuration_initialized?
+
       ElasticConnectorActions.update_connector_configuration(id, configuration) # TODO: I don't like that it's hidden here now
+      ElasticConnectorActions.update_connector_status(id, Connectors::ConnectorStatus::NEEDS_CONFIGURATION)
```
wangch079, line 58: This can be one `write`, and the underlying `update_connector_field` method should be renamed to `update_connector_fields` and accept arrays. WDYT?

wangch079 (commented):
I see the logic for `Connected` and `Error` is different from PR description.

It tests the connection to the remote source. The current implementation is to update the `status` only when the job runs. Do you think we should update in every loop (even when `should_sync? == false`)?
Or maybe we can move this to the heartbeat logic? WDYT @artem-shelkovnikov

PR #180, opened 2022-07-14
added new language options to configuration validator

mchernyavskaya:
#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/connectors-ruby/pull/172

### DIFF
--- a/lib/app/config.rb
+++ b/lib/app/config.rb
@@ -36,6 +36,9 @@ puts "Parsing #{CONFIG_FILE} configuration file."
     optional(:gitlab).hash do
       required(:api_token).value(:string)
     end
+
+    required(:use_analysis_icu).value(:bool)
+    required(:content_language_code).value(:string)
   end
 end
 


lib/app/config.rb
```diff
@@ -36,6 +36,9 @@
     optional(:gitlab).hash do
       required(:api_token).value(:string)
     end
+
+    required(:use_analysis_icu).value(:bool)
+    required(:content_language_code).value(:string)
```
artem-shelkovnikov, line 41: Hmmmm I thought this is actually a part of connector configuration - I saw that you specify language in UI when you create a connector, or is it a different thing?

mchernyavskaya: Yeah, it's the same thing. It's needed for registering connectors via CLI, bypassing Kibana.

mchernyavskaya: On second thought, I think I'll make them optional. Since the CLI won't be the "official" way of creating connectors. So this PR is not needed. =)

PR #181, opened 2022-07-14
Security improvement - moving api key out of configurable fields

mchernyavskaya:
In terms of security, having an `api_token` or any credentials in the configurable fields is now a possible security breach, since we store this data in elastic as plain text. This means that the fields that have confidential information should instead be stored in a local config file, and that file should NOT be under source control (so our `config/connectors.yml` is good).

Changes in this PR:

* moving gitlab api key out of configurable fields
* creating a `local_configuration` method on base connector to read connector specific props from yaml

#### Pre-Review Checklist

- [x] New configuration settings added in this PR follow the [official guidelines](https://github.com/elastic/ent-search/blob/main/doc/enterprise-search-config.md)

### DIFF
--- a/lib/connectors/gitlab/connector.rb
+++ b/lib/connectors/gitlab/connector.rb
@@ -28,7 +28,7 @@ module Connectors
         {
           :base_url => {
             :label => 'Base URL',
-            :value => @local_configuration[:api_base_url] || Connectors::GitLab::DEFAULT_BASE_URL
+            :value => Connectors::GitLab::DEFAULT_BASE_URL
           }
         }
       end
@@ -37,7 +37,7 @@ module Connectors
         super(local_configuration)
         @extractor = Connectors::GitLab::Extractor.new(
           :base_url => self.class.configurable_fields[:base_url][:value],
-          :api_token => self.class.configurable_fields[:api_token][:value]
+          :api_token => @local_configuration[:api_token]
         )
       end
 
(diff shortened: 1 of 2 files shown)


lib/connectors/base/connector.rb
```diff
@@ -33,6 +34,12 @@ def service_type
       def configurable_fields
         {}
       end
+
+      # connector-specific configuration that comes from the local yaml config file
+      # should be stored under the section that has the same name as the connector service_type
+      def local_configuration
+        App::Config[service_type] || {}
```
artem-shelkovnikov, line 41: I believe it should be passed as a parameter into the constructor to avoid having a `App::Config` usage outside of `App` module as we've discussed previously

mchernyavskaya: I think this way might be acceptable too since it's in the base class, so there's supposed to just be one usage? It looks a bit cleaner to me. Yeah we restrict the user a little because we're determining the way the settings should be stored, but in the end, isn't it better than invent a new wheel for every connector?
Or do you mean reading and passing these directly on `Connector.new`? As in, in the Registry?

artem-shelkovnikov: > Or do you mean reading and passing these directly on Connector.new? As in, in the Registry?

Yup, this part. The framework that creates the connector with `Connector.new` can read the config and pass it inside the connector, I feel it's a bit cleaner this way

mchernyavskaya: Agreed. Refactoring.

PR #182, opened 2022-07-14
Provide connector_settings in initializer instead of yield_documents

artem-shelkovnikov:
Improvement that is needed to actually initialize extractors and other classes when connector is initialized.

What happened before:

```ruby
class Connector
  def initialize(local_configuration:)
    @local_configuration = local_configuration || {} # these are the data from config file for the specified connector
  end
  
  def yield_documents(connector_settings) # these connector_settings are the settings coming from the Elasticsearch index
    # do something
  end
end
```

What changes in this PR:

```ruby
class Connector
  def initialize(local_configuration:, connector_settings:)
    @local_configuration = local_configuration || {} # these are the data from config file for the specified connector
    @connector_settings = connector_settings # these connector_settings are the settings coming from the Elasticsearch 
  end
  
  def yield_documents
    # do something
  end
end
```

What this allows in essence is to provide everything that's needed for connector operations upfront, while now it happens only when `yield_document` is ran.


lib/connectors/base/connector.rb
```diff
@@ -26,13 +26,14 @@ def self.service_type
         raise 'Not implemented for this connector'
       end
 
-      def initialize(local_configuration = {})
+      def initialize(local_configuration:, connector_settings:)
```
mchernyavskaya, line 29: I think I am having a hard time trying to understand what should be there in the connector_settings. Are those the settings that are persisted in elasticsearch? The configurable fields? Or something else? I think the naming could be a bit clearer.

lib/connectors/gitlab/connector.rb
```diff
@@ -33,15 +33,19 @@ def self.configurable_fields
         }
       end
 
-      def initialize(local_configuration = {})
-        super(local_configuration)
+      def initialize(local_configuration:, connector_settings:)
```
mchernyavskaya, line 36: The named parameter with no default value implies that we always need to specify it. What if we don't have any settings to store in the yaml? Or do we assume that there's at least some credentials, always?

mchernyavskaya: But it's making the code longer, not just by the `{}`, but by the whole length of `local_configuration: {}`. =)

mchernyavskaya: I think that if we decided to pass both in as empty hashes, it would be confusing if the params were not named. However, with named params and no defaults, it's just not providing an option of just calling a method with no params at all.

wangch079: Vote for 

```
def initialize(local_configuration: {}, connector_settings: {})
```

lib/connectors/stub_connector/connector.rb
```diff
@@ -29,11 +29,18 @@ def self.configurable_fields
         }
       end
 
+      def initialize(local_configuration:, connector_settings:)
+        super(
+          local_configuration: local_configuration,
+          connector_settings: connector_settings
+        )
```
wangch079, line 36:
```suggestion
        super
```

lib/core/sync_job_runner.rb
```diff
@@ -25,7 +25,7 @@ def initialize(connector_settings, service_type)
       @connector_settings = connector_settings
       @sink = Core::OutputSink::EsSink.new(connector_settings.index_name)
       @connector_class = Connectors::REGISTRY.connector_class(service_type)
-      @connector_instance = Connectors::REGISTRY.connector(service_type)
+      @connector_instance = Connectors::REGISTRY.connector(service_type, connector_settings)
```
wangch079, line 28: It seems for a connector, it only needs `connector_settings.configuration`, instead of `connector_settings`

PR #183, opened 2022-07-14
Ad-hoc sync from CLI is now running once and returns control to CLI

mchernyavskaya:
Ad-hoc sync from CLI is now running once and returns control to CLI. It's not starting the job runner anymore, as that would prevent the user from ever getting back to the menu. So the flow becomes more CLI-like.

Initiated by @wangch079 's review of the CLI flow.

The cli app is now directly using the job runner's execute method and doesn't start the endless cycle.

#### Pre-Review Checklist

### DIFF
--- a/lib/app/console_app.rb
+++ b/lib/app/console_app.rb
@@ -26,8 +26,7 @@ module App
     INDEX_NAME_REGEXP = /[a-zA-Z]+[\d_\-a-zA-Z]*/
 
     @commands = [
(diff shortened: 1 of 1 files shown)


wangch079 (commented):
Great work @mchernyavskaya .

`Sync Now` in CLI will directly execute sync job and will not add an entry to `sync-jobs` index, and will not update the status of the connector. cc @danajuratoni in case there's any concern in PM's view.

PR #184, opened 2022-07-14
following up on repo rename

mchernyavskaya:
I think we missed a couple of places after renaming the repo.
Not all of them are required, but there might be less confusion this way.

#### Pre-Review Checklist


artem-shelkovnikov (approved):

wangch079 (approved):
LGTM, I think I will just merge it.

PR #186, opened 2022-07-14
Add a simple heartbeat task implementation

artem-shelkovnikov:
PR adds a simple task that is started when `App::Worker` starts and sends a heartbeat into `.elastic-connectors` index, updating `last_seen` field to current time.

If something makes the thread crash, it's not restarted - is there a way to handle it better?

### DIFF
--- a/lib/app/worker.rb
+++ b/lib/app/worker.rb
@@ -47,7 +47,7 @@ module App
           Utility::Logger.debug("Sending heartbeat for the connector")
           Core::ElasticConnectorActions.heartbeat(App::Config['connector_package_id'])
         rescue StandardError => e
-          Utility::ExceptionTracking.log_exception(e, 'Heartbeat timer terminated due to unexpected error.')
+          Utility::ExceptionTracking.log_exception(e, 'Heartbeat timer encountered unexpected error.')
         end
 
         task.execute
--- a/lib/app/worker.rb
+++ b/lib/app/worker.rb
@@ -41,7 +41,7 @@ module App
       end
 
       def start_heartbeat_task
-        interval_seconds = 5
+        interval_seconds = 60 # seconds
         Utility::Logger.debug("Starting heartbeat timer task with interval #{interval_seconds} seconds.")
         task = Concurrent::TimerTask.new(execution_interval: interval_seconds) do
           Utility::Logger.debug("Sending heartbeat for the connector")
--- a/lib/app/worker.rb
+++ b/lib/app/worker.rb
@@ -11,6 +11,7 @@ require 'connectors'
 require 'core'
 require 'utility'
 require 'app/config'
+require 'concurrent'
 
 module App
   module Worker
@@ -21,6 +22,7 @@ module App
         pre_flight_check
 
         Utility::Logger.info('Starting to process jobs...')
+        start_heartbeat_task
         start_polling_jobs
       end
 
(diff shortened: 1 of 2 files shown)


seanstory:
> If something makes the thread crash, it's not restarted - is there a way to handle it better?

Are there any scenarios where you'd expect the heartbeat thread to crash, but for the other threads/processes still to be 100% healthy? I would think we want it to stay dead if it dies, and use that as a signal that the whole process needs a restart (or the pod should be recycled, in the case of Docker).

lib/app/worker.rb
```diff
@@ -38,6 +40,19 @@ def pre_flight_check
...
       end
 
+      def start_heartbeat_task
+        interval_seconds = 60 # seconds
+        Utility::Logger.debug("Starting heartbeat timer task with interval #{interval_seconds} seconds.")
+        task = Concurrent::TimerTask.new(execution_interval: interval_seconds) do
+          Utility::Logger.debug("Sending heartbeat for the connector")
+          Core::ElasticConnectorActions.heartbeat(App::Config['connector_package_id'])
+        rescue StandardError => e
+          Utility::ExceptionTracking.log_exception(e, 'Heartbeat timer terminated due to unexpected error.')
```
seanstory, line 50: From reading in [the TimerTask rubydocs](https://ruby-concurrency.github.io/concurrent-ruby/1.1.4/Concurrent/TimerTask.html), I don't think catching this exception will actually terminate the task - it'll just log this message out, then run the task again in `interval_seconds`. I see three approaches:
- you could  pass the task as a param to the block like 
```
do |task|
```
and then shut it down in the rescue
```
rescue StandardError => e
  log_exception("message")
  task.shutdown
```  

- or you could attach an `Observer` that watches for errors and terminates the heartbeat task if it sees one.
- or you could just log the exception and move on, not trying to shut down the heartbeat

lib/core/elastic_connector_actions.rb
```diff
@@ -88,6 +88,16 @@ def claim_job(connector_package_id)
         job['_id']
       end
 
+      def heartbeat(connector_package_id)
```
seanstory, line 91: nit - the new taxonomy suggests we shouldn't use `connector_package` but `connector_client`

seanstory (approved):
LGTM, small comments

lib/app/worker.rb
wangch079: > it'll just log this message out, then run the task again in `interval_seconds`.
> 
> * or you could just log the exception and move on, not trying to shut down the heartbeat

I think this is expected behaviour?

wangch079 (commented):
Overall LGTM.

The `status` (`connected` and `error`) in `.elastic-connectors` now reflects whether we can connect to 3rd-party source, not the status of the sync job, I don't think they should exist in `Core::SyncJobRunner`:
https://github.com/elastic/connectors-ruby/blob/438067a25ab33a2a38b3f9c1ba123a021e2aa6b5/lib/core/sync_job_runner.rb#L42-L49

Do you think we can move it to heartbeat?

lib/core/elastic_connector_actions.rb
artem-shelkovnikov: Good one, I will actually rename `connector_package_*` to `connector_client_*` in a separate PR.

artem-shelkovnikov:
> Do you think we can move it to heartbeat?

I think it's a great call, I'll try to address it in a separate PR!

PR #187, opened 2022-07-14
Add validation that connector is configured for sync to go ahead

artem-shelkovnikov:
Every time connector checks if it should run, its configuration is loaded into `@connector_settings` variable.

We want connector to not even attempt to do anything if connector status is `NEEDS_CONFIGURATION`.

Food for thought: is it only for this status, or for any other? Intuitively, `CREATED` status cannot even happen at this moment (we wanna validate this!), status `CONFIGURED` means nothing to connector, `CONNECTED` and `ERROR` are working statuses for the connector.

cc @sphilipse


artem-shelkovnikov:
> I think we'd want connector to try to validate and then connect on every status that is not 'CREATED' or 'NEEDS_CONFIGURATION'.

Makes sense, I'll update the PR accordingly!

PR #188, opened 2022-07-18
Fix connector initialisation when connector settings are not set yet

artem-shelkovnikov:
Problem: we create an instance of `Connector` even before connector was configured. We also have initialisation in the connectors, for example:

```ruby
        @extractor = Connectors::GitLab::Extractor.new(
          :base_url => remote_configuration[:base_url][:value],
          :api_token => local_configuration[:api_token]
        )
```

Since `remote_configuration` comes from Kibana, when user sets the values, it might not yet be available.

Thus this commit changes such initialisation to use `dig` to not raise errors if config is not fully supplied:

```ruby 
        @extractor = Connectors::GitLab::Extractor.new(
          :base_url => remote_configuration.dig(:base_url, :value),
          :api_token => local_configuration[:api_token]
        )
```

### DIFF
--- a/lib/connectors/gitlab/connector.rb
+++ b/lib/connectors/gitlab/connector.rb
@@ -37,7 +37,7 @@ module Connectors
         super
 
         @extractor = Connectors::GitLab::Extractor.new(
-          :base_url => remote_configuration[:base_url][:value],
+          :base_url => remote_configuration.dig(:base_url, :value),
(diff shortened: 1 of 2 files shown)


wangch079 (approved):
>This PR on the contrary is just a simple fix to make it work for now.

Makes sense to me. Let's get it to work first.

PR #189, opened 2022-07-18
Simplify gitlab connector - remove file extraction

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/2269

We're simplifying Gitlab connector by removing the file traversal for now - it's only picking up the root of the repository and does not sync file content.

It makes sense to approach this later when we start addressing the attachment/download functionality.


wangch079 (approved):
LGTM

PR #190, opened 2022-07-18
Simplify EsSink and flush manually

artem-shelkovnikov:
How EsSink works now:

Every N seconds it tries to automatically flush all the queue of requests into Elasticsearch. It was done to work around the problem, that the last batch of data was never sent to Elasticsearch - `@sink.flush` was never used in the end of sync.

Now with all the refactorings, it's trivial to just call `@sink.flush` in the end of sync in `SyncJobRunner.do_sync!`.

This simplifies the sink design + avoids potential awkward interactions when the timer task to flush is triggered in the middle of actual flush (rare case, nevertheless this is added complexity that can backfire).

### DIFF
--- a/lib/core/output_sink/es_sink.rb
+++ b/lib/core/output_sink/es_sink.rb
@@ -7,7 +7,6 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/numeric/time'
-require 'concurrent-ruby'
 require 'core/output_sink/base_sink'
 require 'utility/es_client'
 require 'utility/logger'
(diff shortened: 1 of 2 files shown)


wangch079 (commented):
I remembered the original implementation of `flush` is a loop.

I think the end state of `flush` is an empty queue.

Because there's an `ingest_multiple` method, it is possible that:
1. `ingest_multiple` is called with 120 documents, and `flush` is called at the end of `ingest_multiple` resulting 70 documents left in the queue.
2. `flush` is called at the end of the sync, and another 50 documents are sent.
3. 20 documents are left in the queue.

`Core::SyncJobRunner` is using `ingest` method, so the above won't happen. But we should make `Core::OutputSink::EsSink` more robust.

artem-shelkovnikov:
@wangch079 good point. I think now we don't use `ingest_multiple` so the bug never appeared, but it's something to address indeed. 

I've tried to address it in this PR in commit https://github.com/elastic/connectors-ruby/pull/190/commits/fb1b9bc56b5eca1b9a42536962510a038c34f7cc.

wangch079 (approved):

PR #191, opened 2022-07-18
Make status of a sync local varable for do_sync! method

artem-shelkovnikov:
Minor refactoring: `SyncJobRunner` has instance variable `@status` that is used in only one method.

I think it makes sense to inline the variable into the method. It's also a small step into the direction of initialising the connector once and re-using same instance over the run-time of the connector worker, if needed.

### DIFF
--- a/lib/core/sync_job_runner.rb
+++ b/lib/core/sync_job_runner.rb
@@ -26,11 +26,6 @@ module Core
       @sink = Core::OutputSink::EsSink.new(connector_settings.index_name)
       @connector_class = Connectors::REGISTRY.connector_class(service_type)
       @connector_instance = Connectors::REGISTRY.connector(service_type, connector_settings.configuration)
-      @status = {
-        :indexed_document_count => 0,
-        :deleted_document_count => 0,
-        :error => nil
-      }
     end
 
     def execute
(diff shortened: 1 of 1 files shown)


wangch079 (commented):
I actually think it makes sense to leave it as an instance variable.

It's only used by `do_sync!` for now but consider code like this:
```
def do_sync!
  ...
  helper_method1
  ...
  helper_method2
  ...
end
```

Having `status` as an instance variable can make it usable directly in `helper_method1` and `helper_method2`

artem-shelkovnikov:
To me it adds some mental overhead to the method and class.

For example, this code will produce weird results:
```ruby
job_runner = SyncJobRunner.new(something, "gitlab")

job_runner.execute
job_runner.execute
```

Second sync will go well, but its stats will be including he stats from the first run.

We don't actually call SyncJobRunner this way, but we need to detach stats counting from syncs if we want to (e.g. create 1 SyncJobRunner when app starts and then re-use - consider endless MongoDB sync as an example, it needs connector running forever).

Taking this idea further, maybe it's also a good time to separate SyncJobRunner into multiple classes - one responsible for triggers, one responsible for individual sync run.

What do you think @wangch079

wangch079:
> ```ruby
> job_runner = SyncJobRunner.new(something, "gitlab")
> 
> job_runner.execute
> job_runner.execute
> ```

I don't expect a `SyncJobRunner` to run like this, and I don't expect a `SyncJobRunner` to be re-used. I treat `SyncJobRunner` as a class to run a sync job, it should be GCed once it completes the job. So `status` as an instance variable of an instance of `SyncJobRunner` makes more sense to me.

For the above case, maybe we should add a check to only allow `job_runner.execute` to run only once.

For a long-lived Mongo `SyncJobRunner`, will there be any problem to have `status` as an instance variable?

artem-shelkovnikov:
> For a long-lived Mongo SyncJobRunner, will there be any problem to have status as an instance variable?

Hmmm probably not, though I think the design will also be a bit different - maybe you can have an endless sync running + full sync running on top of it? In this case we'll have to even separate these jobs into different threads and class design will be quite different.

> For the above case, maybe we should add a check to only allow job_runner.execute to run only once.

It feels a bit artificial to me.

Now thinking of it, I'll postpone work in this PR and try a different approach - separate setup from scheduling and actual execution. This way the problem can go away by itself.

PR #194, opened 2022-07-19
Rename all connector package entries

artem-shelkovnikov:
Following the taxonomy, we're getting rid of Connector Package term in favour of Connector or Connector Client (depending on the place of the code).

This PR just renames all entries of Connector Package, connector_package and such.


lib/app/console_app.rb
```diff
@@ -39,7 +39,7 @@ module ConsoleApp
     def start_sync_now
       return unless connector_registered?
       puts 'Initiating synchronization NOW...'
-      connector_id = App::Config[:connector_package_id]
+      connector_id = App::Config[:connector_id]
```
tarekziade, line 42: fly-by review, the connector_id field is used everywhere, maybe we can add it as a class attribute when initializing, with

```
@connector_id = App::Config[:connector_id]
```

lib/app/console_app.rb
```diff
@@ -62,7 +62,7 @@ def show_status
     end
 
     def register_connector
-      id = App::Config['connector_package_id']
+      id = App::Config['connector_id']
```
tarekziade, line 65: you used the hash earlier. and now the string. we should use a single way

tarekziade (commented):
one thing to fix (hash vs string) o/wose looks good

README.md
```diff
@@ -2,7 +2,7 @@
 
 ![logo](logo-enterprise-search.png)
 
-The home of Elastic Enterprise Connector Packages. Use connector packages to
+The home of Elastic Enterprise Connector Clients. Use connector clients to
```
wangch079, line 5: Should do a search of `Enterprise` in the whole repo and clean up accordingly.

artem-shelkovnikov: What's our stance on `Enterprise`?

danajuratoni: I'd vote to keep it unless we hear otherwise. cc: @serenachou

wangch079: My understanding is the new connector has nothing to do with Enterprise Search, therefore I feel it's inappropriate to see `Enterprise` in this repo.

artem-shelkovnikov: @wangch079 We're still under "Enterprise Search" section in Kibana so far, afaik
image

lib/app/console_app.rb
artem-shelkovnikov: Good one, done in https://github.com/elastic/connectors-ruby/pull/194/commits/718867806e78c758c8d95ee4ae56106e8e23d4e4

lib/app/console_app.rb
artem-shelkovnikov: Addressed in https://github.com/elastic/connectors-ruby/pull/194/commits/718867806e78c758c8d95ee4ae56106e8e23d4e4

wangch079 (approved):

README.md
danajuratoni: Correct, the products are still part of Enterprise Search, and the decoupling that happens at a technical level is a different matter. Please keep the naming as it was assigned for Connectors 2.0 and we'll review it when the time comes.

PR #196, opened 2022-07-20
Fix registry.connector call in console_app.rb

artem-shelkovnikov:
Small problem found out when testing CLI - `registry.connector` after refactoring expects 2 arguments - connector type and connector configuration, this PR fixes it by loading connector settings first and providing connector configuration into `registry.connector` method.


lib/app/console_app.rb
```diff
@@ -151,9 +151,10 @@ def wait_for_keypress(message = nil)
     end
 
     def current_connector
+      connector_settings = Core::ConnectorSettings.fetch(App::Config['connector_package_id'])
```
wangch079, line 154: Need to rename to `connector_id` after https://github.com/elastic/connectors-ruby/pull/194 is merged.

wangch079 (approved):

PR #198, opened 2022-07-21
Move connector status update to heartbeat

wangch079:
Several changes:
1. Move the connector status to heartbeat
2. ~Fix a bug in method `fetch_document_ids`~
3. ~Fix a name typo in `es_sink.rb`~
4. Move the configuration update to heartbeat

#### Pre-Review Checklist

### DIFF
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -84,12 +84,18 @@ module Core
         job['_id']
       end
 
-      def heartbeat(connector_id)
+      def heartbeat(connector_id, service_type)
         body = {
           :doc => {
             :last_seen => Time.now
           }
         }
+
+        connector_setting = Core::ConnectorSettings.fetch(connector_id)
+        if connector_setting.connector_status_allows_sync?
+          connector_instance = Connectors::REGISTRY.connector(service_type, connector_setting.configuration)
+          body[:doc][:status] = connector_instance.source_status[:status] == 'OK' ? Connectors::ConnectorStatus::CONNECTED : Connectors::ConnectorStatus::ERROR
+        end
 
         client.update(:index => CONNECTORS_INDEX, :id => connector_id, :body => body)
       end
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -20,13 +20,7 @@ module Core
     class << self
 
       def force_sync(connector_id)
-        body = {
-          :doc => {
-            :scheduling => { :enabled => true },
-            :sync_now => true
-          }
-        }
-        client.update(:index => CONNECTORS_INDEX, :id => connector_id, :body => body)
+        update_connector_fields(connector_id, :scheduling => { :enabled => true }, :sync_now => true)
       end
 
       def create_connector(index_name, service_type)
@@ -63,15 +57,10 @@ module Core
       end
 
       def claim_job(connector_id)
-        body = {
-          :doc => {
-            :sync_now => false,
-            :last_sync_status => Connectors::SyncStatus::IN_PROGRESS,
-            :last_synced => Time.now
-          }
-        }
-
-        client.update(:index => CONNECTORS_INDEX, :id => connector_id, :body => body)
+        update_connector_fields(connector_id,
+                                :sync_now => false,
+                                :last_sync_status => Connectors::SyncStatus::IN_PROGRESS,
+                                :last_synced => Time.now)
 
         body = {
           :connector_id => connector_id,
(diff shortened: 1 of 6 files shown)


lib/core/elastic_connector_actions.rb
```diff
@@ -88,13 +88,19 @@ def claim_job(connector_package_id)
...
             :last_seen => Time.now
           }
         }
 
+        connector_setting = Core::ConnectorSettings.fetch(connector_package_id)
+        if connector_setting.connector_status_allows_sync?
+          connector_instance = Connectors::REGISTRY.connector(service_type, connector_setting.configuration)
+          body[:doc][:status] = connector_instance.source_status[:status] == 'OK' ? Connectors::ConnectorStatus::CONNECTED : Connectors::ConnectorStatus::ERROR
+        end
+
```
artem-shelkovnikov, line 103: ElasticConnectorActions method is not supposed to have any domain logic - it's only a wrapper around calls to Elasticsearch that are needed to support BYOC and BYOEI protocols.

I think this code should be pulled up to the caller of the method or into another class that encapsulates the logic.

lib/core/elastic_connector_actions.rb
```diff
@@ -155,8 +161,8 @@ def fetch_document_ids(index_name)
           hits = response['hits']['hits']
           ids = hits.map { |h| h['_id'] }
           result += ids
-          body[:search_after] = hits.last['sort']
           break if hits.size < page_size
+          body[:search_after] = hits.last['sort']
```
artem-shelkovnikov, line 165: What was the bug here?

lib/core/output_sink/es_sink.rb
```diff
@@ -40,7 +40,7 @@ def delete(doc_id)
     def flush(size: nil)
       flush_size = size || @flush_threshold
 
-      while @queue.any?
+      while @operation_queue.any?
```
artem-shelkovnikov, line 43: My bad, I did not properly fix it during the merge conflict fixes 🤦

artem-shelkovnikov (changes requested):

lib/core/elastic_connector_actions.rb
wangch079: OK, will look deeper tomorrow

lib/core/elastic_connector_actions.rb
wangch079: When hits are empty, `hits.last` is `nil`

artem-shelkovnikov: Ahhhh, good one!

artem-shelkovnikov:
Hi @wangch079, I've moved the small bugfixes into individual PRs to unblock further testing:

https://github.com/elastic/connectors-ruby/pull/200
https://github.com/elastic/connectors-ruby/pull/200

lib/core/elastic_connector_actions.rb
```diff
@@ -247,13 +237,9 @@ def ensure_job_index_exists
...
       end
 
-      def update_connector_field(connector_id, field_name, value)
-        body = {
-          :doc => {
-            field_name => value
-          }
-        }
-        client.update(:index => CONNECTORS_INDEX, :id => connector_id, :body => body)
+      def update_connector_fields(connector_id, doc = {})
```
artem-shelkovnikov, line 240: I'm not 100% sure about the name here - should it just be `update_connector`?

This method can also be reused in other methods throughout the file, for example in `update_connector_status` and such, while it can also be outside of scope of this PR

artem-shelkovnikov (approved):
LGTM! Left a small non-blocking note about method name.

lib/core/elastic_connector_actions.rb
wangch079: > This method can also be reused in other methods throughout the file, for example in `update_connector_status` and such, while it can also be outside of scope of this PR

Good point, I will make the change. It is also reused now by methods like `update_connector_configuration`. The previous method is called `update_connector_field`, which only allow updating one field in one request.

> I'm not 100% sure about the name here - should it just be `update_connector`?

Because this method is heavily used by other methods in `ElasticConnectorActions`, I think it makes more sense to keep it `update_connector_fields`

seanstory:
@wangch079 this was merged to `main` but not backported to 8.4. But the label is v8.4.0. Does this need a backport?

wangch079:
> @wangch079 this was merged to `main` but not backported to 8.4. But the label is v8.4.0. Does this need a backport?

Will backport it along with another PR https://github.com/elastic/connectors-ruby/pull/216.

artem-shelkovnikov:
@wangch079 I've tested it locally and subjectively it made my user experience a bit worse.

Before when I start the connector I immediately get the update of connector configuration and can go to the interface to fill in the configurable fields or set up scheduling.
If I immediately enable scheduling, with next wake-up I can start a sync immediately.

After the change the connector updates the connector configuration not immediately, but after the timer is triggered - in 30 seconds, which in turn makes me wait for 30 seconds to be able to set up configurable fields.

I think we can do a follow-up PR to trigger heartbeat of the connector immediately on start, what do you think?

wangch079:
@artem-shelkovnikov , initially I want to move the configuration initialization to heartbeat because I don't think it should be tied to job execution. I just realised that it runs in every loop because it was coded before the `should_sync?` check. But still, I think it has nothing to do with the sync job.

Both the Heartbeat timer and the polling job loop run every 60 seconds, so I don't really think it make it worse by moving it to heartbeat. But I agree your bad UX is due to `run_now` set to `false`. I can draft the PR, just one line change.

PR #203, opened 2022-07-22
Clean up stubs/connectors.rb and all usages

artem-shelkovnikov:
Not used any more, and seems there's no need for it any more.


lib/connectors/base/extractor.rb
```diff
@@ -12,8 +12,8 @@
 require 'date'
 
 require 'utility'
+require 'stubs/app_config'
 require 'stubs/connectors/stats' unless defined?(Rails)
-require 'stubs/connectors' unless defined?(Rails)
```
wangch079, line 16: When will `defined?(Rails)` be `true`?

wangch079 (commented):
What's the purpose of the `lib/stubs` dir? Can we remove it entirely?

lib/connectors/base/extractor.rb
artem-shelkovnikov: It never will be, it was done so that this file could be used in Rails apps as a part of a gem. At this point it's safe to remove this everywhere.

artem-shelkovnikov:
> What's the purpose of the lib/stubs dir? Can we remove it entirely?

It was done when we had to share a gem between `connectors` repository and `ent-search` repository. At this point there's no need in stubs, but there are some hard dependencies on them in the project that we need to resolve before removing code.

wangch079 (commented):
>It was done when we had to share a gem between `connectors` repository and `ent-search` repository. 

It doesn't feel right that the gem has some logic catering to the user of the gem.

>At this point there's no need in stubs, but there are some hard dependencies on them in the project that we need to resolve before removing code.

Can we aim to remove the whole `lib/stubs` dir?

artem-shelkovnikov:
> It doesn't feel right that the gem has some logic catering to the user of the gem.

It was done purely for Connectors 2.0 code sharing, now we don't need it and I'm all for deleting it!

> Can we aim to remove the whole lib/stubs dir?

Yes, absolutely. Do you think it's better to do it in one PR or in multiple smaller ones? Stuff's quite tangled there.

wangch079:
> Do you think it's better to do it in one PR or in multiple smaller ones? Stuff's quite tangled there.

I'd prefer to delete the whole dir in one PR. I actually don't understand the challenge you mentioned below:

>but there are some hard dependencies on them in the project that we need to resolve before removing code.

artem-shelkovnikov:
@wangch079 I've updated PR to delete all files in `stub` directory, all files outside of `stub` directory removed are the files that actually not used at all and include `stub` inside.

Some of these files (like Monitor) and logic from `Base::Extractor` we will need to introduce back and cover with a good number of tests, but for now they are just not used at all any more.

wangch079 (approved):
I just realized we don't have any `extractor` extending `Base::Extractor` until I see you removed it. I think now the decision is that `extractor` is not one of the required abstractions. I kindly feel `connector` and `extractor` somehow serve the same purpose. I'm fine with removing it. We can always iterate on this later.

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.4|[](https://github.com/elastic/connectors-ruby/pull/257)|

### Questions ?

PR #204, opened 2022-07-22
Remove AppConfig.connectors_logger method and everything connected to it

artem-shelkovnikov:
Does not seem to be needed any more - inlined usage of `AppConfig.connectors_logger` to the only place that used it.

### DIFF
--- a/lib/stubs/app_config.rb
+++ b/lib/stubs/app_config.rb
@@ -4,8 +4,6 @@
 # you may not use this file except in compliance with the Elastic License.
 #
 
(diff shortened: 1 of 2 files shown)


lib/stubs/app_config.rb
```diff
@@ -8,10 +8,6 @@
```
wangch079, line 8:
```
require 'logger'
```

This can be removed as well.

wangch079 (approved):
LGTM expect one small change

PR #205, opened 2022-07-22
Clean up some unused logic from Utility:: namespace

artem-shelkovnikov:
Just a couple small cleanups of `Utility::*` namespace:

- Remove constants that are not needed any more from Utility::Constants
- Remove `lib/utility/extraction_utils.rb` and specs for it
- Remove `lib/utility/job_status.rb`


wangch079 (approved):
LGTM

PR #209, opened 2022-07-25
Change logging a bit to log better and more versatile.

artem-shelkovnikov:
Various logging improvements:

1. ElasticConnectorActions do not do any `info` logs any more - only debug. Callers of ElasticConnectorActions now do logging instead in a way they need to.
2. Moved some messages from "info" to "debug"
3. Added summary logging lines in the end of sync that specify number of added/deleted docs.
4. Added a dot in the end of every log line I've found

### DIFF
--- a/lib/app/console_app.rb
+++ b/lib/app/console_app.rb
@@ -49,7 +49,7 @@ module App
       return unless connector_registered?
       puts 'Initiating synchronization NOW...'
       Core::ElasticConnectorActions.force_sync(connector_id)
-      puts "Successfully scheduled an on-demand sync for connector #{connector_id}"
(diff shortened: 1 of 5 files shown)


lib/app/console_app.rb
```diff
@@ -49,6 +49,7 @@ def start_sync_now
       return unless connector_registered?
       puts 'Initiating synchronization NOW...'
       Core::ElasticConnectorActions.force_sync(connector_id)
+      puts "Successfully scheduled an on-demand sync for connector #{connector_id}"
```
wangch079, line 52: This message is a bit confusing.

```suggestion
      puts "Successfully synced for connector #{connector_id}"
```

wangch079 (approved):

PR #210, opened 2022-07-25
Add human-readable error message when service was unable to fetch connector with specified id

artem-shelkovnikov:
Now if connector service was unable to fetch connector, error message is confusing, for example:
```
/Users/artemshelkovnikov/git_tree/connectors/lib/app/worker.rb:37:in `pre_flight_check': undefined method `index_name' for nil:NilClass (NoMethodError)
        from /Users/artemshelkovnikov/git_tree/connectors/lib/app/worker.rb:22:in `start!'
        from app.rb:21:in `<module:App>'
        from app.rb:15:in `<main>'
make: *** [exec_app] Error 1
```

This PR fixes it by adding code into `ConnectorSettings` that raises an error if Elasticsearch returned `:found => false`. 

Example error message:

```
/Users/artemshelkovnikov/git_tree/connectors/lib/core/connector_settings.rb:20:in `fetch': Connector with id=rR52NYIBZjXva38MAONW~ was not found. (Core::ConnectorSettings::ConnectorNotFoundError)
        from /Users/artemshelkovnikov/git_tree/connectors/lib/app/worker.rb:37:in `pre_flight_check'
        from /Users/artemshelkovnikov/git_tree/connectors/lib/app/worker.rb:22:in `start!'
        from app.rb:21:in `<module:App>'
        from app.rb:15:in `<main>'
make: *** [exec_app] Error 1
```


lib/core/connector_settings.rb
```diff
@@ -13,13 +13,15 @@
...
+
     def self.fetch(connector_id)
-      es_response = ElasticConnectorActions.load_connector_settings(connector_id)
-      if es_response['found'] == false
-        Utility::Logger.debug("Connector settings not found for connector_id: #{connector_id}")
-        return nil
-      end
-      new(es_response.with_indifferent_access)
+      es_response = ElasticConnectorActions.get_connector(connector_id)
+        .with_indifferent_access
```
wangch079, line 21: duplicate `with_indifferent_access`

lib/app/console_app.rb
```diff
@@ -129,13 +129,13 @@ def connector_registered?(warn_if_not: true)
...
 
     def create_connector(index_name, force: false)
-      connector_settings = Core::ConnectorSettings.fetch(connector_id)
-
-      if connector_settings.nil? || force
-        created_id = Core::ElasticConnectorActions.create_connector(index_name, App::Config['service_type'])
-        connector_settings = Core::ConnectorSettings.fetch(created_id)
+      id = connector_id
+      if Core::ElasticConnectorActions.get_connector(connector_id)[:found] == false || force
+        id = Core::ElasticConnectorActions.create_connector(index_name, App::Config['service_type'])
```
wangch079, line 134: If the connector exists and `force = true`, it will try to create the connector in `.elastic-connectors`. It won't raise errors but we should have the restriction at the application level to avoid duplicate index names.

It will also bring problems that two connectors write data to the same content index.

lib/app/console_app.rb
```diff
@@ -129,13 +129,13 @@ def connector_registered?(warn_if_not: true)
...
       end
 
+      connector_settings = Core::ConnectorSettings.fetch(id)
```
wangch079, line 137:
```suggestion
      connector_settings = begin
        Core::ConnectorSettings.fetch(id)
      rescue Core::ConnectorSettings::ConnectorNotFoundError
        Core::ElasticConnectorActions.create_connector(index_name, App::Config['service_type'])
      end
```
With the above suggestion, `Core::ElasticConnectorActions.create_connector` should return connector setting object, not just `id`.

lib/core/connector_settings.rb
artem-shelkovnikov: I deleted it from line 22, so it should not be duplicate any more

lib/app/console_app.rb
artem-shelkovnikov: That is true, however this PR just adds a human-readable error message, not changing the current flow.

This comment should be addressed in a different PR cc @mchernyavskaya

lib/app/console_app.rb
artem-shelkovnikov: +1 to that, but out of scope of this PR as this PR only adds human-readable messages, not changing the way CLI works.

lib/app/console_app.rb
```diff
@@ -129,13 +129,13 @@ def connector_registered?(warn_if_not: true)
     end
 
     def create_connector(index_name, force: false)
-      connector_settings = Core::ConnectorSettings.fetch(connector_id)
-
-      if connector_settings.nil? || force
-        created_id = Core::ElasticConnectorActions.create_connector(index_name, App::Config['service_type'])
-        connector_settings = Core::ConnectorSettings.fetch(created_id)
+      id = connector_id
+      if Core::ElasticConnectorActions.get_connector(connector_id)[:found] == false || force
```
tarekziade, line 133: `== false` is not something you want to keep. never use a bool after `==`

you want:

```
unless Core::ElasticConnectorActions.get_connector(connector_id)[:found]  || force
```
(untested)

wangch079 (approved):
Looks good for the scope of this PR.

lib/app/console_app.rb
tarekziade: Right, I should have changed `||` to `&&`

```
if !Core::ElasticConnectorActions.get_connector(connector_id)[:found] || force
```
or 

```
unless Core::ElasticConnectorActions.get_connector(connector_id)[:found] && force
```

but you get the idea

PR #211, opened 2022-07-25
Use the Quartz Cron

tarekziade:
Converts the Quartz cron syntax provided by Kibana into a Crontab one for now.
We want to fully support the Quartz format later on but that will involve more work


artem-shelkovnikov:
@tarekziade can you update PR description to include some information about the change?

artem-shelkovnikov (approved):
Let's follow up on this ASAP!

PR #214, opened 2022-07-26
Start tracking :deleted_document_count

artem-shelkovnikov:
As of now when a sync job runs we define `@status[:deleted_document_count] = 0` but never update it. 

This PR fixes it, allowing to track number of documents deleted by the sync job with every run

### DIFF
--- a/lib/core/sync_job_runner.rb
+++ b/lib/core/sync_job_runner.rb
@@ -83,7 +83,11 @@ module Core
 
       Utility::Logger.info("Deleting #{ids_to_delete.size} documents from index #{@connector_settings.index_name}.")
 
-      @sink.delete_multiple(ids_to_delete)
+      ids_to_delete.each do |id|
+        @sink.delete(id)
+        @status[:deleted_document_count] += 1
+      end
+
       @sink.flush
     rescue StandardError => e
       @status[:error] = e.message


lib/core/sync_job_runner.rb
```diff
@@ -83,7 +83,11 @@ def do_sync!
 
       Utility::Logger.info("Deleting #{ids_to_delete.size} documents from index #{@connector_settings.index_name}.")
 
-      @sink.delete_multiple(ids_to_delete)
+      ids_to_delete.each do |id|
+        @sink.delete(id)
+        @status[:deleted_document_count] += 1
+      end
```
wangch079, line 89: Why not: 
```suggestion
      @sink.delete_multiple(ids_to_delete)
      @status[:deleted_document_count] += ids_to_delete.size
```

artem-shelkovnikov: I did it this way because it allows for more precise tracking of number of deleted docs.

The way it's done now `@status[:deleted_document_count]` will be more accurate if something happens during deletion - for example, Elasticsearch instance crashes, or there's a bug in deletion code that will raise an error.

In the proposed snippet the deletion count is updated only after everything was deleted, which may not be reached if something goes wrong.

Also this way we can later update these counts during the job run, not in the end of job run.

wangch079 (approved):
>The way it's done now `@status[:deleted_document_count]` will be more accurate if something happens during deletion

Actually, this is not 100% correct. We will only know the real number when `flush` is called.  But not a big concern for me.

serenachou:
Thanks for the flag. Cc @aznick on the data inaccuracy possibility

PR #216, opened 2022-07-27
Remove analyzer settings for content index

wangch079:
Remove the analyzer settings for the content index from the connector service, which should be set by Kibana when the index is created.

#### Pre-Review Checklist

### DIFF
--- a/config/connectors.yml.example
+++ b/config/connectors.yml.example
@@ -13,10 +13,3 @@ service_type: "CHANGEME"
 log_level: info
 
 idle_timeout: 60
-
-# ICU Analysis Plugin is used (false by default)
-# turn this on if the ICU plugin - International Components for Unicode - is installed in your Elasticsearch instance
-# see https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-icu.html
-use_analysis_icu: false
-# language code to use for the content indices
-content_language_code: en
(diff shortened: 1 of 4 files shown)


wangch079 (commented):
I intentionally keep [text_analysis_settings.rb](https://github.com/elastic/connectors-ruby/blob/main/lib/utility/elasticsearch/index/text_analysis_settings.rb) and [language_data.yml](https://github.com/elastic/connectors-ruby/blob/main/lib/utility/elasticsearch/index/language_data.yml).

CLI allows operators to create connectors (but it won't create the content index for now), and those files are needed when the backend needs to create the content index.

artem-shelkovnikov (approved):
LGTM, minor concern - it can break the creation of connectors via CLI and it's not yet possible to test it without https://github.com/elastic/connectors-ruby/pull/218

wangch079:
> LGTM, minor concern - it can break the creation of connectors via CLI and it's not yet possible to test it without #218

For now, `the creation of connectors via CLI` doesn't create the content index. That's why I want to keep the 2 files not in use for now, as mentioned in https://github.com/elastic/connectors-ruby/pull/216#pullrequestreview-1052278795

seanstory:
@wangch079 this was merged to `main` but not backported to `8.4`. But the label is v8.4.0. Does this need a backport?

vidok:
@wangch079 is there any GitHub issue or other references for this change?

wangch079:
> @wangch079 is there any GitHub issue or other references for this change?

No. I just linked the slack discussion in the description of the PR.

PR #223, opened 2022-07-28
Add analyzer when creating content index

wangch079:
A follow-up PR on https://github.com/elastic/connectors-ruby/pull/216.

The analyzer should be added when the content index is created from the CLI.

#### Pre-Review Checklist

### DIFF
--- a/config/connectors.yml.example
+++ b/config/connectors.yml.example
@@ -13,10 +13,3 @@ service_type: "CHANGEME"
 log_level: info
 
 idle_timeout: 60
-
-# ICU Analysis Plugin is used (false by default)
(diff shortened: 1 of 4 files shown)


artem-shelkovnikov:
I'm really not sure about exposing settings that are used only by CLI in configuration file (`use_analysis_icu` and `content_language_code`).

@wangch079  what do you think about prompting for these settings when attempting to create a connector via CLI? (while providing default values, so that you could just hit enter two times if you're okay with defaults)

wangch079:
>what do you think about prompting for these settings when attempting to create a connector via CLI? (while providing default values, so that you could just hit enter two times if you're okay with defaults)

Great idea. Add it to my to-do list next week (need to focus on QA now)

artem-shelkovnikov (approved):

PR #231, opened 2022-08-01
add default crawler mappings

seanstory:
## part of https://github.com/elastic/enterprise-search-team/issues/2391

our default mappings were not creating all the expected schema fields for crawler indices, which conflicted with [the default crawler deduplication fields](https://github.com/elastic/ent-search/blob/main/shared_togo/lib/shared_togo/config_parser/schemas/connector.rb#L4-L11)

#### Pre-Review Checklist

## Related Pull Requests

- https://github.com/elastic/ent-search/pull/6729

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,13 +46,13 @@ tag:
 
 build_gem:
 	mkdir -p .gems
-	gem build connectors_utility.gemspec
+	bundle _$(shell cat .bundler-version)_ exec gem build connectors_utility.gemspec
 	rm -f .gems/*
 	mv *.gem .gems/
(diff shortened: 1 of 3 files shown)


vidok (approved):
Great job! 🎉 

Now it's clear for me that Connector and Crawler mapping definitions should be defined in "base" connector and "base" crawler accordingly. 

so `Utility::Elasticsearch::Index::Mappings` is kind of a base and Connector and Crawler can extend the mapping configuration. 

I don't want to block this PR so let's put it into our tech debt.

PR #232, opened 2022-08-02
Disable warnings emited by the Elasticsearch lib

tarekziade:
## Closes https://github.com/elastic/enterprise-search-team/issues/2423

Intercept the 401 and disable `warn` calls by default

#### Pre-Review Checklist


lib/app/worker.rb
```diff
@@ -19,9 +19,26 @@ module Worker
 
     class << self
       def start!
+        disable_warnings = if App::Config.has_key?(:disable_warnings)
+                             App::Config[:disable_warnings]
```
wangch079, line 23: Should add to [connectors-ruby/config/connectors.yml.example](https://github.com/elastic/connectors-ruby/blob/main/config/connectors.yml.example)

lib/utility/es_client.rb
```diff
@@ -34,6 +35,9 @@ def connection_configs
       else
         raise 'Either elasticsearch.cloud_id or elasticsearch.hosts should be configured.'
       end
+      configs[:log] = es_config[:log] || false
+      configs[:trace] = es_config[:trace] || false
+      configs[:logger] = es_config[:logger] || ::Logger.new(IO::NULL)
```
wangch079, line 40: Should add to [connectors-ruby/config/connectors.yml.example](https://github.com/elastic/connectors-ruby/blob/main/config/connectors.yml.example)

Why not default to `Utility::Logger.logger`?

lib/app/worker.rb
```diff
@@ -19,9 +19,26 @@ module Worker
 
     class << self
       def start!
+        disable_warnings = if App::Config.has_key?(:disable_warnings)
```
wangch079, line 22: Maybe considering moving this up one level to `app.rb` and `console_app.rb`?

lib/utility/es_client.rb
tarekziade: Because we want to silence the Elasticsearch lib logger

tarekziade: oh but the way, logger is an object, so we won't surface this option to the config. here the intent is just to pass a null logger

wangch079: Then why not just set `config[:log] = false`, if we are not gonna log it anywhere?

tarekziade: Because as a developer, I want to have the ability to set the Elasticsearch logger to trace and log if I need to debug something.  For `logger` we could set it to `Utility::Logger.logger` I guess *IF* log or trace are set to true.
And if both are false, just use ::Logger.new(IO::NULL)

tarekziade:
TODO fails with 

```
rspec spec --order rand:19242
```

lib/app/console_app.rb
```diff
@@ -236,37 +236,39 @@ def read_configurable_fields
       end
     end
```
artem-shelkovnikov, line 238: This file still has the initialization of logger and TZ on top:

https://github.com/elastic/connectors-ruby/blob/ae4750a59d25889fb20153aed792a75038ea93a1/lib/app/console_app.rb#L21-23

Should we delete it here too?

lib/utility/logger.rb
```diff
@@ -40,4 +40,24 @@ def new_line
       end
     end
   end
+
+  def self.with_logging(config, &block)
+    # Set UTC as the timezone
+    ENV['TZ'] = 'UTC'
```
artem-shelkovnikov, line 46: ENV['TZ'] is not logger-specific setting, I think it adds some confusion setting it here specifically, though not sure yet about the best way to share the code, can improve later

artem-shelkovnikov (approved):

lib/utility/logger.rb
tarekziade: >  not sure yet about the best way to share the code, can improve later

maybe we could rename `with_logging` to `set_execution_environment` ?

artem-shelkovnikov: Yep, let's move it to a different class, something like `App::Environment` and do it there!

artem-shelkovnikov (approved):

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.4|[](https://github.com/elastic/connectors-ruby/pull/246)|

### Questions ?

PR #235, opened 2022-08-03
Extract logic for scheduling to a separate class

artem-shelkovnikov:
A refactoring to move away the logic about the triggering of the sync to a different place.

This PR creates a `lib/core/scheduler.rb` file that got all the logic related to scheduling moved from the `lib/core/sync_job_runner.rb` file.

This will allow for simpler testing + easier feature implementation for scheduling (specifically, for the logic to notify user about the next time the job will run).

### DIFF
--- a/lib/core/scheduler.rb
+++ b/lib/core/scheduler.rb
@@ -1,0 +1,97 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+require 'time'
+require 'fugit'
+require 'core/connector_settings'
+require 'utility/cron'
+require 'utility/logger'
+require 'utility/exception_tracking'
+
+module Core
+  class Scheduler
+    def initialize(connector_id, poll_interval)
+      @connector_id = connector_id
+      @poll_interval = poll_interval
+    end
+    
+    def when_triggered
+      loop do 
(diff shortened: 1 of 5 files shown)


lib/core/scheduler.rb
```diff
@@ -0,0 +1,97 @@
...
+          sleep(@poll_interval)
+        end
+      end
+    end
+
+    private
+
+    def sync_triggered?(connector_settings)
+      unless connector_settings.connector_status_allows_sync?
+        Utility::Logger.info("Connector #{connector_settings.id} is in status \"#{connector_settings.connector_status}\" and won't sync yet. Connector needs to be in one of the following statuses: #{Connectors::ConnectorStatus::STATUSES_ALLOWING_SYNC} to run.")
```
tarekziade, line 44: we starting to have a lot of messages here. I would put them in constants at the top of the file to keep the code easy to read

tarekziade (approved):
Neat! Just a nit.
also, whenever we have time we can translate this https://github.com/elastic/connectors-python/blob/main/connectors/quartz.py into ruby and use plain Quartz

lib/core/scheduler.rb
artem-shelkovnikov: Good call, I'll make a follow-up PR, though it'll change the usage of logger a bit to style of `printf "Connector %s is in status '%s'", connector_settings.id, connector_settings.status` - which is pretty okay too.

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.4|[](https://github.com/elastic/connectors-ruby/pull/256)|

### Questions ?

PR #241, opened 2022-08-08
[8.4] Start heartbeat immediately (#227)

wangch079:
Backports the following commits to 8.4:
 - Start heartbeat immediately (#227)


artem-shelkovnikov:
Hmmm, actually not needed any more, see https://github.com/elastic/connectors-ruby/pull/247/files#diff-d39431d5c5b01c49b79b2c8fe5a0ab42c4cdd9b3cdf760258eadc8ef4b2f9868R24

PR #243, opened 2022-08-08
Add 100% spec coverage for Core::SyncJobRunner

artem-shelkovnikov:
This PR adds tests for Core::SyncJobRunner to reach full condition/code coverage.

One small change to the logic: now when connector is unable to claim the job, it will not even attempt to run the sync any more.

### DIFF
--- a/lib/core/sync_job_runner.rb
+++ b/lib/core/sync_job_runner.rb
@@ -45,35 +45,40 @@ module Core
 
       job_id = ElasticConnectorActions.claim_job(@connector_settings.id)
 
-      Utility::Logger.debug("Successfully claimed job for connector #{@connector_settings.id}.")
-
-      incoming_ids = []
-      existing_ids = ElasticConnectorActions.fetch_document_ids(@connector_settings.index_name)
-
-      Utility::Logger.debug("#{existing_ids.size} documents are present in index #{@connector_settings.index_name}.")
-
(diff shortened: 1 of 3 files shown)


seanstory:
Was this intended to backport to 8.4?

artem-shelkovnikov:
@seanstory not really, should it be?

seanstory:
You have more context than I do, so I'll trust you to make the call. But I'd prefer to have tests in 8.4's branch before we release 8.4, especially as some of these changes (like this one: https://github.com/elastic/connectors-ruby/pull/243/files#diff-b7553380580871ff031b526852d48633f01559e06eb918c2c6785a76261bdfc6R14) look to be fixing subtle bugs.

It also seems like it would be a good idea to have tests backported before we have to start any development on an 8.4.1. Would be good to be protected against regressions as we make any necessary bugfixes.

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.4|[](https://github.com/elastic/connectors-ruby/pull/259)|

### Questions ?

artem-shelkovnikov:
@seanstory I went ahead and backported everything.

>  before we have to start any development on an 8.4.1.

Are we planning to though? I don't honestly expect that connectors are used as of 8.4, so we won't need to fix anything there - we're not GA yet.

PR #244, opened 2022-08-08
[8.4] Rename stub connector to example connector (#237)

wangch079:
Backports the following commits to 8.4:
 - Rename stub connector to example connector (#237)


mchernyavskaya (approved):
LGTM

PR #248, opened 2022-08-08
GitLab connector does not sync the files and folders properly #2461

mchernyavskaya:
## Closes https://github.com/elastic/enterprise-search-team/issues/2461

Cleaning up leftover code for GitLab files/folders sync that might have been confusing.

There is just one code block that needed to be removed - less than I thought. And this block was already unused. So this change should be pretty safe.


artem-shelkovnikov (approved):
LGTM!

tarekziade (approved):

PR #252, opened 2022-08-09
Add 100% test coverage for Core::ElasticCoreActions

artem-shelkovnikov:
Another PR that adds a spec for our core modules - this time for `Core::ElasticCoreActions`.

### DIFF
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -9,10 +9,7 @@
 require 'active_support/core_ext/hash'
 require 'connectors/connector_status'
 require 'connectors/sync_status'
-require 'utility/elasticsearch/index/mappings'
-require 'utility/elasticsearch/index/text_analysis_settings'
-require 'utility/es_client'
-require 'utility/logger'
+require 'utility'
 
 module Core
   class ElasticConnectorActions
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -7,10 +7,12 @@
 # frozen_string_literal: true
 #
 require 'active_support/core_ext/hash'
-require 'utility/elasticsearch/index/mappings'
-require 'utility/elasticsearch/index/text_analysis_settings'
 require 'connectors/connector_status'
 require 'connectors/sync_status'
+require 'utility/elasticsearch/index/mappings'
+require 'utility/elasticsearch/index/text_analysis_settings'
+require 'utility/es_client'
+require 'utility/logger'
 
 module Core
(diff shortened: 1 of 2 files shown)


spec/core/elastic_connector_actions_spec.rb
```diff
@@ -0,0 +1,558 @@
+require 'core/elastic_connector_actions'
+require 'connectors/connector_status'
+require 'connectors/sync_status'
+require 'utility/es_client'
+
+describe Core::ElasticConnectorActions do
+  let(:connector_id) { 'one-two-three' }
+  let(:es_client) { double }
```
seanstory, line 8: Not for this PR, but has there been any discussion about getting a local Elasticsearch available on the CI instances for connectors? That would make these more "integration" tests than unit tests, but also might significantly lower the mocking burden.

seanstory (approved):
My eyes glazed over after a few hundred lines, but kudos for writing so many tests in one go!

lib/core/elastic_connector_actions.rb
```diff
@@ -7,10 +7,12 @@
...
 #
 require 'active_support/core_ext/hash'
-require 'utility/elasticsearch/index/mappings'
-require 'utility/elasticsearch/index/text_analysis_settings'
 require 'connectors/connector_status'
 require 'connectors/sync_status'
+require 'utility/elasticsearch/index/mappings'
+require 'utility/elasticsearch/index/text_analysis_settings'
+require 'utility/es_client'
+require 'utility/logger'
```
wangch079, line 15: Why not
```suggestion
require 'utility'
```

I also observed such cases before, even though there's no `require` for some files, it still doesn't throw any error.

wangch079 (approved):
Nice work

spec/core/elastic_connector_actions_spec.rb
artem-shelkovnikov: We do have integration test that is ran nightly - you can run it with `make ftest` - it's a functional test though, not an integration one.

We did not have a discussion about that and it's something nice to talk about indeed.

IMO there's not much mocking burden as long as the class doesn't do too much - and mocking helps see class dependencies on 3rd-party components when looking at the test, which I think is quite valuable too. 

All in all, I think the strategy I personally like the most is - as many as possible unit tests + a small number of integration tests that test big chunks of the system + a small number of functional tests that test high-level scenarios.

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.4|[](https://github.com/elastic/connectors-ruby/pull/262)|

### Questions ?

PR #254, opened 2022-08-09
Add tests for Core::Output_Sink::* + refactor sinks a bit

artem-shelkovnikov:
This PR adds tests for files under `lib/core/output_sink/` directory.

Minor changes to sinks:

- Standartize `not implemented` message for BaseSink
- Use `puts` instead of `Logger.info` in ConsoleSink - it seems like usage of logger here does not make sense, cause sink is expected to always output by design
- Move `attr_accessor :index_name` for EsSink to `private` region

### DIFF
--- a/lib/core/output_sink/base_sink.rb
+++ b/lib/core/output_sink/base_sink.rb
@@ -18,7 +18,7 @@ module Core
       end
 
       def delete(_id)
-        raise 'not_implemented'
+        raise 'not implemented'
       end
 
       def delete_multiple(_ids)
--- a/lib/core/output_sink/console_sink.rb
+++ b/lib/core/output_sink/console_sink.rb
@@ -13,37 +13,38 @@ module Core::OutputSink
   class ConsoleSink < Core::OutputSink::BaseSink
     def ingest(document)
       print_header 'Got a single document:'
-      Utility::Logger.info document
+      puts document
     end
 
     def flush(size: nil)
       print_header 'Flushing'
+      puts "Flush size: #{size}"
     end
 
     def ingest_multiple(documents)
       print_header 'Got multiple documents:'
-      Utility::Logger.info documents
+      puts documents
     end
 
     def delete(id)
       print_header "Deleting single id: #{id}"
-      Utility::Logger.info id
+      puts id
     end
 
     def delete_multiple(ids)
       print_header "Deleting several ids: #{ids}"
-      Utility::Logger.info ids
+      puts ids
     end
 
     private
 
     def print_delim
-      Utility::Logger.info '----------------------------------------------------'
+      puts '----------------------------------------------------'
     end
 
     def print_header(header)
       print_delim
-      Utility::Logger.info header
+      puts header
       print_delim
     end
   end
--- a/lib/core/output_sink/es_sink.rb
+++ b/lib/core/output_sink/es_sink.rb
@@ -13,8 +13,6 @@ require 'utility/logger'
 
 module Core::OutputSink
   class EsSink < Core::OutputSink::BaseSink
-    attr_accessor :index_name
-
     def initialize(index_name, flush_threshold = 50)
       super()
       @client = Utility::EsClient.new
@@ -57,6 +55,8 @@ module Core::OutputSink
     end
 
     private
+
+    attr_accessor :index_name
 
     def send_data(ops)
       return if ops.empty?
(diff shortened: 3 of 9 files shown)


lib/core/output_sink/console_sink.rb
```diff
@@ -13,37 +13,38 @@ module Core::OutputSink
   class ConsoleSink < Core::OutputSink::BaseSink
     def ingest(document)
       print_header 'Got a single document:'
-      Utility::Logger.info document
+      puts document
```
seanstory, line 16: my gut feel is that it's better to use a logger than `puts`. My guess is that this change makes it easier to test, though?

what if you gave this class its own `print` function, so that the tests could say something like:
```
expect(subject).to receive(:print).with(includes(expected_value))
```
?

That way the output could still get the nice configs that come from a real logger (and the ability to turn the output on/off with log level, change highlighting, etc), but still easy testability?

lib/core/output_sink/base_sink.rb
```diff
@@ -18,7 +18,7 @@ def ingest_multiple(_documents)
       end
 
       def delete(_id)
-        raise 'not_implemented'
+        raise 'not implemented'
```
seanstory, line 21: Should this just use a [NotImplementedError](https://ruby-doc.org/core-2.5.0/NotImplementedError.html)?

lib/core/output_sink/es_sink.rb
```diff
@@ -58,6 +56,8 @@ def delete_multiple(ids)
 
     private
 
+    attr_accessor :index_name
```
seanstory, line 59: Why not keep it public? Since we're developing this for public consumption, I am inclined to think we should err towards public access, but that's a loosely held intuition, and I'm more just curious about your motivation.

lib/core/output_sink/base_sink.rb
wangch079: TIL, but this is worth a separate PR as we have similar cases in other places.

lib/core/output_sink/es_sink.rb
wangch079: I think we can make it public but `attr_reader`

lib/core/output_sink/console_sink.rb
wangch079: This is a great idea. you can still have `puts` implemented in `print`, and it's easy to change later in one place instead of search and replace in the whole file.

> my gut feel is that it's better to use a logger than `puts`. My guess is that this change makes it easier to test, though?

I guess the main reason to use `puts` instead of `logger` for console (not only this file, but also CLI), is that logger logs a message like

> INFO -- : a message

And we really want is simply `a messsage`

spec/core/output_sink/base_sink_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    let(:sink) { Core::OutputSink::ConsoleSink.new }
+
+    it_behaves_like 'implements all sink methods'
+  end
+
+  context 'Core::OutputSink::EsSink' do
+    let(:sink) { Core::OutputSink::EsSink.new('something') }
+
+    it_behaves_like 'implements all sink methods'
+  end
```
wangch079, line 51: Feel it's more appropriate to move these to different spec files.

lib/core/output_sink/console_sink.rb
artem-shelkovnikov: There's not much overhead added when testing the logic

I can indeed change it to `print` and do `puts` inside.

Why I changed it to `puts`: this sink is not used by default, you can use it manually when needed, so when you use it you use it with intent to see the documents that would have been ingested into the Elasticsearch instance. Since there is this intent, I think it makes sense to always output the information, not based on log level.

lib/core/output_sink/es_sink.rb
artem-shelkovnikov: I don't actually see any place using the `index_name` outside of the `EsSink` itself, so it's just a cleanup. Also since there's a base class `BaseSink`, I think it makes sense to keep the same public facing interface as the other `BaseSink` implementation - it might not be the case for us if we _need_ to access the `index_name` from outside of sink, but it's already available when you instantiate it - see `initialize` method!

spec/core/output_sink/base_sink_spec.rb
artem-shelkovnikov: Good call, will move them to individual files!

seanstory (approved):

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.4|[](https://github.com/elastic/connectors-ruby/pull/264)|

### Questions ?

PR #275, opened 2022-08-22
[Native Connectors] Provide a host for connectors-ruby that can handle native connectors

mchernyavskaya:
## Backend changes

We need to provide a way to run a host that is able to handle Native connectors.

This host is different from current externally hosted connector. New host should be able to run like Python one does - it listens to all connectors in `.elastic-connector` instance, checks one that are supported by the service (for example, only `native` connectors) and run sync jobs for such connectors in a separate thread.

Need to remember, that this host will be later deployed into Cloud along OR as a part of Enterprise Search application.

## Acceptance criteria

By the end of this epic, a user will be able to ...

- [x] Have a Connector Service host that can handle Native connectors
- [x] Have a Connector Service host that can be hosted in Cloud
- [x] Ingest data from multiple data sources of the same type via native connectors in one Enterprise Search deployment. (i.e. multiple MongoDB databases)
- [x] Ingest data from multiple connectors at the same time, in parallel

#### Pre-Review Checklist


lib/app/dispatcher.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    attr_reader :workers
+
+    def initialize
+      @workers = {}.with_indifferent_access
+      @pool = Concurrent::ThreadPoolExecutor.new(
+        min_threads: 5,
+        max_threads: 10,
+        max_queue: 100,
+        fallback_policy: :caller_runs
```
wangch079, line 29: Any particular reason for using `caller_runs`?

I would suggest `abort`, and the dispatcher catches it and logs it appropriately. We don't want the dispatcher thread to be blocked by a sync job, and it's acceptable a sync job is discarded if the workers are busy.

lib/app/dispatcher.rb
```diff
@@ -0,0 +1,88 @@
...
+      )
+      @is_shutting_down = false
+    end
+
+    def start!
+      if App::Config['mode'] == 'dispatcher'
```
wangch079, line 35: Is this line necessary?

lib/app/dispatcher.rb
```diff
@@ -0,0 +1,88 @@
...
+        loop do
+          connectors = Core::ElasticConnectorActions.native_connectors
+          if connectors.empty?
+            Utility::Logger.info('No native connectors found.')
+            return
+          end
+          Utility::Logger.info("Total workers: #{workers.size}. Checking if all native connectors are running...")
+          connectors.each do |connector|
+            if !workers[connector[:id]].nil?
```
wangch079, line 44: How do you reset the worker?

wangch079 (commented):
I had the discussion with @sphilipse in the [implementation doc](https://docs.google.com/document/d/1zRBk4_3hVI27utStSl0KrR9d2gixIpCO2lR5GZb9uCw/edit?disco=AAAAdFIzchk), but I think GitHub might be a better place to continue the discussion. 

As an operator, how do I 
1. create a native connector? e.g. mongo.
2. create a native connector but with custom connector implementation? e.g. gitlab, but I want to sync files.
3. create a custom connector which doesn't supported by Elastic officially? e.g. wikipedia.

For 1), I expect the operator will choose from a list of available connectors in Kibana, and `is_native` is set to `true` by default.
For 2), I expect the operator will still choose (`gitlab`) from a list of available connectors in Kibana, but toggle off `is_native`, fork our `connectors-ruby`, remove all connectors in dir `connectors` except `gitlab`, make necessary changes and host it externally.
For 3), I expect the operotr will need to key in  the service type, e.g. `wikipedia`, and `is_native` is set to `false` and greyed out. The operator then need to fork our `connectors-ruby`, remove all connectors in dir `connectors`, add `wikipedia`, and host it externally.

For both 2) and 3), the connector service needs to know it runs in non-native mode. Instead of having `dispatcher` mode and `worker` mode, can we just have `native_mode: true | false`?

In such case, even it's not native_mode, we don't have to specify `connector_id` and `service_type` in the config, we just need to make sure the service type in `.elastic-connectors` matches any connector in dir `connectors`. WDYT?

artem-shelkovnikov:
@wangch079 
> For 1), I expect the operator will choose from a list of available connectors in Kibana, and is_active is set to true by default.

What does `is_active` mean in this context? 

In the end I think the new hosting way (the Dispatcher one) should be only used inside Enterprise Search Cloud and Standalone deployments, user should not need to run it at all on their own (is there a case when they would want that?)

So the Dispatcher host always runs only on connectors who are `is_native=true`, and standalone host runs only on connectors who are `is_native=false`.

That means that if user will create a new connector or run an existing connector with modified code, Kibana will always create it with `is_native=false`, so native connectors host would never handle these connectors.

Do you think this problem deserves a sync @mchernyavskaya @wangch079?

sphilipse:
> 2) create a native connector but with custom connector implementation? e.g. gitlab, but I want to sync files.
> 3) create a custom connector which doesn't supported by Elastic officially? e.g. wikipedia.

These two use cases are essentially identical: you're running custom code, outside of Enterprise Search, outside of our managed deployments to index data. Once you modify the code you're no longer running a native connector.

> In such case, even it's not native_mode, we don't have to specify connector_id and service_type in the config, we just need to make sure the service type in .elastic-connectors matches any connector in dir connectors. WDYT?

We do still have to specify connector_id and service_type for all non-native connectors. connector_id is necessary to give users the flexibility to run multiple different versions, and to mix and match different connectors. Service_type is also used for UI feedback/signalling.

I think we're running up against mismatching mental models again. These suggestions make perfect sense when you think of a Connectors Service as a single application that handles as much as possible in one shell. But the protocol we've designed is agnostic about what the application looks like, and is designed to be able to handle more modern microservice-like architectures where you have many different, distributed, small applications each dealing with just one source of data (or even a slice of that data). Rather than having one Connectors Service do everything, you can have 100 small connectors each concerned with just their own index--and many of those may have the same service type. That architecture lends itself much better to user customization and extension, and we should encourage that customization. 

If we design this to work for non-native connectors without a connector_id, you're going to run into problems as soon as users want to scale up and run multiple different connectors for the same service_type at the same time. Without a connector_id, you can't slice up a single source into different bits and pieces. Parallellization and coordination becomes much harder. 

And what's the benefit of removing that connector_id for custom connectors? The user is already running custom connectors, which means they're doing significant manual configuration. Copy-pasting one connector_id value is not a significant inconvenience for them.

wangch079:
> What does `is_active` mean in this context?

It should be `is_native`, I will update my comment. I agree with every statement you posted @artem-shelkovnikov 

> and standalone host runs only on connectors who are `is_native=false`.

Just confirm, this standalone host (or in Maryna's PR, `worker` mode) is used for customers to run customized connectors in self-managed infrastructure.

>you're going to run into problems as soon as users want to scale up and run multiple different connectors for the same service_type at the same time.

connector service without connector ID and service types can be scaled up post 8.5, with DBLock. 

>And what's the benefit of removing that connector_id for custom connectors? 

We can make our connector service code work for both native and non-native connectors. All we need to do is configure `native_mode` in YAML.

>The user is already running custom connectors, which means they're doing significant manual configuration. Copy-pasting one connector_id value is not a significant inconvenience for them.

This is another reason but your statement makes sense.

I'm also fine with connector ID and service type for now. We can always re-iterate, not a big deal.

sphilipse:
> connector service without connector ID and service types can be scaled up post 8.5, with DBLock.
This is a solution for our current code, but we are designing a much more generic HTTP protocol. View this from the perspective of how the protocol can best work generically, rather than how you can get this to work in Enterprise Search.

lib/app/dispatcher.rb
mchernyavskaya: What does it mean to reset the worker?

lib/app/worker.rb
```diff
@@ -13,45 +13,47 @@
...
-          Core::ElasticConnectorActions.ensure_content_index_exists(connector_settings.index_name)
-        rescue Elastic::Transport::Transport::Errors::Unauthorized => e
-          raise "Elasticsearch is not authorizing access #{e}"
-        end
-      end
+    def initialize(connector_id:, service_type:, is_native:)
+      super()
+      @connector_id = connector_id
+      @service_type = service_type
+      @is_native = is_native
```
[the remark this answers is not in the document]
mchernyavskaya: It's not. I'll remove it.

lib/core/elastic_connector_actions.rb
```diff
@@ -36,6 +37,38 @@ def get_connector(connector_id)
...
+          response = client.search(
+            :index => CONNECTORS_INDEX,
+            :ignore => 404,
+            :body => {
+              :size => DEFAULT_PAGE_SIZE,
+              :from => offset,
+              :query => {
+                :term => { :is_native => true }
+              },
+              :sort => ['name']
```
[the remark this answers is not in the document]
mchernyavskaya: I tried it at first, but I was having this exception:

`"Fielddata access on the _id field is disallowed"`

I think we might be missing something in our mappings for this index, but I opted for sorting on something else, rather than blowing out the scope of this task.

lib/app/dispatcher.rb
mchernyavskaya: Yeah I went back and forth on the strategy. In the end, I chose `caller_runs` as something that's like a poor man's backpressure solution, as in, by tying up the main thread it won't let other jobs that can't be processed be submitted. But I think for the initial implementation, we might be better off with the "fail early" strategy, and that would be aborting. I'll change it.

lib/app/dispatcher.rb
wangch079: The `worker` mode is meant for externally hosted connectors running custom code. I don't even know if connector service should take care of that.

lib/app/dispatcher.rb
wangch079: https://github.com/elastic/connectors-ruby/blob/a841b92fd6b0ec8a16da4d6524f780aa379b64a8/lib/app/dispatcher.rb#L43-L44

With the above code, I suppose what you want to avoid is to have 2 jobs running for the same connector. But you must have a way to reset `@workers` when the job completes, so that `@workers[connector[:id]]` is `nil`. Otherwise there won't be any job running for a particular connector after the first job.

I don't think we need this check here. Even though there's no active job running for a connector in this connector service, there might be another one running in another instance of connector service, and we should make the connectors resilient to multi connector services (not a goal for 8.5)

lib/app/dispatcher.rb
mchernyavskaya: I removed it.

lib/app/dispatcher.rb
mchernyavskaya: I don't quite get this.

> But you must have a way to reset `@workers` when the job completes, so that `@workers[connector[:id]]` is nil. Otherwise there won't be any job running for a particular connector after the first job.

To me, it doesn't sound true, because the worker never stops. It keeps polling for jobs until the program is killed or fails with an unrecoverable exception, AFAIK. A job might be over, but the worker will still keep polling for a new one.

The note you made brings up another concern though, and that is, what if the connector/source is removed? Then it should go away, as in it won't be present in the `.elastic-connectors`, and __then__ the dispatcher should nullify that worker (and kill the thread). For that, I don't provision now, and I suppose I should.

lib/app/dispatcher.rb
```diff
@@ -0,0 +1,90 @@
...
+          else
+            worker = App::Worker.new(
+              connector_id: connector[:id],
+              service_type: connector[:service_type],
+              is_native: true
+            )
+            @workers[connector[:id]] = worker
+            @pool.post do
+              Utility::Logger.info("Starting #{connector[:id]} for service type #{connector[:service_type]}... Total workers: #{@workers.count}")
+              worker.start!
```
artem-shelkovnikov, line 54: `worker.start!` starts heartbeat timer, so having 5 workers will start 5 heartbeat timers?

lib/app/dispatcher.rb
```diff
@@ -0,0 +1,92 @@
...
+      )
+      @is_shutting_down = false
+    end
+
+    def start!
+      loop do
+        connectors = Core::ElasticConnectorActions.native_connectors
+        if connectors.empty?
+          Utility::Logger.info('No native connectors found.')
+          return
```
wangch079, line 39: Should not return, but go to next loop, after sleeping.

wangch079:
Hi @mchernyavskaya @artem-shelkovnikov , currently when `worker.starts!` is called, it will run 
https://github.com/elastic/connectors-ruby/blob/f1633e3ce2d6a7f9691f32449061fccd983d703b/lib/app/worker.rb#L54-L57

which will keep calling `GET /.elastic-connectors/_doc/{connector_id}`.  I think the scheduler was designed when one connector service only handles one connector.

Do you think it makes sense if we change the `Worker` to do a one-off job, instead of looping? So the looping only happens at the app level (in this case, `dispatcher`),  where all native connectors are retrieved in one request (not considering paging).

lib/app/console_app.rb
```diff
@@ -8,16 +8,13 @@
 
 $LOAD_PATH << '../'
 
+require 'fugit'
```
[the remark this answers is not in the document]
mchernyavskaya: Removed.

lib/app/dispatcher.rb
mchernyavskaya: Right. Fixing.

wangch079 (commented):
I see the code split into dispatcher mode and worker mode (you also introduced native scheduler and single scheduler). Actually worker mode is just a special case of dispatcher mode (connectors array of size 1 vs connectors array of size N). Could you make it work for both modes?

lib/core/scheduler.rb
```diff
@@ -15,26 +15,41 @@
 
 module Core
   class Scheduler
-    def initialize(connector_id, poll_interval)
-      @connector_id = connector_id
+    attr_reader :is_shutting_down
+
+    def initialize(poll_interval)
       @poll_interval = poll_interval
+      @is_shutting_down = is_shutting_down
```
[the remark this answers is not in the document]
mchernyavskaya: From the dispatcher's `shutdown` method. Also, it's used in tests.

artem-shelkovnikov:
I think the code is moving to a great direction, one minor thing that we've discussed with Maryna on the call is that we can batch health check calls together instead of having a timer for each one, it will help us to go around some problems like stopping healthchecks when connector was deleted or occupying too many threads.

I think we can merge this PR and create issues/checklist for things to address further, so that we can have more targeted discussions on the code, while this code can already be used to unblock the deployment of ruby connectors in cloud.

lib/core/scheduler.rb
```diff
@@ -15,26 +15,41 @@
...
     end
 
     def when_triggered
       loop do
-        connector_settings = Core::ConnectorSettings.fetch(@connector_id)
-
-        if sync_triggered?(connector_settings)
-          yield connector_settings
+        connector_ids.each do |connector_id|
+          connector_settings = Core::ConnectorSettings.fetch(connector_id)
```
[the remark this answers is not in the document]
mchernyavskaya: Done.

lib/app/dispatcher_app.rb
```diff
@@ -0,0 +1,20 @@
...
+$LOAD_PATH << '../'
+
+require 'app/dispatcher'
+require 'app/config'
+require 'utility/environment'
+require 'utility/logger'
+
+module App
+  Utility::Environment.set_execution_environment(App::Config) do
+    App::Dispatcher.run_dispatcher!
```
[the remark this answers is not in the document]
wangch079: I think `set_execution_environment` is still called twice. Here and inside `run_dispatcher!`

lib/core/scheduler.rb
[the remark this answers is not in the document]
wangch079: I don't see `is_shutting_down` is passed in anywhere.

wangch079 (commented):
Left some comments. I agree with what @artem-shelkovnikov said. Let's have small PRs for dedicated purposes. We need to move fast.

lib/app/dispatcher_app.rb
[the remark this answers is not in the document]
mchernyavskaya: Fixed both.

lib/core/scheduler.rb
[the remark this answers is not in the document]
mchernyavskaya: `dispatcher.shutdown` calls `scheduler.shutdown`.
The variable is used in the cycle to check if it should loop again.
The variable is also set in tests - `native_scheduler_spec`.
```
    def shutdown
      Utility::Logger.info("Shutting down dispatcher with pool [#{@pool&.class}] and scheduler [#{@scheduler&.class}]...")
      @is_shutting_down = true
      @scheduler&.shutdown
      @pool.shutdown
      @pool.wait_for_termination(TERMINATION_TIMEOUT)
    end
```

lib/app/dispatcher.rb
```diff
@@ -0,0 +1,83 @@
...
+      @scheduler.when_triggered do |connector_settings|
+        service_type = connector_settings.service_type
+        connector_id = connector_settings.id
+
+        # connector-level checks
+        Core::ElasticConnectorActions.ensure_content_index_exists(connector_settings.index_name)
+        raise "[#{service_type}] is not a supported connector" unless Connectors::REGISTRY.registered?(service_type)
+
+        @pool.post do
+          send_heartbeat(connector_id, service_type)
```
wangch079, line 49: heartbeat should be sent every half hour according to the [protocol](https://github.com/elastic/connectors-ruby/blob/main/docs/CONNECTOR_PROTOCOL.md), but I'm if we tackle it in a separate PR.

lib/core/scheduler.rb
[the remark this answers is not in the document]
wangch079: I think it should be initialized to `false` in method `initialize`:
```
@is_shutting_down = false
```

Otherwise `is_shutting_down` is not defined here?

wangch079 (approved):
LGTM except for one comment.

We can iterate for other changes in separate PRs.

lib/app/dispatcher_app.rb
wangch079, line 18: I just realized `run_dispatcher!` is a private method @mchernyavskaya

PR #276, opened 2022-08-22
Store service_type for connector in Kibana `.elastic-connectors` index.

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/2419

When connector service starts up, now it will attempt to populate `service_type` for the record in Kibana `.elastic-connectors` index so that service type is now displayed in Kibana.

There is still something we can follow up on later, for example when connector tries to run a sync, it should check that it runs against expected service type.

#### Pre-Review Checklist

## Screenshots

Before the fix:
image

After the fix:
image


wangch079 (approved):
LGTM

PR #278, opened 2022-08-23
Make EsClient independent of App::Config

vidok:
## https://github.com/elastic/enterprise-search-team/issues/2543

To be able to use `Utility::EsClient` in ent-search we need to make it independent of `App::Config`. Ent-search doesn't have `connectors.yml` file that is being used in `App::Config`. 

#### Pre-Review Checklist

### DIFF
--- a/lib/core/output_sink/es_sink.rb
+++ b/lib/core/output_sink/es_sink.rb
@@ -15,7 +15,7 @@ module Core::OutputSink
   class EsSink < Core::OutputSink::BaseSink
     def initialize(index_name, flush_threshold = 50)
       super()
-      @client = Utility::EsClient.new
+      @client = Utility::EsClient.new(App::Config[:elasticsearch])
       @index_name = index_name
       @operation_queue = []
       @flush_threshold = flush_threshold
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -220,7 +220,7 @@ module Core
       private
 
       def client
-        @client ||= Utility::EsClient.new
+        @client ||= Utility::EsClient.new(App::Config[:elasticsearch])
       end
     end
   end
--- a/lib/utility/es_client.rb
+++ b/lib/utility/es_client.rb
@@ -8,7 +8,6 @@
 
 require 'logger'
 require 'elasticsearch'
-require 'app/config'
 
 module Utility
   class EsClient < ::Elasticsearch::Client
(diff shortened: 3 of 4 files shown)


artem-shelkovnikov (commented):
Looks great! One nit: @vidok do you mind adding explicit `require app/config` into the files that use this class?

In the end I think we'd like to make only the classes under `App` namespace have access to `App::Config`, making `Core::ElasticConnectorActions` and `Core::OutputSink:*` classes not depend on it either.

vidok:
> Looks great! One nit: @vidok do you mind adding explicit `require app/config` into the files that use this class?
> 
> In the end I think we'd like to make only the classes under `App` namespace have access to `App::Config`, making `Core::ElasticConnectorActions` and `Core::OutputSink:*` classes not depend on it either.

I was thinking about it however, `app/config` is required on the top level of the app [here](https://github.com/elastic/connectors-ruby/blob/main/lib/app/app.rb#L12). I probably have poor knowledge of how you run the code so if think requiring `app/config` is mandatory for every file that uses `App::Config` I will add it.

artem-shelkovnikov:
> I probably have poor knowledge of how you run the code so if think requiring app/config is mandatory to every file that uses App::Config I will add it.

Yeah it works this way, just sometimes running individual tests on individual files provide `missing require` errors, so I prefer to `require` everything the class uses.

I'm also not a ruby expert, so the way it's done now might be the wrong one :)

artem-shelkovnikov (approved):

lib/core/output_sink/es_sink.rb
```diff
@@ -15,7 +15,7 @@ module Core::OutputSink
   class EsSink < Core::OutputSink::BaseSink
     def initialize(index_name, flush_threshold = 50)
       super()
-      @client = Utility::EsClient.new
+      @client = Utility::EsClient.new(App::Config[:elasticsearch])
```
wangch079, line 18: I'm thinking, what is the difference if we change it to `Elasticsearch::Client.new(App::Config[:elasticsearch])`?

We initially have `Utility::EsClient` because we only need one instance of ES client. With multi-connectors support, I think we are going to create a new instance of ES client for each sync job.

wangch079 (approved):
LGTM. Just one comment, but it's not related to the purpose of this PR.

lib/core/output_sink/es_sink.rb
artem-shelkovnikov: There's one feature that we've added on top of `bulk`: 

```
def bulk(arguments = {})
      raise_if_necessary(super(arguments))
    end

    private

    def raise_if_necessary(response)
      if response['errors']
        first_error = response['items'][0]
        raise IndexingFailedError.new("Failed to index documents into Elasticsearch.\nFirst error in response is: #{first_error.to_json}")
      end
      response
    end
```

Removing the client fully will make us do this logic somewhere else, which is not a problem, but something to consider during refactoring.

PR #280, opened 2022-08-26
Add simple POC for asymmetric encryption

artem-shelkovnikov:
Part of work to do asymmetric encryption on the side of the connectors.

Some rough things that were done:

1. Private key is automatically created once and put into `/tmp/temp_cert.pem` - you'll lose it if you restart the app. Also if you want different private key per connector, you can't have it yet, needs to be implemented
2. I'm writing and reading connector configuration from `:encrypted_configuration` field just for testing, when Kibana with Sander's POC is used, this can be changed to just `:configuration`
3. Curve that is used is `prime256v1` - it was a surprise to me, but `25519` is not available by default in Ruby OpenSSL package, there's a separate gem `x25519` - not sure how good it is. Need to discuss it and see if Python properly implements it, maybe we can choose a curve that's available everywhere, as long as it does not compromise security.
4. Encryption is just implemented to test things, it uses `ActiveSupport::MessageEncryptor` which uses unknown to me encryption algorithm and is definitely not a `AES-256-CBC` that we want, but can be easily changed.
5. Class structure is rudimentary - it's a PoC so I did not give a big thought to it.
6. Serialization/deserialization of the public key is arbitrary - I convert it to a number, but in reality it should be a PEM format.

To test that this code works you can add logging statements to `encrypt` and `decrypt` methods, and update connector record in `.elastic-connectors` to have the following object in `_source`:

```json
"encryption": {
  "client_public_key": "56824295195220096543251332234389801363768895710006339571233598387083553451534797619870711079289989347057382078386499248727730849166791941719712819299360377"
}
```

This way you'll see that connector is able to encrypt and decrypt data.

### DIFF
--- a/lib/core/connector_settings.rb
+++ b/lib/core/connector_settings.rb
@@ -11,6 +11,7 @@ require 'active_support/core_ext/object/blank'
 require 'connectors/connector_status'
 require 'time'
 require 'utility/logger'
+require 'utility/encryption_helper'
 
 module Core
   class ConnectorSettings
@@ -51,7 +52,7 @@ module Core
     end
 
     def configuration
-      self[:configuration]
+      Utility::EncryptionHelper.decrypt(self[:encryption], self[:encrypted_configuration])
     end
 
     def scheduling_settings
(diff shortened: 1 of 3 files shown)


artem-shelkovnikov:
Closing as we won't do it now and will address the problem in 8.6 differently

PR #282, opened 2022-08-30
[Partial] Connectors recent syncs view #2487 - affected counts on backend

mchernyavskaya:
## Related to https://github.com/elastic/enterprise-search-team/issues/2487

We're missing some information for the **Recent Syncs** view; namely, the affected counts. This PR adds indexed and deleted counts; together, they are what the job changes.

NOTE: it's not necessarily the same as current document count at the time. Since we assume that for 8.5, we always do the full sync, the job deletes all the old documents and creates the new. So in the end, the new documents count = indexed count, and the old documents count = deleted count, because all the old documents are deleted.

#### Pre-Review Checklist
- [x] Added the changes to the protocol

### DIFF
--- a/docs/CONNECTOR_PROTOCOL.md
+++ b/docs/CONNECTOR_PROTOCOL.md
@@ -96,6 +96,8 @@ This is our main communication index, used to communicate the connector's config
                            regularly (UTC)
   last_sync_error: string;   -> Optional last sync error message
   last_synced: date;    -> Date/time of last completed sync (UTC)
+  last_indexed_count: number;    -> How many documents were inserted into the index
+  last_deleted_count: number;    -> How many documents were deleted from the index
   last_sync_status: string;  -> last sync Enum, see below
   scheduling: {
     enabled: boolean; -> Is sync schedule enabled?
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -10,6 +10,7 @@ require 'active_support/core_ext/hash'
 require 'connectors/connector_status'
 require 'connectors/sync_status'
 require 'utility'
+require 'app/config'
 
 module Core
   class ElasticConnectorActions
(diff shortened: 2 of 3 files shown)


sphilipse (approved):

lib/core/elastic_connector_actions.rb
```diff
@@ -82,11 +83,15 @@ def complete_sync(connector_id, job_id, status)
...
-                                :last_synced => Time.now)
+                                :last_synced => Time.now,
+                                :last_indexed_count => status[:indexed_document_count],
+                                :last_deleted_count => status[:deleted_document_count])
 
         body = {
           :doc => {
             :status => sync_status,
+            :indexed_document_count => status[:indexed_document_count],
+            :deleted_document_count => status[:deleted_document_count],
```
wangch079, line 94: It's already in `status`, see below

```
  }.merge(status)
```

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -96,6 +96,8 @@ This is our main communication index, used to communicate the connector's config
                            regularly (UTC)
   last_sync_error: string;   -> Optional last sync error message
   last_synced: date;    -> Date/time of last completed sync (UTC)
+  last_indexed_count: number;    -> How many documents were inserted into the index
+  last_deleted_count: number;    -> How many documents were deleted from the index
```
wangch079, line 100: Mappings of `.elastic-connectors` (see below) should also be updated.

wangch079 (commented):
For now, Connector service also handles the creation of system indices, so the 2 new fields should be added [here](https://github.com/elastic/connectors-ruby/blob/149868fdd12d62041e448b51b551d93daa7a7701/lib/core/elastic_connector_actions.rb#L212-L232), although this will be removed by https://github.com/elastic/enterprise-search-team/issues/2583

PR #283, opened 2022-08-30
add ingest fields to byoc definition

seanstory:
## Part of https://github.com/elastic/enterprise-search-team/issues/2607

This modifies the BYOC protocol (from the ruby connector client's side) to expect ingestion flags to be present in the `.elastic-connectors` document. If the settings are not there, they are looked up in the index's meta mappings. If that fails, it falls back to hardcoded default values.

These ingest settings impact which pipeline is used at request time to index documents, and what additional fields get set on the documents as they are sent to Elasticsearch.

#### Pre-Review Checklist

## Related Pull Requests

* default pipeline definition: https://github.com/elastic/ent-search/pull/6805
* kibana PR to make these settings available: TODO

## Release Note

New fields were added to the definition of the `.elastic-connectors` index to represent ingest pipeline options.

  - will be handled in https://github.com/elastic/enterprise-search-team/issues/2614

### DIFF
--- a/docs/CONNECTOR_PROTOCOL.md
+++ b/docs/CONNECTOR_PROTOCOL.md
@@ -90,7 +90,6 @@ This is our main communication index, used to communicate the connector's config
   };                    -> Definition and values of configurable
                            fields
   error: string;        -> Optional error message
-  extract_binary_content: boolean; -> Whether or not the `request_pipeline` should handle binary data
   index_name: string;   -> The index data will be written to
   language: string;     -> the language used for the analyzer
   last_seen: date;      -> Connector writes check-in date-time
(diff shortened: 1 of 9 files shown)


mchernyavskaya (approved):
LGTM

artem-shelkovnikov:
@seanstory late question - `self[:pipeline][:extract_binary_content] || @globals[:pipeline][:default_extract_binary_content]` will raise an exception if `self[:pipeline]` or `@globals[:pipeline]` is not defined. Is it a possible scenario?

seanstory:
It shouldn't be, no.

artem-shelkovnikov:
Currently on latest Kibana and connectors I cannot run a connector without receiving an error :/ 

Could it be that Kibana does not populate the field yet?

sphilipse:
Correct, Sean's pull request for Kibana hasn't merged yet.

Existing connector documents don't have this field either, though, so connectors really shouldn't error out if the field isn't present.

seanstory:
ah shoot, didn't consider how my lingering PR would cause issues. 🤦 
- let's change the connectors code to use `dig`
- let me try to quickly wrap up my kibana PR

seanstory:
fix: https://github.com/elastic/connectors-ruby/pull/288

PR #284, opened 2022-09-05
Update pre-flight check

wangch079:
#### Pre-Review Checklist


lib/app/preflight_check.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    #-------------------------------------------------------------------------------------------------
+    # Ensures that the version of Elasticsearch is compatible with connector service
+    def check_es_version!
+      info = client.info
+      version = info.dig('version', 'number')
+      fail_check!("Cannot retrieve version from Elasticsearch response:\n#{info.to_json}") unless version
+
+      if match_es_version?(version)
```
wangch079, line 56: I only checked if `Major.Minor` matches. There is more comprehensive [version check in ent-search](https://github.com/elastic/ent-search/blob/38279fff22c3da29b715a59a88684f1a13508a8d/shared_togo/lib/shared_togo/elasticsearch_checks.rb#L51-L61), which I'm not sure whether we want to adopt. @elastic/ingestion-team

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -96,8 +96,8 @@ This is our main communication index, used to communicate the connector's config
                            regularly (UTC)
   last_sync_error: string;   -> Optional last sync error message
   last_synced: date;    -> Date/time of last completed sync (UTC)
-  last_indexed_count: number;    -> How many documents were inserted into the index
-  last_deleted_count: number;    -> How many documents were deleted from the index
+  last_indexed_document_count: number;    -> How many documents were inserted into the index
```
wangch079, line 99: I renamed the field name as I suggested [here](https://github.com/elastic/connectors-ruby/pull/282#discussion_r959697636), to be more consistent.

cc @sphilipse in case this affects FE

wangch079 (commented):
I still kept `ensure_connectors_index_exists` and `ensure_job_index_exists` in [Core::ElasticConnectorActions](https://github.com/elastic/connectors-ruby/blob/3a72fd6838b101833629209735a5478c3d3b9455/lib/core/elastic_connector_actions.rb#L217-L258), because they are used by ftest.rb:
https://github.com/elastic/connectors-ruby/blob/3a72fd6838b101833629209735a5478c3d3b9455/tests/ftest.rb#L96-L97

@tarekziade , do you think we still want to keep the ftest in nightly QA? It contradicts the purpose of managing system indices in Kibana only.

lib/app/preflight_check.rb
```diff
@@ -0,0 +1,87 @@
...
+      parse_minor_version(App::VERSION) == parse_minor_version(es_version)
+    end
+
+    def parse_minor_version(version)
+      version.split('.').slice(0, 2).join('.')
+    end
+
+    def client
+      @client ||= Utility::EsClient.new(App::Config[:elasticsearch])
+    end
```
artem-shelkovnikov, line 81: Not 100% sure, but do we want to reuse EsClient that's used by the caller?

I think in the end we'd benefit from sticking with 1 EsClient per thread (if not thread-safe) or 1 EsClient per app, which will be passed to everyone who needs it.

artem-shelkovnikov (commented):
Added a couple comments, but looks good overall.

Great test coverage!

I saw only 1 case is not handled, might be worth to add a test case for that as well:
image

lib/app/preflight_check.rb
```diff
@@ -0,0 +1,87 @@
...
+
+require 'app/version'
+require 'utility/logger'
+
+module App
+  class PreflightCheck
+    class CheckFailure < StandardError; end
+
+    CONNECTORS_INDEX = '.elastic-connectors'
+    JOB_INDEX = '.elastic-connectors-sync-jobs'
```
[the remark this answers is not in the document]
wangch079: Good idea. will add it

lib/app/preflight_check.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    def self.run!
+      App::PreflightCheck.new.run!
```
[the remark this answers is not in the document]
wangch079: Agree. I think it's not necessary. I think [dispatcher.rb](https://github.com/elastic/connectors-ruby/blob/main/lib/app/dispatcher.rb) has the same pattern. I should probably change it as well.

lib/app/preflight_check.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    #-------------------------------------------------------------------------------------------------
+    # Ensures that the version of Elasticsearch is compatible with connector service
+    def check_es_version!
+      info = client.info
+      version = info.dig('version', 'number')
+      fail_check!("Cannot retrieve version from Elasticsearch response:\n#{info.to_json}") unless version
+
+      if match_es_version?(version)
+        Utility::Logger.info("Connector service version (#{App::VERSION}) matches Elasticsearch version (#{version}).")
```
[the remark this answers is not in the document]
wangch079: Basically I want to log it as `info` if it passes one check. See one example from [ent-search pre-flight-check](https://github.com/elastic/ent-search/blob/d4c9874d0f7f8bd1c2e386141ea9587d17ba1568/shared_togo/lib/shared_togo/elasticsearch_checks.rb#L140)

wangch079:
>I saw only 1 case is not handled, might be worth to add a test case for that as well

That case is when status is not `green`/`yellow`/`red`, which is impossible. I added an else simply because I don't like an if-else doesn't cover everything.

But I can add one spec example for that else case.

>This change might be worth to move to a different PR - it's not related to pre-flight checks thus harder to track where it's coming from.

Agree, and we had the same discussion before. I will make a separate PR.

lib/app/preflight_check.rb
wangch079: In one preflight-check `run!`, the client can be reused. The `App::PreflightCheck` class is only used at the beginning of the app, i.e. one pre-flight-check per process.

>I think in the end we'd benefit from sticking with 1 EsClient per thread (if not thread-safe) or 1 EsClient per app

I think currently `EsClient` is instantiable, and instantiated by `Core::OutputSink::EsSink` and `Core::ElasticConnectorActions`. This itself is worth a discussion (see [here](https://elastic.slack.com/archives/C03GAPTQ2JF/p1662436126055189)), and a separate PR.

docs/CONNECTOR_PROTOCOL.md
wangch079: Drafted a separate PR https://github.com/elastic/connectors-ruby/pull/287

lib/app/preflight_check.rb
artem-shelkovnikov: Yep, agreed that it needs to be addressed as a separate PR

artem-shelkovnikov (approved):
Looks good overall, just one minor comment about constants file for `CONNECTORS_INDEX` and `JOB_INDEX`

lib/app/preflight_check.rb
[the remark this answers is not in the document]
wangch079: Moved to `Utility::Constants`: https://github.com/elastic/connectors-ruby/pull/284/commits/bbc7edf1403e8271c3e102cd0de6c91dab0c0913

lib/app/preflight_check.rb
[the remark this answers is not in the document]
wangch079: Make `PreflightCheck` singleton class: https://github.com/elastic/connectors-ruby/pull/284/commits/ec13b075275928bc8ae134d8c56f8cd926c4cec8

spec/app/preflight_check_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      allow(described_class).to receive(:client).and_return(client)
+    end
+
+    context 'when Elasticsearch is not running' do
+      let(:cluster) { double }
+
+      it 'should retry multiple times and fail the check' do
+        stub_const('App::PreflightCheck::STARTUP_RETRY_INTERVAL', 1)
+        stub_const('App::PreflightCheck::STARTUP_RETRY_TIMEOUT', 3)
+        allow(client).to receive(:cluster).and_return(cluster)
```
artem-shelkovnikov, line 24: I'm not sure if it's a common pattern, but I feel that stubbing/allowing is more readable when it lives in `before(:each)` section, what do you think @wangch079 ?

wangch079: I see `allow` happens in both `before` and `it` and I have no strong preference. I can make it to `before`.

artem-shelkovnikov: It's not a strong preference from my side :)

tarekziade:
> I still kept `ensure_connectors_index_exists` and `ensure_job_index_exists` in [Core::ElasticConnectorActions](https://github.com/elastic/connectors-ruby/blob/3a72fd6838b101833629209735a5478c3d3b9455/lib/core/elastic_connector_actions.rb#L217-L258), because they are used by ftest.rb:
> 
> https://github.com/elastic/connectors-ruby/blob/3a72fd6838b101833629209735a5478c3d3b9455/tests/ftest.rb#L96-L97
> 
> @tarekziade , do you think we still want to keep the ftest in nightly QA? It contradicts the purpose of managing system indices in Kibana only.

We do, unless we find a way to automate this part using Kibana in the nightly QA.

PR #286, opened 2022-09-06
Add Core::Scheduler for Crawler

vidok:
To make Crawler more like a connector we need to start using `Core::Scheduler`. 

In this PR:

- Introduce a new `Crawler::Scheduler`
- A small `ElasticConnectorActions` refactoring
- Export files to `connectors_utility` to make them available in ent-search

#### Pre-Review Checklist

## Related Pull Requests

- https://github.com/elastic/ent-search/pull/6803
- https://github.com/elastic/kibana/pull/140475

### DIFF
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -7,7 +7,6 @@
 # frozen_string_literal: true
 #
 require 'active_support/core_ext/hash'
-require 'app/config'
 require 'connectors/connector_status'
 require 'connectors/sync_status'
 require 'utility'
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -129,13 +129,7 @@ module Core
 
         body = {
           :doc => {
-<<<<<<< HEAD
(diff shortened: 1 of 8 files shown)


lib/core/elastic_connector_actions.rb
```diff
@@ -44,6 +43,14 @@ def connectors_meta
       end
 
       def native_connectors
+        connectors({ term: { is_native: true } })
+      end
+
+      def crawler_connectors
+        connectors({ term: { service_type: Utility::Constants::CRAWLER_SERVICE_TYPE } })
```
artem-shelkovnikov, line 50: Do we want to also add a term to ignore crawler in all other places? Probably not a big deal to not do this, just thinking out loud

artem-shelkovnikov (approved):

lib/core/elastic_connector_actions.rb
vidok: Yes, I totally agree.

vidok: I'll add it in separate PR

PR #288, opened 2022-09-07
Fix brittle pipeline param access

seanstory:
see: https://github.com/elastic/connectors-ruby/pull/283#issuecomment-1239604384

### DIFF
--- a/lib/core/connector_settings.rb
+++ b/lib/core/connector_settings.rb
@@ -9,6 +9,7 @@
 require 'active_support/core_ext/hash/indifferent_access'
 require 'active_support/core_ext/object/blank'
 require 'connectors/connector_status'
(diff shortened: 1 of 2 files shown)


lib/core/connector_settings.rb
```diff
@@ -86,5 +91,12 @@ def initialize(es_response)
       @elasticsearch_response = es_response.with_indifferent_access
       @globals = ElasticConnectorActions.connectors_meta
     end
+
+    def return_if_present(*args)
```
seanstory, line 95: The original implementation had a bug that the `||` would cause us to ignore `false` values set in higher precedence locations. This should now return falsey values as long as they are not `nil`

artem-shelkovnikov (approved):

PR #289, opened 2022-09-08
Refactor Core::ConnectorSettings for more consistency

artem-shelkovnikov:
This PR just does a minor refactoring for `Core::ConnectorSettings` class.

Points addressed:

- Fixed overusage of `.with_indiffirent_access` when it's not actually needed -`ElasticConnectorActions.get_connector` already returns hash with indifferent access
- All data needed to initialise an instance of `Core::ConnectorSettings` is loaded in method caller instead of being loaded by the initializer.
- Removed `dig(*args*)` method as it's not used by anyone outside of class and replaced calls to it by explicit `@elasticsearch_response.dig` calls

#### Pre-Review Checklist

### DIFF
--- a/lib/core/connector_settings.rb
+++ b/lib/core/connector_settings.rb
@@ -26,10 +26,10 @@ module Core
 
     def self.fetch(connector_id)
       es_response = ElasticConnectorActions.get_connector(connector_id)
-        .with_indifferent_access
+      globals = ElasticConnectorActions.connectors_meta
 
       raise ConnectorNotFoundError.new("Connector with id=#{connector_id} was not found.") unless es_response[:found]
-      new(es_response)
+      new(es_response, globals)
     end
 
     def id
@@ -39,10 +39,6 @@ module Core
     def [](property_name)
       # TODO: handle not found
       @elasticsearch_response[:_source][property_name]
-    end
-
-    def dig(*args)
-      @elasticsearch_response.dig(*args)
     end
 
     def index_name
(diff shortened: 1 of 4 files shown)


lib/core/connector_settings.rb
```diff
@@ -70,26 +66,26 @@ def scheduling_settings
...
-      return_if_present(dig(:pipeline, :run_ml_inference), @globals.dig(:pipeline, :default_run_ml_inference), DEFAULT_RUN_ML_INFERENCE)
+      return_if_present(@elasticsearch_response.dig(:pipeline, :run_ml_inference), @globals.dig(:pipeline, :default_run_ml_inference), DEFAULT_RUN_ML_INFERENCE)
     end
 
     private
 
-    def initialize(es_response)
-      @elasticsearch_response = es_response.with_indifferent_access
-      @globals = ElasticConnectorActions.connectors_meta
+    def initialize(es_response, globals)
```
wangch079, line 86: I think we can have `globals` as optional:
```suggestion
    def initialize(es_response, globals=nil)
```

and have a method `globals` in `Core::ConnectorSettings`:

```
def globals
  @globals ||= ElasticConnectorActions.connectors_meta
end
```

artem-shelkovnikov: From what I understood from @seanstory's comment [here](https://github.com/elastic/connectors-ruby/pull/283#issuecomment-1239520285), global settings will always be present.

The calls to `connectors_meta` will be only in 2 methods that fetch connector settings: `ConnectorSettings.fetch_by_id` and `ConnectorSettings.fetch_native`(this one is now in ElasticConnectorActions), I think it should be okay to keep the calls inlined there.

PR #290, opened 2022-09-08
Remove harmful rescue/raise for native_connectors method

artem-shelkovnikov:
Small fix for logic that makes debugging harder.

This rescue statement transforms `StandardError` to `Utility::ClientError` that is not handled by the caller differently than StandardError, but also it loses the call stack of original error, so that newly thrown error will always point to this rescue block, not to the place where original error happened.

### DIFF
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -70,9 +70,6 @@ module Core
           offset += DEFAULT_PAGE_SIZE
         end
         result
-      rescue StandardError => e
-        Utility::ExceptionTracking.log_exception(e, 'Error while getting native connectors')
-        raise Utility::ClientError.new(e)
       end
 
       def update_connector_configuration(connector_id, configuration)


lib/core/elastic_connector_actions.rb
```diff
@@ -70,9 +70,6 @@ def native_connectors
           offset += DEFAULT_PAGE_SIZE
         end
         result
-      rescue StandardError => e
-        Utility::ExceptionTracking.log_exception(e, 'Error while getting native connectors')
-        raise Utility::ClientError.new(e)
```
navarone-feekery, line 75: On the off chance you want to keep this, I think you can also add something like `Utility::ExceptionTracking.log_exception(e, e.backtrace)` to get the stack trace.
https://ruby-doc.org/core-2.6/Exception.html#method-i-backtrace

navarone-feekery (approved):
LGTM!

lib/core/elastic_connector_actions.rb
artem-shelkovnikov: We can indeed, but if it's not caught I believe the caller will do it here: https://github.com/elastic/connectors-ruby/blob/main/lib/core/native_scheduler.rb

PR #291, opened 2022-09-08
Connector Schedule does not properly support "Every Week" setting

mchernyavskaya:
## Closes https://github.com/elastic/enterprise-search-team/issues/2508

It wasn't failing for all weekly schedules - only on Saturdays. Which, from the Kibana scheduler component, are saved with the index 7.

What was failing wasn't the [fugit](https://github.com/floraison/fugit) Cron parser we used - it was just [this regexp from our own code](https://github.com/elastic/connectors-ruby/blob/main/lib/utility/cron.rb#L13):

```ruby
CRON_REGEXP = /^\s*($|#|\w+\s*=|(\?|\*|(?:[0-5]?\d)(?:(?:-|\/|,)(?:[0-5]?\d))?(?:,(?:[0-5]?\d)(?:(?:-|\/|,)(?:[0-5]?\d))?)*)\s+(\?|\*|(?:[0-5]?\d)(?:(?:-|\/|,)(?:[0-5]?\d))?(?:,(?:[0-5]?\d)(?:(?:-|\/|,)(?:[0-5]?\d))?)*)\s+(\?|\*|(?:[01]?\d|2[0-3])(?:(?:-|\/|,)(?:[01]?\d|2[0-3]))?(?:,(?:[01]?\d|2[0-3])(?:(?:-|\/|,)(?:[01]?\d|2[0-3]))?)*)\s+(\?|\*|(?:0?[1-9]|[12]\d|3[01])(?:(?:-|\/|,)(?:0?[1-9]|[12]\d|3[01]))?(?:,(?:0?[1-9]|[12]\d|3[01])(?:(?:-|\/|,)(?:0?[1-9]|[12]\d|3[01]))?)*)\s+(\?|\*|(?:[1-9]|1[012])(?:(?:-|\/|,)(?:[1-9]|1[012]))?(?:L|W)?(?:,(?:[1-9]|1[012])(?:(?:-|\/|,)(?:[1-9]|1[012]))?(?:L|W)?)*|\?|\*|(?:JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC)(?:(?:-)(?:JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC))?(?:,(?:JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC)(?:(?:-)(?:JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC))?)*)\s+(\?|\*|(?:[0-6])(?:(?:-|\/|,|#)(?:[0-6]))?(?:L)?(?:,(?:[0-6])(?:(?:-|\/|,|#)(?:[0-6]))?(?:L)?)*|\?|\*|(?:MON|TUE|WED|THU|FRI|SAT|SUN)(?:(?:-)(?:MON|TUE|WED|THU|FRI|SAT|SUN))?(?:,(?:MON|TUE|WED|THU|FRI|SAT|SUN)(?:(?:-)(?:MON|TUE|WED|THU|FRI|SAT|SUN))?)*)(|\s)+(\?|\*|(?:|\d{4})(?:(?:-|\/|,)(?:|\d{4}))?(?:,(?:|\d{4})(?:(?:-|\/|,)(?:|\d{4}))?)*))$/
```

And it was failing because it was checking for days [0-6], instead of [1-7] (as Kibana stores them).

So it became obvious that there was a discrepancy somewhere in how the days are indexed.

In the Kibana scheduler component, a [weekday map looks like this](https://github.com/elastic/kibana/blob/main/src/plugins/es_ui_shared/public/components/cron_editor/services/humanized_numbers.ts#L16-L24):

```javascript
// The international ISO standard dictates Monday as the first day of the week, but cron patterns
// use Sunday as the first day, so we're going with the cron way.
const dayOrdinalToDayNameMap = {
  0: i18n.translate('esUi.cronEditor.day.sunday', { defaultMessage: 'Sunday' }),
  1: i18n.translate('esUi.cronEditor.day.monday', { defaultMessage: 'Monday' }),
  2: i18n.translate('esUi.cronEditor.day.tuesday', { defaultMessage: 'Tuesday' }),
  3: i18n.translate('esUi.cronEditor.day.wednesday', { defaultMessage: 'Wednesday' }),
  4: i18n.translate('esUi.cronEditor.day.thursday', { defaultMessage: 'Thursday' }),
  5: i18n.translate('esUi.cronEditor.day.friday', { defaultMessage: 'Friday' }),
  6: i18n.translate('esUi.cronEditor.day.saturday', { defaultMessage: 'Saturday' }),
};
```

But the day options of the component are [decremented when converted into ordinals](https://github.com/elastic/kibana/blob/main/src/plugins/es_ui_shared/public/components/cron_editor/constants.ts#L37-L40):

```
export const DAY_OPTIONS = makeSequence(1, 7).map((value) => ({
  value: value.toString(),
  text: getDayName((value - 1) as DayOrdinal),
}));
```

So it seems that the Kibana scheduler indexes day 1 to 7, Sun to Mon (which we already know).

This does not quite align with how the [cron specification does it](https://en.wikipedia.org/wiki/Cron). In crontab, Mon-Sat are 1 to 6, and Sunday can be either 0 or in some implementation also 7 (see also [this response on Stackoverflow](https://stackoverflow.com/questions/18919151/crontab-day-of-the-week-syntax)).

```
 ┌────────── minute (0 - 59)
 │ ┌──────── hour (0 - 23)
 │ │ ┌────── day of month (1 - 31)
 │ │ │ ┌──── month (1 - 12)
 │ │ │ │ ┌── day of week (0 - 6 => Sunday - Saturday, or
 │ │ │ │ │                1 - 7 => Monday - Sunday)
 ↓ ↓ ↓ ↓ ↓
 *  *   *  *  * command to be executed

 ```

The fugit library, which we use in our ruby code, [treats Monday as 1](https://github.com/floraison/fugit#the-first-monday-of-the-month). I also checked it with 0 and 7 and it treats both as SUN, which aligns with the cron expectations.

**So it looked like we could fix our implementation and make it work with a very simple change of fixing the regexp. However, it's not enough, since it won't work correctly - it'll wrongly recognize the day.**

This is why I also made an adjustment [described in this comment to the original issue](https://github.com/elastic/enterprise-search-team/issues/2508#issuecomment-1240680143). I adopted the [hack we had in connectors-python](https://github.com/elastic/connectors-python/blob/main/connectors/quartz.py#L71-L92) to work with what is produced by the Kibana scheduler.

Also, the Kibana implementation seems to be consistent with [Quartz](http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html). So if we're aiming for that one, then this PR is the way to go.

#### Pre-Review Checklist

### CODE REVIEW

File: spec/utility/cron_spec.rb [hunk 1]
```diff
@@ -29,11 +29,17 @@
...
+      ['0 0 12 ? * 2', '0 12 * * 1'], # every Monday at 12:00 PM
+      ['0 0 0 ? * 7', '0 0 * * 6'], # every Saturday at 12:00 AM
+      ['0 0 0 ? * 0', '0 0 * * 0'], # every Sunday at 12:00 AM
     ]
 
     conversions.each do |quartz, crontab|
       expect(subject.quartz_to_crontab(quartz)).to eq(crontab)
-      expect(Fugit::Cron.do_parse(crontab).next_time).to be > Time.now
+      next_time = Fugit::Cron.do_parse(crontab).next_time
+      print "### next_time for scheduler [#{quartz}], cron [#{crontab}]: [#{next_time}]\n"
```
tarekziade, line 41: should we remove this line?

lib/utility/cron.rb
```diff
@@ -10,14 +10,14 @@
...
+  CRON_REGEXP = /^\s*($|#|\w+\s*=|(\?|\*|(?:[0-5]?\d)(?:(?:-|\/|,)(?:[0-5]?\d))?(?:,(?:[0-5]?\d)(?:(?:-|\/|,)(?:[0-5]?\d))?)*)\s+(\?|\*|(?:[0-5]?\d)(?:(?:-|\/|,)(?:[0-5]?\d))?(?:,(?:[0-5]?\d)(?:(?:-|\/|,)(?:[0-5]?\d))?)*)\s+(\?|\*|(?:[01]?\d|2[0-3])(?:(?:-|\/|,)(?:[01]?\d|2[0-3]))?(?:,(?:[01]?\d|2[0-3])(?:(?:-|\/|,)(?:[01]?\d|2[0-3]))?)*)\s+(\?|\*|(?:0?[1-9]|[12]\d|3[01])(?:(?:-|\/|,)(?:0?[1-9]|[12]\d|3[01]))?(?:,(?:0?[1-9]|[12]\d|3[01])(?:(?:-|\/|,)(?:0?[1-9]|[12]\d|3[01]))?)*)\s+(\?|\*|(?:[1-9]|1[012])(?:(?:-|\/|,)(?:[1-9]|1[012]))?(?:L|W)?(?:,(?:[1-9]|1[012])(?:(?:-|\/|,)(?:[1-9]|1[012]))?(?:L|W)?)*|\?|\*|(?:JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC)(?:(?:-)(?:JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC))?(?:,(?:JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC)(?:(?:-)(?:JAN|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC))?)*)\s+(\?|\*|(?:[0-7])(?:(?:-|\/|,|#)(?:[0-7]))?(?:L)?(?:,(?:[0-7])(?:(?:-|\/|,|#)(?:[0-7]))?(?:L)?)*|\?|\*|(?:MON|TUE|WED|THU|FRI|SAT|SUN)(?:(?:-)(?:MON|TUE|WED|THU|FRI|SAT|SUN))?(?:,(?:MON|TUE|WED|THU|FRI|SAT|SUN)(?:(?:-)(?:MON|TUE|WED|THU|FRI|SAT|SUN))?)*)(|\s)+(\?|\*|(?:|\d{4})(?:(?:-|\/|,)(?:|\d{4}))?(?:,(?:|\d{4})(?:(?:-|\/|,)(?:|\d{4}))?)*))$/
```
tarekziade, line 13: should `0-7` be `1-7` instead?

tarekziade (approved):
Thanks! good findings 👍 

Could you document the code, to explain the discrepancy for `1-6` vs `0-7`
I think most of this issue description could be added in the ruby module.

spec/utility/cron_spec.rb
mchernyavskaya: I used it for debugging - to check which index means which actual day. I don't think they're harmful in tests.

mchernyavskaya:
> Thanks! good findings 👍
> 
> Could you document the code, to explain the discrepancy for `1-6` vs `0-7` I think most of this issue description could be added in the ruby module.

Yeah sure - added more comments to the code.

PR #292, opened 2022-09-08
Follow-up on PR-284

wangch079:
1. Add comments to warn not to use methods `ensure_connectors_index_exists` and `ensure_job_index_exists`
2. Use default `{}` if meta is not set in connector index.


lib/core/elastic_connector_actions.rb
```diff
@@ -40,7 +40,7 @@ def get_connector(connector_id)
       def connectors_meta
         alias_mappings = client.indices.get_mapping(:index => Utility::Constants::CONNECTORS_INDEX).with_indifferent_access
         index = get_latest_index_in_alias(Utility::Constants::CONNECTORS_INDEX, alias_mappings.keys)
-        alias_mappings.dig(index, 'mappings', '_meta')
+        alias_mappings.dig(index, 'mappings', '_meta') || {}
```
seanstory, line 43: 👍 good call

seanstory (approved):

PR #294, opened 2022-09-09
Add ECS logging

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/2666

This PR provide additional option of ECS logging.

#### Pre-Review Checklist


lib/utility/logger.rb
```diff
@@ -20,7 +21,7 @@ def level=(log_level)
       end
 
       def logger
-        @logger ||= ::Logger.new(STDOUT)
+        @logger ||= Settings[:ecs_logging] ? EcsLogging::Logger.new(STDOUT) : ::Logger.new(STDOUT)
```
wangch079, line 24: I face issue if I use `App::Config` instead of `Settings` here:
```
/lib/app/config.rb:57:in `ent_search_es_config': uninitialized constant Utility::Logger (NameError)
```
I guess it's because there is a circular dependency between `App::Config` and `Utility::Logger`.

I believe we purposely designed it so that `App::Config` is only accessed by code under `App` module. Shouldn't the config be globally available?

The `Settings` object is [made available globally](https://github.com/rubyconfig/config#accessing-the-settings-object), and we can stop other code from accessing it.

cc @elastic/ingestion-team

artem-shelkovnikov: It's a chicken-and-egg problem that doesn't have a good way to be resolved IMO.

App::Config provides settings that are required for `Utility::Logger` initialization, such as `Logger.level` for example, but also uses `Utility::Logger` to log things.

We already do some logger init in https://github.com/elastic/connectors-ruby/blob/main/lib/utility/environment.rb#L16, so you can probably put the logic that instantiates the logger there.

I think also we need to remove usage of Logger from `App::Config` class, at least for now until we find a better way to initialise cross-dependent classes on startup.

wangch079: >We already do some logger init in https://github.com/elastic/connectors-ruby/blob/main/lib/utility/environment.rb#L16, so you can probably put the logic that instantiates the logger there.

Yes, considered. But Utility::Logger is a singleton class and there's no `initialize` method (I don't plan to add one).

> I think also we need to remove usage of Logger from `App::Config` class, at least for now until we find a better way to initialise cross-dependent classes on startup.

If the `ent-search` elasticsearch config overriding logic is still there, the usage of Logger from `App::Config` is absolutely necessary. Instead of removing `Logger`, I'm seriously considering removing the overriding logic. See my slack proposal [here](https://elastic.slack.com/archives/C03GAPTQ2JF/p1662704552916849).

artem-shelkovnikov: I honestly feel that we should move away from logger being a singleton and instantiate it when needed. App::Config can instantiate a regular one when doing initialization, workers will instantiate another one and pass to the classes that need it.

Gemfile
```diff
@@ -17,6 +17,7 @@ gem 'tzinfo', '1.2.10'
 gem 'nokogiri', '>= 1.13.7', :require => false
 gem 'fugit', '~> 1.5.3'
 gem 'remedy', '~> 0.3.0'
+gem 'ecs-logging'
```
seanstory, line 20: This isn't used by ent-search, and instead looks like we rolled our own. Any idea why? Will the output of this gem mix nicely with the output of ent-search? Without any tests here, it's hard to understand what the resulting output will look like.
See:
- https://github.com/elastic/ent-search/blob/main/shared_togo/lib/shared_togo/json_log_formatter.rb
- https://github.com/elastic/ent-search/blob/main/lib/ecs_formatter.rb
- https://github.com/elastic/ent-search/blob/main/lib/json_formatter.rb

wangch079: When I tested it locally, I saw the messages were printed in the terminal in ECS-compatible JSON, something like this:
```
{"@timestamp":"2022-09-09T17:16:48.390Z","log.level":"INFO","message":"ENT_SEARCH_CONFIG_PATH is not found, use connector service config.","ecs.version":"1.4.0"}
```

How is it shipped to Elasticsearch? I think it's handled by Filebeat. According to this [guide](https://www.elastic.co/guide/en/ecs-logging/ruby/current/setup.html#setup-step-3), we only need to make sure our app logs to stdout/stderr.

tarekziade: Our app needs to spit out in stderr or stdout lines that are compatible with filebeat, and then we redirect those streams to a log file on cloud. It's done in Py and I can see the Py logs in the Cloud console.

This output won't interfere with ES since it's not in the same stdout/stderr

seanstory: Just being "ecs compatible" isn't sufficient - we need to make sure that the same data is actually going to the same fields. Like if one of us is shipping `"log.level"` and the other one is shipping `"logger.level"`, support will end up having a real headache when they try to filter for ERROR messages only. ECS is pretty flexible, but we probably need our solution to be strongly consistent in its usage.

tarekziade: Sure, sounds awesome. `ecs-logging` is just converting a log line into a line filebeat is understanding -- it would probably be good to use it in ent-search as well.

Using the Cloud console, it's super easy to follow logs because we can differentiate sources through filters. But I agree that a standard is always better. If you want to start standardizing key names across services in a doc it would be a great thing to have for the future.

Now for this patch, we're just activating a *format*, not what fields are actually logged.

seanstory: > If you want to start standardizing key names across services in a doc it would be a great thing to have for the future.

I don't think that's necessary. I'm suggesting that we instead lift the JSON log formatter from ent-search, to ensure consistency. It's easy to just do this the right way, now.

> it's super easy to follow logs because we can differentiate sources through filters.

That's the thing - those filters only work if all the logs in that index are utilizing the same fields. I remember from your POC that you're piping the native connector logs into the `app_server.log` file for Filebeat to read. That means we'd have mixed content in the "app-server" source for the Enterprise Search logs. I don't think I'd be able to look Pius in the eye and explain to him why we thought it was worth it to not have consistent logging when he's trying to troubleshoot enteprise search+connectors, and is fighting to get all the right logs visible in discover.

tarekziade: > we instead lift the JSON log formatter from ent-search, to ensure consistency. It's easy to just do this the right way, now.

lib/utility/logger.rb
wangch079: @artem-shelkovnikov , let's have a separate discussion (and a separate PR) on how to refactor the logger. Any concern if I keep the usage of `Settings` here? Or any more comments on the ECS logging itself?

artem-shelkovnikov: Absolutely, let's do it in another PR!

I also see there's an ongoing discussion about the format of logs that I don't have much to add to, so my concerns were addressed, thanks!

Gemfile
wangch079: @tarekziade @seanstory do you have any more concerns?

artem-shelkovnikov (approved):

lib/utility/logger.rb
```diff
@@ -20,12 +21,16 @@ def level=(log_level)
...
       end
 
       SUPPORTED_LOG_LEVELS.each do |level|
         define_method(level) do |message|
-          logger.public_send(level, message)
+          if logger.is_a?(EcsLogging::Logger)
+            logger.public_send(level, message, service: { name: 'connectors-ruby' })
```
wangch079, line 30: Hi @tarekziade , according to [ECS documentation](https://www.elastic.co/guide/en/ecs/8.4/ecs-custom-fields-in-ecs.html), custom fields should be added under `labels`, but we should try to utilize the defined fields first.

I found that [service.name](https://www.elastic.co/guide/en/ecs/8.4/ecs-service.html#field-service-name) is the most appropriate field for `app_source`:

```
{"@timestamp":"2022-09-15T03:14:56.569Z","log.level":"INFO","message":"Successfully connected to Elasticsearch","ecs.version":"1.4.0","service":{"name":"connectors-ruby"}}
```

config/connectors.yml.example
```diff
@@ -19,6 +19,7 @@ thread_pool:
   max_queue: 100
 
 log_level: info
+ecs_logging: true
```
wangch079, line 22: I try to make it 
```
ecs_logging: ${ECS_LOGGING:false}
```

But it fails validation when `ECS_LOGGING` is not set:
```
ecs_logging: must be boolean
```

I have no idea why it doesn't work (when `ECS_LOGGING` is set to `ture` or `false`, it works fine), and I will leave it true by default.

wangch079:
Hi @tarekziade , feel free to merge it if you think it's good to go.

Gemfile
seanstory: > Does "lift" means "use" here?

Yes - I'm suggesting that the same log formatters be used. I think copy-paste from ent-search to connectors-ruby would be my vote at this late stage.

> If so, why not use the standard ecs-logging packages that are maintained by the Elastic clients team and documented here https://www.elastic.co/guide/en/ecs-logging/overview/current/intro.html ?

Changing how all of ent-search does logging seems like a bad idea to do 2 days before FF. As connectors-ruby is in tech-preview, I think it's much more reasonable to make it use an established logging pattern that the rest of our solution uses, rather than to use the ecs gem to avoid copying 3 small classes.


> Being able to distinguish each emitter gets us what we need afaik

No, I think that's a misunderstanding. Support will want to be able to look at ALL Enterprise Search logs in a single view when troubleshooting. If we can't given them a consistent view, where connectors stack traces are in the same subfield as crawler stack traces, they will be frustrated with us. We have a responsibility to our support team to make their work as easy as possible, especially when it costs us very little to do so.

> @tarekziade @seanstory do you have any more concerns?

Sorry, Chenhui, I was out sick yesterday. Didn't mean to leave you hanging here. My review request here is to either prove via unit tests that we're logging all the same info in all the same fields as ent-search, or to copy ent search's log formatters into this repo rather than use the ECS gem.

tarekziade (approved):

Gemfile
wangch079: Hi @seanstory , no worries.

I digged into `ent-search` logging, and found that although it didn't use `ecs-logging-ruby` (I don't know why), it does exactly the same thing as ECS logging does:
1. It adds [ecs.version](https://github.com/elastic/ent-search/blob/ed98b152d8e28c15afe482cde71d0f2aa6db2cf6/lib/ecs_formatter.rb#L7)
2. It adds [@timestamp](https://github.com/elastic/ent-search/blob/ed98b152d8e28c15afe482cde71d0f2aa6db2cf6/lib/ecs_formatter.rb#L13) and [log.level](https://github.com/elastic/ent-search/blob/ed98b152d8e28c15afe482cde71d0f2aa6db2cf6/shared_togo/lib/shared_togo/json_log_formatter.rb#L9)

It also adds some other fields [here](https://github.com/elastic/ent-search/blob/ed98b152d8e28c15afe482cde71d0f2aa6db2cf6/shared_togo/lib/shared_togo/json_log_formatter.rb#L8-L15), e.g. it uses [service.type](https://www.elastic.co/guide/en/ecs/8.4/ecs-service.html#field-service-type) (in this case `enterprise-search`) instead of [service.name](https://www.elastic.co/guide/en/ecs/8.4/ecs-service.html#field-service-name) as in this PR, I don't know which field makes more sense, but I'm fine to change it in `connectors-ruby`.

wangch079: > it didn't use `ecs-logging-ruby` (I don't know why)

I guess I know why. `ecs-logging-ruby` was only made available end of 2020:
```
curl https://api.github.com/repos/elastic/ecs-logging-ruby

{
  ...
  "created_at": "2020-11-23T14:54:00Z",
  ...
}
```

I definitely think `ent-search` should use `ecs-logging-ruby` (`ent-search` custom ecs-logging formatter still logs `ecs.version` as `1.7.0` while `ecs-logging-ruby` logs as `1.14.0` now), but agree that we don't want to change it so close to FF.

I think it's reasonable to make `connectors-ruby` (and `connectors-python`) to log the same fields as `ent-search` does [here](https://github.com/elastic/ent-search/blob/ed98b152d8e28c15afe482cde71d0f2aa6db2cf6/shared_togo/lib/shared_togo/json_log_formatter.rb#L8-L15). WDYT? @elastic/ingestion-team

seanstory: > I think it's reasonable to make connectors-ruby (and connectors-python) to log the same fields as ent-search does [here](https://github.com/elastic/ent-search/blob/ed98b152d8e28c15afe482cde71d0f2aa6db2cf6/shared_togo/lib/shared_togo/json_log_formatter.rb#L8-L15). WDYT? https://github.com/orgs/elastic/teams/ingestion-team

What happened here? I thought we'd reached agreement, but it's not clear to me that this PR did end up using the same fields as ent-search. Was this merged prematurely?

tarekziade: >  Was this merged prematurely?

Chenhui said 5 days ago we could merge it as it worked, so I did in order to be able to verify that it worked on cloud. It does! I can see the logs in the cloud console and it has the nice `service.name` field.

If we want to tweak the fields name, let's do this as a follow-up PR

wangch079: >but it's not clear to me that this PR did end up using the same fields as ent-search.

No it does not. as @tarekziade mentioned, this will be addressed in separate PRs.

Created 2 issues to track this:
1. https://github.com/elastic/enterprise-search-team/issues/2795
2. https://github.com/elastic/enterprise-search-team/issues/2796

PR #295, opened 2022-09-09
Improve and refactor health checks; add healthcheck before a sync

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/2667

Several very related changes bundled in one:

- Changed `Connectors::Base::Connector` method signature from `source_status` to `is_healthy?`. The method now always returns just a true or false.
- Changed base `Connectors::Base::Connector` method `health_check` to `do_health_check` - I think that this method method gives a better hint that this method is not supposed to return anything, but just "do something and raise an error"
- Added `Connectors::Base::Connector#do_health_check!` that raises `HealthCheckFailedError`if health check fails. 
- To improve on the previous point, I changed `Example::Connector.do_health_check` to include a comment that explains it and sample `raise` statement that user can uncomment and see what happens.
- Now when connector tries to run a sync, it always attempts to also do a health check - this can also help surface the problems with how `health_check` function is written. I think eventually health_check will be removed from heartbeat as heartbeat will only update `last_seen` for the connector.

#### Pre-Review Checklist

### DIFF
--- a/lib/connectors/base/connector.rb
+++ b/lib/connectors/base/connector.rb
@@ -9,6 +9,7 @@
 require 'bson'
 require 'core/output_sink'
 require 'utility/exception_tracking'
+require 'utility/errors'
 require 'app/config'
 
 module Connectors
@@ -35,6 +36,13 @@ module Connectors
 
       def do_health_check(params)
         raise 'Not implemented for this connector'
+      end
+
+      def do_health_check!(params)
+        do_health_check(params)
+      rescue StandardError => e
+        Utility::ExceptionTracking.log_exception(e, "Connector for service #{self.class.service_type} failed the health check for 3rd-party service.")
+        raise Utility::Errors::HealthCheckFailedError.new
       end
 
       def is_healthy?(params = {})
--- a/lib/connectors/base/connector.rb
+++ b/lib/connectors/base/connector.rb
@@ -33,6 +33,10 @@ module Connectors
 
       def yield_documents; end
 
+      def do_health_check(params)
+        raise 'Not implemented for this connector'
+      end
+
       def is_healthy?(params = {})
         do_health_check(params)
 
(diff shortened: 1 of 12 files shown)


lib/connectors/base/connector.rb
```diff
@@ -33,11 +34,24 @@ def initialize(local_configuration: {}, remote_configuration: {})
...
       def yield_documents; end
 
-      def source_status(params = {})
-        health_check(params)
-        { :status => 'OK', :statusCode => 200, :message => "Connected to #{self.class.display_name}" }
+      def do_health_check(_params)
+        raise 'Not implemented for this connector'
+      end
+
+      def do_health_check!(params)
```
wangch079, line 41: I'm thinking we should make `do_health_check!` and `is_healthy?` non-overridable. But it seems there's no `final` keyword in Ruby as in Java.

Another point I would like to discuss, our `connectors-ruby` is mixed of connector service (the framework) and the connectors (the logic of each connector, in dir `lib/connectors`). 

The `Connectors::Base::Connector` is the only ambiguous file. It's in dir `lib/connectors`, but we certainly don't want anyone who wants to customize to modify this file.

Just my random thoughts, and I don't have any better suggestions.

wangch079 (approved):

lib/connectors/base/connector.rb
artem-shelkovnikov: > I'm thinking we should make do_health_check! and is_healthy? non-overridable. But it seems there's no final keyword in Ruby as in Java.

I don't think it's possible indeed, the only way to make them non-overridable is not exposing them as a part of a base class, therefore splitting the abstraction into two, and I honestly think it's the best way to move forward, we can think how to make it nice in a different PR.

> Another point I would like to discuss, our connectors-ruby is mixed of connector service (the framework) and the connectors (the logic of each connector, in dir lib/connectors).

The Connectors::Base::Connector is the only ambiguous file. It's in dir lib/connectors, but we certainly don't want anyone who wants to customize to modify this file.

+1 to that, very very important points. I'd +1 moving them away and leaving only actual connectors in the `connectors` directory.

PR #296, opened 2022-09-09
Make it work on cloud [WONT BE MERGED]

tarekziade:
Changes made to make it work

- [x] Update `mongo` to `2.18` so we can use the `mongodb+srv` driver with the DNS resolver for Atlas
- [x] added user/password support -- it would be nice to support Mongo Atlas default notation `mongodb+srv://...` that how they tell you to access when using `cloud.mongodb.com`
- [x] Close mongo clients to avoid Threads and Memory leaks added https://github.com/elastic/enterprise-search-team/issues/2706

Data status:

I get 

```
E, [2022-09-10T18:12:45.668522 #2102] ERROR -- : Preflight check failed: Required system indices for connector service don't exist. Make sure to run Kibana first to create system indices.
2022-09-10 18:12:45,681 INFO exited: connectors-ruby (exit status 0; not expected)
```

with the latest Kibana snapshot. When are the indices created ? cc @sphilipse @byronhulcher 
@wangch079 the service needs to try this in a loop, not exit.  Also the exit status is 0, should be -1

Weird bugs 

- [x] I had to do a bunch of `.transform_keys(&:to_sym)` calls to make the code work. is that because we need to have `active_support/core_ext/hash` explicitly? 
- [x] for some reason, the mongo client loses its `@host` and other values set in `initialize` when I use them in `yield_documents` -- for now I hardcoded them
- [ ] if I use log_level=debug it grows indefinitely in memory ...

Syncing

- [ ] the bulk batch size should be configurable - 50 can probably be bumped to 100 on our cloud -- I surfaced it in my branch as a config option
- [x] we should surface all the pools sizes and timeouts options

Sizing:

- [x] added client options for Mongo. it would be good to surface those in the config -- the mongo client has a ton of tweaks available
- [x] I don't think we need that many `min_threads` in the pool by default, at least with the MongoDB connector that creates itself a bunch of threads already when it runs
- [x] added timeouts and retries on both sides
- [ ] the Airbnb collection (5555 documents) sync makes the service settle around ~250MB-300MB


wangch079 (approved):

wangch079:
>@wangch079 the service needs to try this in a loop, not exit. Also the exit status is 0, should be -1

Now we have 3 checks:

https://github.com/elastic/connectors-ruby/blob/3df386559cebc68c44d9c8041a0765258e8415b5/lib/app/preflight_check.rb#L23-L25

The first one is checking the elasticsearch connection, and it uses hardcoded retry interval and timeout, do you think we should make it configurable?

The second check is to make sure connector service version matches Elasticsearch version. If it doesn't match, we just exist. I think we can keep this without retrying.

The last check is to make sure the system indices exist. I can change it to keep looping until timeout. I will use the same retry interval and timeout as the first one.

tarekziade:
> > @wangch079 the service needs to try this in a loop, not exit. Also the exit status is 0, should be -1
> 
> Now we have 3 checks:
> 
> https://github.com/elastic/connectors-ruby/blob/3df386559cebc68c44d9c8041a0765258e8415b5/lib/app/preflight_check.rb#L23-L25
> 
> The first one is checking the elasticsearch connection, and it uses hardcoded retry interval and timeout, do you think we should make it configurable?

ultimately yes, but for now I think we don't care as long as hardcoded defaults make sense

> 
> The second check is to make sure connector service version matches Elasticsearch version. If it doesn't match, we just exist. I think we can keep this without retrying.

Yes this should not happen on cloud 

> 
> The last check is to make sure the system indices exist. I can change it to keep looping until timeout. I will use the same retry interval and timeout as the first one.

Yes let's loop on this one, thanks

and when something makes it exit, the exit code should be different from 0 always

sphilipse:
System indices are created whenever a user with sufficient permissions opens one of the Kibana Enterprise Search Content pages @tarekziade

tarekziade:
> System indices are created whenever a user with sufficient permissions opens one of the Kibana Enterprise Search Content pages @tarekziade

gotcha -- so yes, we need the service to idle in a loop until that happens when we deploy the stack.

PR #297, opened 2022-09-09
Make connector service gracefully handle errors

wangch079:
#### Pre-Review Checklist

### DIFF
--- a/lib/app/dispatcher.rb
+++ b/lib/app/dispatcher.rb
@@ -17,6 +17,9 @@ module App
   class Dispatcher
     POLL_IDLING = (App::Config[:idle_timeout] || 60).to_i
     TERMINATION_TIMEOUT = (App::Config[:termination_timeout] || 60).to_i
+    MIN_THREADS = (App::Config.thread_pool&.min_threads || 0).to_i
+    MAX_THREADS = (App::Config.thread_pool&.max_threads || 5).to_i
+    MAX_QUEUE = (App::Config.thread_pool&.max_queue || 100).to_i
 
     @running = Concurrent::AtomicBoolean.new(false)
 
@@ -45,9 +48,9 @@ module App
 
       def pool
         @pool ||= Concurrent::ThreadPoolExecutor.new(
-          min_threads: 5,
-          max_threads: 10,
-          max_queue: 100,
+          min_threads: MIN_THREADS,
+          max_threads: MAX_THREADS,
+          max_queue: MAX_QUEUE,
           fallback_policy: :abort
         )
       end
--- a/lib/app/dispatcher.rb
+++ b/lib/app/dispatcher.rb
@@ -18,62 +18,82 @@ module App
     POLL_IDLING = (App::Config[:idle_timeout] || 60).to_i
     TERMINATION_TIMEOUT = (App::Config[:termination_timeout] || 60).to_i
 
-    attr_reader :scheduler
+    @running = Concurrent::AtomicBoolean.new(false)
 
-    def initialize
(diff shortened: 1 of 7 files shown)


tarekziade:
Check-list -- the service needs to handle those cases gracefully without exiting, and being stable in memory and threads usage (after a few syncs it should not continue to grow)

- [ ] configure the service to run on a local Elasticsearch and distant MongoDB (cloud.mongodb.com ask Tarek for an access)
- [ ] run the service without Elasticsearch or MongoDB started -- it should loop until they are ready
- [ ] sync the data
- [ ] stop Elasticsearch during a new sync and restart it
- [ ] stop MongoDB during a new sync and restart it
- [ ] let the service run for 2-3 hours, rince and repeat

wangch079:
> Check-list -- the service needs to handle those cases gracefully without exiting, and being stable in memory and threads usage (after a few syncs it should not continue to grow)
> 
> * [ ]  configure the service to run on a local Elasticsearch and distant MongoDB (cloud.mongodb.com ask Tarek for an access)
> * [ ]  run the service without Elasticsearch or MongoDB started -- it should loop until they are ready
> * [ ]  sync the data
> * [ ]  stop Elasticsearch during a new sync and restart it
> * [ ]  stop MongoDB during a new sync and restart it
> * [ ]  let the service run for 2-3 hours, rince and repeat

I used the `Connectors::MongoDB` in @tarekziade 's PR https://github.com/elastic/connectors-ruby/pull/296, which threw error like this:
```
/Users/chenhui.wang/.rbenv/versions/2.6.9/lib/ruby/gems/2.6.0/gems/mongo-2.4.3/lib/mongo/uri.rb:339:in `raise_invalid_error!': Bad URI: mongodb+srv://elastic:{password}@cluster0.9n16jhs.mongodb.net/sample_airbnb (Mongo::Error::InvalidURI)
Invalid scheme. Scheme must be 'mongodb://'
MongoDB URI must be in the following format: mongodb://[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]][/[database][?options]]
```

I changed it to `mongodb://` instead of `mongodb+srv://` then I got
```
/Users/chenhui.wang/.rbenv/versions/2.6.9/lib/ruby/gems/2.6.0/gems/mongo-2.4.3/lib/mongo/server_selector/selectable.rb:115:in `select_server': No server is available matching preference: #<Mongo::ServerSelector::Primary:0x70130603130380 tag_sets=[] max_staleness=nil> using server_selection_timeout=30 and local_threshold=0.015 (Mongo::Error::NoServerAvailable)
```

Both errors mentioned above are handled well and the loop keeps running. I don't have the credentials to log into Mongo Cloud to stop the instance, but I think it will be another error handled by the connector service if I do. I will leave the service running for another 2-3 hours as suggested in the last item.

lib/app/dispatcher.rb
```diff
@@ -18,62 +18,82 @@ class Dispatcher
...
-    def shutdown
-      Utility::Logger.info("Shutting down dispatcher with pool [#{@pool&.class}] and scheduler [#{@scheduler&.class}]...")
-      @is_shutting_down = true
-      @scheduler&.shutdown
-      @pool.shutdown
-      @pool.wait_for_termination(TERMINATION_TIMEOUT)
-    end
+      def pool
+        @pool ||= Concurrent::ThreadPoolExecutor.new(
+          min_threads: 5,
```
tarekziade, line 48: could you set the default to min_thread: 0 max: 5 and surface the options in the config file (can be a follow up bug)

tarekziade (approved):
LGTM - one small remark on surfacing the pool options

PR #298, opened 2022-09-12
Add system indices retries and change exit code

wangch079:
A follow-up on https://github.com/elastic/connectors-ruby/pull/284, as suggested by @tarekziade in https://github.com/elastic/connectors-ruby/pull/296#issue-1368091963:

>the service needs to try this in a loop, not exit. Also the exit status is 0, should be -1

#### Pre-Review Checklist

### DIFF
--- a/lib/app/app.rb
+++ b/lib/app/app.rb
@@ -24,5 +24,6 @@ module App
     worker.start!
   rescue App::PreflightCheck::CheckFailure => e
     Utility::Logger.error("Preflight check failed: #{e.message}")
+    exit(-1)
   end
 end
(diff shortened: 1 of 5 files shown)


lib/app/app.rb
```diff
@@ -24,5 +24,6 @@ module App
     worker.start!
   rescue App::PreflightCheck::CheckFailure => e
     Utility::Logger.error("Preflight check failed: #{e.message}")
+    exit(-1)
```
tarekziade, line 27: the rule of thumb is that if anything bad happens when you exit you need something different than zero so the supervisor knows the app is unhealthy

tarekziade (approved):
LGTM

PR #299, opened 2022-09-12
Connector service refactor

wangch079:
Some bug fixing and refactors:

1. changes to `connector_settings` method of `NativeScheduler` and `SingleScheduler`, to make sure they always return `[]` if something bad happens
2. Add `.with_indifferent_access` to `initialize` method of `Core::ConnectorSettings`. I remembered @artem-shelkovnikov made the change to remove it, but I think it's reasonable to keep it as we use symbolic keys in `Core::ConnectorSettings` and it does no harm to call `with_indifferent_access` twice (I actually encountered an issue because of this).
3. Move `initialize` method of `Core::ConnectorSettings` to public. I don't understand how it worked before, but we have method calls like below, and I think it's reasonable to make it public.

https://github.com/elastic/connectors-ruby/blob/2bc25637418b142cdfb2138c1201bd351aae49c6/lib/core/elastic_connector_actions.rb#L66

#### Pre-Review Checklist

### DIFF
--- a/lib/core/connector_settings.rb
+++ b/lib/core/connector_settings.rb
@@ -32,6 +32,11 @@ module Core
       new(es_response, globals)
     end
 
+    def initialize(es_response, globals)
+      @elasticsearch_response = es_response.with_indifferent_access
+      @globals = globals.with_indifferent_access
+    end
+
     def id
       @elasticsearch_response[:_id]
     end
@@ -82,11 +87,6 @@ module Core
     end
 
     private
-
-    def initialize(es_response, globals)
-      @elasticsearch_response = es_response
-      @globals = globals
-    end
 
     def return_if_present(*args)
       args.each do |arg|
--- a/lib/core/native_scheduler.rb
+++ b/lib/core/native_scheduler.rb
@@ -18,6 +18,7 @@ module Core
       Core::ElasticConnectorActions.native_connectors || []
     rescue StandardError => e
       Utility::ExceptionTracking.log_exception(e, 'Could not retrieve native connectors due to unexpected error.')
+      []
     end
   end
 end
--- a/lib/core/single_scheduler.rb
+++ b/lib/core/single_scheduler.rb
@@ -21,6 +21,9 @@ module Core
     def connector_settings
       connector_settings = Core::ConnectorSettings.fetch(@connector_id)
       [connector_settings]
+    rescue StandardError => e
+      Utility::ExceptionTracking.log_exception(e, "Could not retrieve the connector by id #{@connector_id} due to unexpected error.")
+      []
     end
   end
 end


artem-shelkovnikov (approved):

lib/core/connector_settings.rb
```diff
@@ -32,6 +32,11 @@ def self.fetch(connector_id)
       new(es_response, globals)
     end
 
+    def initialize(es_response, globals)
+      @elasticsearch_response = es_response.with_indifferent_access
+      @globals = globals.with_indifferent_access
```
artem-shelkovnikov, line 37: Also about `with_indifferent_access` - I've encountered this problem only when in tests, do you remember how it happened for you?

vidok: I just encountered this problem in dev:

```
swiftype(dev)> Core::ElasticConnectorActions.connectors({ term: { is_native: true } }).first.index_name
       16: from railties (5.2.8.1) lib/rails/commands.rb:18:in `<main>'
       15: from railties (5.2.8.1) lib/rails/command.rb:46:in `invoke'
       14: from railties (5.2.8.1) lib/rails/command/base.rb:69:in `perform'
       13: from thor (0.20.3) lib/thor.rb:387:in `dispatch'
       12: from thor (0.20.3) lib/thor/invocation.rb:126:in `invoke_command'
       11: from thor (0.20.3) lib/thor/command.rb:27:in `run'
       10: from railties (5.2.8.1) lib/rails/commands/console/console_command.rb:96:in `perform'
        9: from railties (5.2.8.1) lib/rails/commands/console/console_command.rb:19:in `start'
        8: from railties (5.2.8.1) lib/rails/commands/console/console_command.rb:64:in `start'
        6: from org/jruby/RubyKernel.java:1237:in `catch'
        5: from org/jruby/RubyKernel.java:1507:in `loop'
        4: from org/jruby/RubyKernel.java:1091:in `eval'
        3: from (irb):1:in `evaluate'
        2: from connectors_utility (8.5.0.0.pre.20220912T090241Z) lib/core/connector_settings.rb:45:in `index_name'
        1: from connectors_utility (8.5.0.0.pre.20220912T090241Z) lib/core/connector_settings.rb:41:in `[]'
NoMethodError (undefined method `[]' for nil:NilClass)
```

vidok:
@wangch079 Could you please add some unit tests to cover [internal access](https://github.com/elastic/connectors-ruby/blob/bd54be06e25c02bac0cbd561ed589027a3356d1f/lib/core/connector_settings.rb#L32) to elastic search response? 

It seems like it's been broken and no one noticed that.

artem-shelkovnikov:
I've attempted to refactor the code that does internal access in this PR: https://github.com/elastic/connectors-ruby/pull/300

lib/core/connector_settings.rb
vidok: > @vidok , `Core::ElasticConnectorActions.connectors`, is this a new method you defined locally?
> 
> Hi @artem-shelkovnikov , I believe it's here:
> 
> https://github.com/elastic/connectors-ruby/blob/233288027638c3a5212b3d0d4d98428290db91c3/lib/core/elastic_connector_actions.rb#L71
> 
> you will get `nil` for `hit[:_source]` but `hit['_source']` works fine.

It's not local anymore https://github.com/elastic/connectors-ruby/blob/main/lib/core/elastic_connector_actions.rb#L53. 

I made it to support different query terms when looking up connector documents.

PR #300, opened 2022-09-12
Resolve circular dependency between ElasticConnectorActions and ConnectorSettings

artem-shelkovnikov:
Found a problem that `ElasticConnectorActions` and `ConnectorSettings` classes have a circular dependency. 

It happens because `Core::ElasticConnectorActions.native_connectors` attempts to create instances of `Core::ConnectorSettings` inside of the method, while `Core::ConnectorSettings.fetch_by_id` instead just calls `Core::ElasticConnectorActions.get_connector` to get the connector and does creation on its side.

Resolution is - make `Core::ElasticConnectorActions.native_connectors` more primitive "query" method that does not attempt to stitch pages/join results and move the stitching/creation logic into a method in `Core::ConnectorSettings`, see `Core::ConnectorSettings.fetch_native` method.

As a side effect added a test that checks that when fetching native connectors we ask the server for `connectors_meta` only once.

### DIFF
--- a/lib/core/connector_settings.rb
+++ b/lib/core/connector_settings.rb
@@ -21,10 +21,12 @@ module Core
     DEFAULT_REDUCE_WHITESPACE = true
     DEFAULT_RUN_ML_INFERENCE = false
 
+    DEFAULT_PAGE_SIZE = 100
+
     # Error Classes
     class ConnectorNotFoundError < StandardError; end
 
-    def self.fetch(connector_id)
+    def self.fetch_by_id(connector_id)
       es_response = ElasticConnectorActions.get_connector(connector_id)
       globals = ElasticConnectorActions.connectors_meta
 
@@ -35,6 +37,14 @@ module Core
     def initialize(es_response, globals)
       @elasticsearch_response = es_response.with_indifferent_access
       @globals = globals.with_indifferent_access
+    end
+
+    def self.fetch_native_connectors(page_size = DEFAULT_PAGE_SIZE)
+      fetch_connectors_by_query({ term: { is_native: true } })
+    end
+
+    def self.fetch_crawler_connectors
+      fetch_connectors_by_query({ term: { service_type: Utility::Constants::CRAWLER_SERVICE_TYPE } })
     end
 
     def id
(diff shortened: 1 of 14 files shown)


artem-shelkovnikov:
@wangch079 I think this PR will also address your point no.3 from this PR: https://github.com/elastic/connectors-ruby/pull/299#event-7366058374

> Move initialize method of Core::ConnectorSettings to public. I don't understand how it worked before, but we have method calls like below, and I think it's reasonable to make it public.

Though again, I have no idea how it could have worked, I even see @seanstory did a workaround when writing a test here:

https://github.com/elastic/connectors-ruby/blob/a2f99266a9a524ee0f21484944779a4c5930a1d0/spec/core/connector_settings_spec.rb#L14

seanstory:
> Though again, I have no idea how it could have worked, I even see @seanstory did a workaround when writing a test here:

Oops, sorry, this was not intentional workaround.

artem-shelkovnikov:
@seanstory it was a solid and correct one for this situation IMO :D

lib/core/connector_settings.rb
```diff
@@ -21,17 +21,40 @@ class ConnectorSettings
...
+    def self.fetch_by_id(connector_id)
       es_response = ElasticConnectorActions.get_connector(connector_id)
       globals = ElasticConnectorActions.connectors_meta
 
       raise ConnectorNotFoundError.new("Connector with id=#{connector_id} was not found.") unless es_response[:found]
       new(es_response, globals)
     end
 
+    def self.fetch_native(page_size = DEFAULT_PAGE_SIZE)
+      globals = ElasticConnectorActions.connectors_meta
```
vidok, line 38: Just curious why `globals` and not connectors_meta? It confuses me a bit.

lib/core/connector_settings.rb
```diff
@@ -21,17 +21,40 @@ class ConnectorSettings
...
+
+      results = []
+      offset = 0
+      loop do
+        puts "#{page_size}, #{offset}"
```
vidok, line 43: forgotten `puts`

lib/core/connector_settings.rb
wangch079: +1, I think `connectors_meta` or even `meta` is a much better name.

vidok: Also, this variable is not global :)

vidok (approved):
Great job!

PR #305, opened 2022-09-13
[Native Connectors] Heartbeat flow should be separate from sync flow

mchernyavskaya:
## Closes https://github.com/elastic/enterprise-search-team/issues/2716

Currently, the heartbeats don't start until the first sync, and this isn't a user-friendly experience on the UI because the user might sit and wait for a long time before the connector signifies that it's alive.

It's an oversight that should be fixed.

Bonus points: make one timer task per cycle, not one per connector (that would eat up threads doing very little).

## Steps to reproduce the behavior:

* Create an index for the connector
* Start the connector service dispatcher with a native connector and with sync disabled on it
* Heartbeats should start and show connector as available

#### Pre-Review Checklist

### DIFF
--- a/lib/app/dispatcher.rb
+++ b/lib/app/dispatcher.rb
@@ -60,7 +60,7 @@ module App
       end
 
       def start_polling_jobs!
-        scheduler.when_triggered do |connector_settings|
+        scheduler.when_polling_jobs do |connector_settings, do_sync|
           service_type = connector_settings.service_type
           connector_id = connector_settings.id
           index_name = connector_settings.index_name
@@ -82,9 +82,11 @@ module App
 
           pool.post do
             send_heartbeat(connector_id, service_type)
-            Utility::Logger.info("Starting a job for connector (ID: #{connector_id}, service type: #{service_type})...")
-            job_runner = Core::SyncJobRunner.new(connector_settings, service_type)
-            job_runner.execute
+            if do_sync
+              Utility::Logger.info("Starting a job for connector (ID: #{connector_id}, service type: #{service_type})...")
+              job_runner = Core::SyncJobRunner.new(connector_settings, service_type)
+              job_runner.execute
+            end
           rescue StandardError => e
             Utility::ExceptionTracking.log_exception(e, "Job for connector (ID: #{connector_id}, service type: #{service_type}) failed due to unexpected error.")
           end
--- a/lib/core/heartbeat.rb
+++ b/lib/core/heartbeat.rb
@@ -6,7 +6,7 @@
 
 # frozen_string_literal: true
 
-require 'concurrent'
+require 'concurrent-ruby'
 require 'connectors/connector_status'
 require 'connectors/registry'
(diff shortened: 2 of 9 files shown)


mchernyavskaya:
@wangch079 and @artem-shelkovnikov - this is an early edition but I was thinking to propose this as another alternative. The difference between this and what we discussed during the sync:

* It doesn't create a separate config option but reuses the one we already have
* It's not directly dependent on the syncs but it does use the same scheduler's cycle

Advantages: we only query for all the native connectors once and we reuse the `should_sync` flag - there's no extra recalculations of the state.
I can also go with the implementation we wanted but I wanted to throw this into the mix of possibilities first.

lib/app/dispatcher.rb
```diff
@@ -60,7 +60,7 @@ def scheduler
       end
 
       def start_polling_jobs!
-        scheduler.when_triggered do |connector_settings|
+        scheduler.when_polling_jobs do |connector_settings, do_sync|
```
artem-shelkovnikov, line 63: I'd probably rename `when_polling_jobs` to `with_each_polling_interval_tick` or something similar that indicates how often/when this is triggered

artem-shelkovnikov: `do_sync` by the signature looks like a method in inheritance chain, I'm trying to think of a better name here - does `sync_triggered` sound better?

lib/app/dispatcher.rb
```diff
@@ -82,9 +82,11 @@ def start_polling_jobs!
 
           pool.post do
             send_heartbeat(connector_id, service_type)
```
artem-shelkovnikov, line 84: I think it's also worth to separate actual heartbeat from the configuration change - probably for a different PR

lib/app/dispatcher.rb
```diff
@@ -82,9 +82,11 @@ def start_polling_jobs!
...
-            Utility::Logger.info("Starting a job for connector (ID: #{connector_id}, service type: #{service_type})...")
-            job_runner = Core::SyncJobRunner.new(connector_settings, service_type)
-            job_runner.execute
+            if do_sync
```
artem-shelkovnikov, line 85: Do you think`else` with `Logger.debug` could be nice to have here?

lib/core/heartbeat.rb
```diff
@@ -6,7 +6,7 @@
 
 # frozen_string_literal: true
 
-require 'concurrent'
+require 'concurrent-ruby'
```
artem-shelkovnikov, line 9: Probably not a needed change here, or was that a wrong module name all along?

artem-shelkovnikov (commented):
Left a couple comments, but I'll follow up in the Slack threads with my high-level thoughts about this PR and https://github.com/elastic/connectors-ruby/pull/306

mchernyavskaya:
To be superseded by https://github.com/elastic/connectors-ruby/pull/306

PR #306, opened 2022-09-13
Connector service refactor

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/2716

The changes made in this PR:
1. Unify native mode and non-native mode to dispatcher.rb. To run in native mode, set `native_mode = true` in YAML. To run in non-native mode, set `native_mode = false` and configure `connector_id` and `service_type`
2. Refactor `Scheduler` to handle scheduling of 3 tasks: 1. sync, 2. heartbeat, 3. configuration

#### Pre-Review Checklist


wangch079:
It's still in draft state, and I would like to get some early feedback from the team. If this approach looks good, I will start polishing it and adding specs tomorrow. @elastic/ingestion-team

lib/app/dispatcher.rb
```diff
@@ -27,13 +27,17 @@ class << self
       def start!
         running!
         Utility::Logger.info('Starting connector service...')
+        start_heartbeat_task!
```
mchernyavskaya, line 30: I don't see changes in spec files related to this change.

lib/app/dispatcher.rb
```diff
@@ -55,12 +59,21 @@ def pool
         )
       end
 
-      def scheduler
-        @scheduler ||= Core::NativeScheduler.new(POLL_IDLING)
+      def sync_scheduler
+        @sync_scheduler ||= Core::NativeScheduler.new(:sync, POLL_IDLING)
```
mchernyavskaya, line 63: I am not sure I like the direction it's taking. This triples the number of pools, therefore increasing the number of threads and consumed memory.

lib/app/dispatcher.rb
```diff
@@ -55,12 +59,21 @@ def pool
...
+      end
+
+      def heartbeat_scheduler
+        @heartbeat_scheduler ||= Core::NativeScheduler.new(:heartbeat, POLL_IDLING)
+      end
+
+      def configuration_scheduler
+        @configuration_scheduler ||= Core::NativeScheduler.new(:configuration, 5)
```
mchernyavskaya, line 71: Why do we reuse the `POLL_IDLING` for the heartbeat and sync, but hardcode the config scheduler? If the idle intervals should be different, wouldn't it still make sense to make the other one configurable, too?

lib/app/dispatcher.rb
```diff
@@ -93,9 +105,36 @@ def start_polling_jobs!
         Utility::ExceptionTracking.log_exception(e, 'connector service failed due to unexpected error.')
       end
 
-      def send_heartbeat(connector_id, service_type)
-        Utility::Logger.info("Sending a heartbeat on [#{connector_id} - #{service_type}]...")
-        Core::Heartbeat.send(connector_id, service_type)
+      def start_heartbeat_task!
+        Thread.new do
+          Utility::Logger.info('Start heartbeat task...')
```
mchernyavskaya, line 110: I really don't like the code repetition in the `start_xxx_task` methods. I think it's a clear case for refactoring.

lib/app/dispatcher.rb
```diff
@@ -93,9 +105,36 @@ def start_polling_jobs!
         Utility::ExceptionTracking.log_exception(e, 'connector service failed due to unexpected error.')
       end
 
-      def send_heartbeat(connector_id, service_type)
-        Utility::Logger.info("Sending a heartbeat on [#{connector_id} - #{service_type}]...")
-        Core::Heartbeat.send(connector_id, service_type)
+      def start_heartbeat_task!
+        Thread.new do
```
mchernyavskaya, line 109: What's the reasoning behind starting all of it in new threads? The work units are submitted to the thread pools anyway, right? Why make threads to create threads?

lib/app/dispatcher.rb
```diff
@@ -93,9 +105,36 @@ def start_polling_jobs!
...
+          rescue StandardError => e
+            Utility::ExceptionTracking.log_exception(e, 'connector service failed due to unexpected error.')
+          end
+        end
+      end
+
+      def start_configuration_task!
+        Thread.new do
+          Utility::Logger.info('Start configuration task...')
+          configuration_scheduler.when_triggered do |connector_settings|
```
mchernyavskaya, line 129: As far as I can see, the configuration scheduler [can only be triggered when the connector is in status `CREATED`](https://github.com/elastic/connectors-ruby/blob/692375321e20bda9596fa45a5a46e64b22334318/lib/core/scheduler.rb#L129). How many times can this happen in the life of the connector, and is it enough to start a thread per connector that would poll for it every 5 seconds?

lib/core/scheduler.rb
```diff
@@ -27,7 +29,7 @@ def connector_settings
     def when_triggered
       loop do
         connector_settings.each do |cs|
-          if sync_triggered?(cs)
+          if send("#{@type}_triggered?", cs)
```
mchernyavskaya, line 32: This one is going to be a hell to do refactoring. It might be the ruby way to do stuff - so maybe it's actually OK. But it will make anything like `Refactor - Rename` practically impossible.

lib/app/dispatcher.rb
wangch079: Not yet. I just want to get feedback of this approach before I worked on specs.

lib/app/dispatcher.rb
wangch079: Yes, I try to make the tasks isolated.

>This triples the number of pools

It won't triple the number of pools. We will still have one thread pool, which is responsible for task execution of 1) sync job, 2) heartbeat, 3) configuration

Besides the main thread, now we do have 2 more threads to handle heartbeat and configuration (I used `Thread.new` instead of thread pool because they are not meant to be reused). The drawback is we now have 3 threads trying to call `client.indices.search`, but I think this is acceptable. Because the configuration task needs to be run more frequently if we use the same loop it has to run as frequently as the configuration task, overall the number of `client.indices.search` calls will be similar.

lib/app/dispatcher.rb
wangch079: Yes, again, will make them configurable once I get feedback from the team.

lib/app/dispatcher.rb
wangch079: Yes, the real work will be submitted to the pool, e.g. the heartbeat task, and the configuration. But we can't have a job polling loop, a heartbeat loop and a configuration loop running in one thread at the same time. That's why we need to have 3 threads to run the loop.

`Thread.new` is used here instead of thread pool because the threads here are not meant to be reused.

lib/app/dispatcher.rb
wangch079: I don't see it necessary to refactor here. the only part is duplicating is this
```
Utility::Logger.info('Start heartbeat task...')
heartbeat_scheduler.when_triggered do |connector_settings|
  ...
rescue StandardError => e
  Utility::ExceptionTracking.log_exception(e, 'connector service failed due to unexpected error.')
end
```
So it may be rafactored to 
```
def start_task(type)
  Utility::Logger.info("Start #{type} task...")
  send("#{type}_scheduler".to_sym).when_triggered do |connector_settings|
    yield connector_settings
  rescue StandardError => e
    Utility::ExceptionTracking.log_exception(e, "#{type} task failed due to unexpected error.")
  end
end
```
But I don't really think this is necessary.

lib/app/dispatcher.rb
wangch079: >How many times can this happen in the life of the connector
Once, after it's created.

I was trying hard to [get rid of this task in connector service](https://elastic.slack.com/archives/C03GAPTQ2JF/p1661505372210029?thread_ts=1661351497.576319&cid=C03GAPTQ2JF), but this is the solution at least for 8.5

lib/core/scheduler.rb
wangch079: I don't like it either, that's why I need the feedback. I can have 3 different methods `when_sync_triggered`, `when_heartbeat_triggered`, `when_configuration_triggered`.

config/connectors.yml.example
```diff
@@ -20,7 +20,9 @@ thread_pool:
 
 log_level: info
 
-idle_timeout: 60
+poll_interval: 3
+termination_timeout: 60
```
mchernyavskaya, line 24: I don't see where this one is used? And the name doesn't tell me about its intention.

wangch079: It is used here:
https://github.com/elastic/connectors-ruby/blob/bf9a3252cfc63fd1e1f4e0a999f396d388f3de99/lib/app/dispatcher.rb#L19

artem-shelkovnikov (commented):
I see it's still a draft, but it looks like it's going in a very right direction, I like how `App` namespace shrinks!

And it seems like we can easily refactor it to a single main thread + thread per sync job later

lib/app/dispatcher.rb
```diff
@@ -7,31 +7,32 @@
...
   class Dispatcher
-    POLL_IDLING = (App::Config[:idle_timeout] || 60).to_i
-    TERMINATION_TIMEOUT = (App::Config[:termination_timeout] || 60).to_i
-    MIN_THREADS = (App::Config.thread_pool&.min_threads || 0).to_i
-    MAX_THREADS = (App::Config.thread_pool&.max_threads || 5).to_i
-    MAX_QUEUE = (App::Config.thread_pool&.max_queue || 100).to_i
+    POLL_INTERVAL = (App::Config.poll_interval || 3).to_i
+    TERMINATION_TIMEOUT = (App::Config.termination_timeout || 60).to_i
+    HEARTBEAT_INTERVAL = (App::Config.heartbeat_interval || 60 * 30).to_i
+    MIN_THREADS = (App::Config.dig(:thread_pool, :min_threads) || 0).to_i
```
mchernyavskaya, line 21: What'll happen when we have 0 threads? Is this a valid setting at all? If yes, what will happen when something is submitted to this pool?
TLDR: maybe we should default this to 1?

tests/connectors.yml
```diff
@@ -11,4 +11,4 @@ connector_id: "mongo"
 service_type: "mongo"
 log_level: info
 
-idle_timeout: 60
+poll_interval: 60
```
mchernyavskaya, line 14: I think we agreed on 5 seconds, right?

wangch079: I actually don't know what is under the dir `tests`, but I will update it.

lib/app/dispatcher.rb
wangch079: >TLDR: maybe we should default this to 1?

The default will only work if it's not configured in YAML.

I tried to add validation (e.g. when both `min_threads` and `max_threads` are available, `max_threads` must not be less than `min_threads`) but it doesn't work. There are other validations needed, e.g. when it's non-native mode, both `connector_id` and `service_type` must be configured. I have created an issue https://github.com/elastic/enterprise-search-team/issues/2749 to track this. cc @elastic/ingestion-team

lib/core/scheduler.rb
```diff
@@ -78,6 +78,13 @@ def sync_triggered?(connector_settings)
         return false
       end
 
+      # We want to sync when sync never actually happened
+      last_synced = connector_settings[:last_synced]
+      if last_synced.nil? || last_synced.empty?
+        Utility::Logger.info("Connector #{connector_settings.id} has never synced yet, running initial sync.")
+        return true
+      end
```
wangch079, line 86: If it's never synced, we should sync it only if sync is enabled. therefore I move this block after the `scheduling_settings[:enabled]` check.

lib/core/scheduler.rb
```diff
@@ -90,6 +97,7 @@ def sync_triggered?(connector_settings)
         Utility::Cron.quartz_to_crontab(current_schedule)
       rescue StandardError => e
         Utility::ExceptionTracking.log_exception(e, "Unable to convert quartz (#{current_schedule}) to crontab.")
+        return false
```
wangch079, line 100: I think this is a bug. If the quartz can't be parsed correctly, we should not sync.

lib/app/dispatcher.rb
```diff
@@ -56,46 +57,74 @@ def pool
...
-        Utility::ExceptionTracking.log_exception(e, 'connector service failed due to unexpected error.')
+        Utility::ExceptionTracking.log_exception(e, 'The connector service failed due to unexpected error.')
       end
 
-      def send_heartbeat(connector_id, service_type)
-        Utility::Logger.info("Sending a heartbeat on [#{connector_id} - #{service_type}]...")
-        Core::Heartbeat.send(connector_id, service_type)
+      def start_sync_task(connector_settings)
+        # connector-level checks
+        unless Connectors::REGISTRY.registered?(connector_settings.service_type)
```
wangch079, line 86: In non-native mode, `service_type` is populated by the configuration task. If this runs before the configuration task, this will fail because `connector_settings.service_type` should be `nil`, but will eventually pass.

lib/core/scheduler.rb
artem-shelkovnikov: Ha, good catch!

lib/core/scheduler.rb
artem-shelkovnikov: But we do not sync as far as I remember, is it not the case?

lib/app/dispatcher.rb
artem-shelkovnikov: That should be fixed by running everything in 1 thread, right?

lib/app/dispatcher.rb
```diff
@@ -7,31 +7,32 @@
...
-    TERMINATION_TIMEOUT = (App::Config[:termination_timeout] || 60).to_i
-    MIN_THREADS = (App::Config.thread_pool&.min_threads || 0).to_i
-    MAX_THREADS = (App::Config.thread_pool&.max_threads || 5).to_i
-    MAX_QUEUE = (App::Config.thread_pool&.max_queue || 100).to_i
+    POLL_INTERVAL = (App::Config.poll_interval || 3).to_i
+    TERMINATION_TIMEOUT = (App::Config.termination_timeout || 60).to_i
+    HEARTBEAT_INTERVAL = (App::Config.heartbeat_interval || 60 * 30).to_i
+    MIN_THREADS = (App::Config.dig(:thread_pool, :min_threads) || 0).to_i
+    MAX_THREADS = (App::Config.dig(:thread_pool, :max_threads) || 5).to_i
+    MAX_QUEUE = (App::Config.dig(:thread_pool, :max_queue) || 100).to_i
```
artem-shelkovnikov, line 23: What do you think about moving these lines to become private methods, just to show this way that they should not be used outside of `App::Dispatcher`? When they are a constant they give a feeling, that they can be used by anyone who import this class (which is not the case as far as I see)

lib/app/dispatcher.rb
```diff
@@ -56,46 +57,74 @@ def pool
...
+          Utility::Logger.info("The service type (#{connector_settings.service_type}) is not supported. Skipping...")
+          return
+        end
+        unless connector_settings.valid_index_name?
+          Utility::Logger.info("The index name of #{connector_settings.formatted} is invalid. Skipping...")
+          return
+        end
+        Core::ElasticConnectorActions.ensure_content_index_exists(connector_settings.index_name)
+
+        start_heartbeat_task(connector_settings)
```
artem-shelkovnikov, line 96: Hmmm, so one task calls another? Is it needed here?

lib/connectors/crawler/scheduler.rb
```diff
@@ -19,6 +19,7 @@ def connector_settings
         Core::ConnectorSettings.fetch_crawler_connectors || []
       rescue StandardError => e
         Utility::ExceptionTracking.log_exception(e, 'Could not retrieve Crawler connectors due to unexpected error.')
+        []
```
artem-shelkovnikov, line 22: Nice catch!

lib/core/connector_settings.rb
```diff
@@ -7,11 +7,9 @@
 # frozen_string_literal: true
 
 require 'active_support/core_ext/hash/indifferent_access'
-require 'active_support/core_ext/object/blank'
```
artem-shelkovnikov, line 10: I think that's required for `.empty?` that you use on line 106, might be wrong tho

lib/core/connector_settings.rb
```diff
@@ -98,6 +96,20 @@ def run_ml_inference?
...
     end
 
+    def formatted
+      properties = ["ID: #{id}"]
+      properties << "Service type: #{service_type}" if service_type
+      "connector (#{properties.join(', ')})"
+    end
+
+    def needs_service_type?
+      service_type.to_s.strip.empty?
```
artem-shelkovnikov, line 106: Hmmm, why `to_s.strip`? is it possible to be something like " " or not be a string?

lib/core/elastic_connector_actions.rb
```diff
@@ -144,7 +144,6 @@ def fetch_document_ids(index_name)
         result
       end
 
-      # should only be used in CLI
```
artem-shelkovnikov, line 147: I think it's good to leave some comment here about why we don't use it but still keep it, maybe mentioning that index is normally created by Kibana or Enterprise Search.

lib/core/heartbeat.rb
```diff
@@ -6,62 +6,24 @@
...
-                                   Connectors::ConnectorStatus::NEEDS_CONFIGURATION
-                                 end
-
-          doc[:status] = new_connector_status
-
-          Utility::Logger.info("Heartbeat updated configuration for connector #{connector_id}.")
-          Utility::Logger.info("Changing connector status to #{new_connector_status}.")
-        elsif connector_settings.connector_status_allows_sync?
-          connector_instance = Connectors::REGISTRY.connector(service_type, connector_settings.configuration)
+        if connector_settings.connector_status_allows_sync?
```
artem-shelkovnikov, line 21: I think this part can go to Configuration task too, leaving heartbeat to only do this small `last_seen` update.

I think this update of status of connector with heartbeat is a bit confusing and led me to really awkward UI interactions. (Heartbeat thread decided that connector is not connected due to small network lag, but connector was running in the meantime, so connector was marked as ERROR, while it was working actually.)

spec/core/scheduler_spec.rb
```diff
@@ -0,0 +1,167 @@
+require 'core/native_scheduler'
+require 'core/scheduler'
+
+describe Core::Scheduler do
+  subject { Core::NativeScheduler.new(poll_interval, heartbeat_interval) }
```
artem-shelkovnikov, line 5: We can even probably stub this by creating a stub class that inherits from Core::Scheduler and implements needed methods with stubs. Probably not worth to change in this PR, just thinking out loud.

artem-shelkovnikov (approved):
I did not thoroughly check tests, but altogether I think it's a good change, we can get this merge and fix non-breaking issues in different PRs to not create merge conflicts or endless rebases/merges in this PR.

So I'm up to merge if nobody else sees potential bad blockers or problems.

lib/core/scheduler.rb
wangch079: We do not sync, because in the next step, `cron_parser` is `nil`. But I think we shouldn't go to the next step at all.

https://github.com/elastic/connectors-ruby/blob/93403cc4c8a568e4b1c65df7a8e357cad5b05f8a/lib/core/scheduler.rb#L102-L108

lib/app/dispatcher.rb
wangch079: Yes you are right. ignore my comment above

lib/app/dispatcher.rb
wangch079: 👍

lib/app/dispatcher.rb
wangch079: Yes, according to [the connector protocol](https://github.com/elastic/connectors-ruby/blob/main/docs/CONNECTOR_PROTOCOL.md):

>Connectors should update the last_seen field with their current datetime in UTC format (timezone-agnostic) every half hour, and every time they perform a sync job

lib/core/connector_settings.rb
wangch079: I don't see any error when I tested it locally. I think `.empty?` is a method of string in core ruby?

lib/core/connector_settings.rb
wangch079: Yes, to prevent something like `" "`, maybe it's over-defensive?

lib/core/elastic_connector_actions.rb
wangch079: I remove this comment because it's no longer true.

lib/core/heartbeat.rb
wangch079: This basically checks the connection to remote source, so I think it falls well in the scope of heartbeat?

>Heartbeat thread decided that connector is not connected due to small network lag

I would see chances are really small as we have retry enabled in our ES client.

spec/core/scheduler_spec.rb
wangch079: I don't quite understand this. Please do refactor if you think it's necessary.

lib/core/scheduler.rb
artem-shelkovnikov: I don't understand though, next step after what?

lib/app/dispatcher.rb
artem-shelkovnikov: I've seen it in 2 places, that's why I'm asking:
image

I was curious if it's needed in both, I would remove it from the sync task if possible

lib/core/connector_settings.rb
artem-shelkovnikov: You are right, sorry for the noise :)

lib/core/connector_settings.rb
artem-shelkovnikov: I'm on the fence here - I like code that can recover from failure, on the other hand I like it when the code raises problems as fast as possible in the development phase.

I think it's okay to keep it this way here, but also to think how many errors or invalid data we can recover from in the whole repo.

lib/core/elastic_connector_actions.rb
artem-shelkovnikov: It's not used anywhere, right?

lib/core/heartbeat.rb
artem-shelkovnikov: > This basically checks the connection to remote source, so I think it falls well in the scope of heartbeat?

I think it depends on our view here, idk how much customer value it brings in terms of UX. 

I just had a feeling that making heartbeat write-only (no read operation) might be nice, but I think I might be overthinking this.

spec/core/scheduler_spec.rb
artem-shelkovnikov: Sounds good, I'll take a look sometime if I can improve this, but the code here is already good enough :)

lib/app/dispatcher.rb
wangch079: Indeed, `start_heartbeat_task` is called in 2 places, and this is according to the protocol, that heartbeat will be sent 
1. every half an hour
2. every time a sync job runs

lib/core/elastic_connector_actions.rb
wangch079: It's used here:
https://github.com/elastic/connectors-ruby/blob/aac87fd740f52e965480af5936167f33589640ad/lib/app/dispatcher.rb#L94

And in CLI, which I think we need to re-visit later.

lib/core/scheduler.rb
wangch079: Here is the code before this PR:
https://github.com/elastic/connectors-ruby/blob/18082d7e652d89f69f692891f5ff00c5861ea8e4/lib/core/scheduler.rb#L89-L100

Because there's no `return false` after line 92, it goes to line 94 with `current_schedule = nil`, which results `cron_parser = nil`. In the end, this method `returns false` at line 99.

But we should not reach line 94 as we know `current_schedule` is `nil`.

lib/core/heartbeat.rb
wangch079: I don't even think it's necessary to report remote source status in heartbeat. We only care about remote source status when we want to sync data. This is how Workplace Search works.

PR #312, opened 2022-09-15
Fix MongoDB leaking threads if auth is provided

artem-shelkovnikov:
Big mistake by me when adding logic that adds basic auth for mongo in this line:

```
client = client.with(
            user: @user,
            password: @password
          )
```

This line actually creates an new client for this MongoDB instance and overwrites content of `client` variable with it. Old client is left hanging indefinitely and never deallocates the resources, and since new client is actually not an independent client - it's "derivative" client, it seems to also not deallocate the resources any more.

This PR removes usage of `client.with` and just initialize client once and not do fancy stuff with it.

### DIFF
--- a/lib/connectors/mongodb/connector.rb
+++ b/lib/connectors/mongodb/connector.rb
@@ -72,22 +72,20 @@ module Connectors
         end
       end
 
-      def with_client
-        client = Mongo::Client.new(
-          @host,
-          database: @database,
-          app_name: 'Elastic Enterprise Search',
-          max_pool_size: 8,
-          socket_timeout: 60,
-          heartbeat_frequency: 120,
-          connect_timeout: 30,
-          direct_connection: @direct_connection
-        )
-
-        if @user.present? && @password.present?
-          client = client.with(
+      def with_client(&block)
+        client = if @user.present? || @password.present?
+          Mongo::Client.new(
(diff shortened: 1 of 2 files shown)


lib/connectors/mongodb/connector.rb
```diff
@@ -73,23 +73,21 @@ def do_health_check
...
-          direct_connection: @direct_connection
-        )
-
-        if @user.present? && @password.present?
-          client = client.with(
-            user: @user,
-            password: @password
-          )
-        end
+        client = if @user.present? || @password.present?
```
seanstory, line 76:
```suggestion
        client = if @user.present? && @password.present?
```

seanstory: Unsure if this was intentional, but wanted to flag it in case

seanstory (approved):

lib/connectors/mongodb/connector.rb
artem-shelkovnikov: Indeed, it was intentional though slightly unrelated to memory leak - we can still attempt to create a client for mongo when user provided only username or only password, and mongodb will raise an error itself, which we'll show to the user - a bit of a lazy validation of username/password on my side, can improve on that for sure!

PR #314, opened 2022-09-15
Expose human-readable and actionable connector errors through connector servic

mchernyavskaya:
## Closes https://github.com/elastic/enterprise-search-team/issues/2696

See discussion: https://elastic.slack.com/archives/C03GAPTQ2JF/p1662722221419509

## Related issues

* [Corresponding frontend issue](https://github.com/elastic/enterprise-search-team/issues/2730)

## Requirements to error messages proposed by @byronhulcher:

* plaintext english
* no line breaks
* no length limit (still, we need the messages to be reasonably short)

## Example error scenarios that need to produce an actionable error message:

* Healthcheck to 3rd-party service failed
* User provided configuration that does not make much sense (invalid fields, wrong urls)
* User provided invalid auth data

## What the PR addresses

- [x] error messages in exceptions are short enough and align with requirements
- [x] we don't log full json but instead have a trace ID to search in logs
- [x] sync errors are written in the `error` field
- [x] other connector errors are also written into the error field
- [x] connector cleans the `error` field on successful state

#### Pre-Review Checklist

### DIFF
--- a/lib/utility/logger.rb
+++ b/lib/utility/logger.rb
@@ -6,6 +6,7 @@
 
 require 'logger'
 require 'active_support/core_ext/module'
+require 'active_support/core_ext/string/filters'
 require 'ecs_logging/logger'
 
 module Utility
@@ -50,7 +51,7 @@ module Utility
       end
 
       def abbreviated_message(message)
-        message.gsub(/[\n\t]/, ' ').gsub(/\s+/, ' ')[0, MAX_SHORT_MESSAGE_LENGTH].strip
+        message.gsub(/\s+/, ' ').strip.truncate(MAX_SHORT_MESSAGE_LENGTH)
       end
     end
   end


mchernyavskaya:
@artem-shelkovnikov @wangch079 I'd appreciate a review here.

lib/core/elastic_connector_actions.rb
```diff
@@ -98,6 +105,7 @@ def complete_sync(connector_id, job_id, status)
         update_connector_fields(connector_id,
                                 :last_sync_status => sync_status,
                                 :last_sync_error => status[:error],
+                                :error => status[:error],
```
wangch079, line 108: Is `sync_error` supposed to be populated into `error`? It's already been set in `last_sync_error`

lib/core/heartbeat.rb
```diff
@@ -21,6 +21,8 @@ def send(connector_settings)
         if connector_settings.connector_status_allows_sync?
           connector_instance = Connectors::REGISTRY.connector(connector_settings.service_type, connector_settings.configuration)
           doc[:status] = connector_instance.is_healthy? ? Connectors::ConnectorStatus::CONNECTED : Connectors::ConnectorStatus::ERROR
+          message = "Health check for 3d party service failed for connector [#{connector_settings.id}], service type [#{connector_settings.service_type}]. Check the application logs for more information."
+          doc[:error] = doc[:status] == Connectors::ConnectorStatus::ERROR ? message : nil
```
wangch079, line 25:
```suggestion
          if connector_instance.is_healthy?
  doc[:status] = Connectors::ConnectorStatus::CONNECTED
  doc[:error] = nil
else
  doc[:status] = Connectors::ConnectorStatus::ERROR
  doc[:error] = "Health check for 3d party service failed for connector [#{connector_settings.id}], service type [#{connector_settings.service_type}]. Check the application logs for more information."
end
```

lib/core/elastic_connector_actions.rb
artem-shelkovnikov: Yeah I think this one should be `last_sync_error`, though I find having two errors in the index a bit confusing

lib/utility/logger.rb
```diff
@@ -43,6 +44,14 @@ def error_with_backtrace(message: nil, exception: nil, prog_name: nil)
...
         logger.info("\n")
       end
+
+      def generate_trace_id
+        SecureRandom.uuid
+      end
+
+      def abbreviated_message(message)
+        message.gsub(/[\n\t]/, ' ').gsub(/\s+/, ' ')[0, MAX_SHORT_MESSAGE_LENGTH].strip
+      end
```
artem-shelkovnikov, line 54: We can probably use `https://apidock.com/rails/String/truncate` here - but what are gsubs here for? 
I see first one replaces `\n\t` - I think we can remove it from errors that we raise.

lib/core/elastic_connector_actions.rb
mchernyavskaya: It's supposed to be that way, AFAIK. The [UI issue](https://github.com/elastic/enterprise-search-team/issues/2730) clearly states that the error message that will bubble up will come from the `error` field, and that field only. It won't be an aggregation of two.

As far as I understand, the `last_sync_error` is there just for the sync errors, but the `error` is for everything - not **everything else**. (Or at least that's how I understood the [Slack discussion](https://elastic.slack.com/archives/C03GAPTQ2JF/p1662736785742689?thread_ts=1662722221.419509&cid=C03GAPTQ2JF) - WDYT @sphilipse ?)

lib/core/heartbeat.rb
mchernyavskaya: Why would we replace 3 lines with 6?

lib/utility/logger.rb
mchernyavskaya: We don't have `rails` here, do we? As to the gsubs, - we're supposed to clean up line breaks from messages that we bubble up, but it doesn't mean that we need to remove the breaks from the logs - because in the logs, the breaks are for improving the readability. I don't like the idea of just getting rid of all the breaks that make the long text blocks structured.

artem-shelkovnikov: It's available within `active_support/core_ext/string/filters` - https://www.rubydoc.info/gems/activesupport/String:truncate

My point here is that we wanna display user-friendly error messages, that means that we need to actually sort out the errors ourselves and never show messages of errors that we do not control (e.g. errors raised by ruby/used libraries and anything). In this case we won't need to replace `\n\t` and such, our error handling layer will wrap these errors in user-friendly errors while also logging the real ones, but this indeed is something that we can handle in future.

artem-shelkovnikov (approved):
All in all good functional change, I think it's okay to iron out small problems in a separate PR (one that I see is `last_sync_error` vs `error` discussion)

lib/utility/logger.rb
mchernyavskaya: @artem-shelkovnikov got it. Changing the code.

lib/core/heartbeat.rb
wangch079: I proposed the change because I see you checked `connector_instance.is_healthy?` multiple times, directly or indirectly.

But I have no strong objection to keep it.

PR #315, opened 2022-09-15
Display actual first error that happened when applying bulk operations

artem-shelkovnikov:
Problem found: when doing bulk insert/delete operations and an error happened, we always just pick the first operation and display it in the error message, when in reality it can be any operation in the array.

This PR attempts to find the first index/delete operation that failed and use it as an error.

If no error is found, another error message is displayed.

### DIFF
--- a/lib/utility/es_client.rb
+++ b/lib/utility/es_client.rb
@@ -60,7 +60,7 @@ module Utility
         first_error = nil
 
         response['items'].each do |item|
-          [ 'index', 'delete' ].each do |op|
+          ['index', 'delete'].each do |op|
             if item.has_key?(op) && item[op].has_key?('error')
               first_error = item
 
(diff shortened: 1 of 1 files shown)


lib/utility/es_client.rb
```diff
@@ -57,8 +57,23 @@ def bulk(arguments = {})
...
+          ['index', 'delete'].each do |op|
+            if item.has_key?(op) && item[op].has_key?('error')
+              first_error = item
+
+              break
+            end
+          end
+        end
+
+        if first_error
```
tarekziade, line 72: that stops the whole sync. shouldn't we ignore just the doc and report back?

artem-shelkovnikov: I think for now we're okay - ignoring it will cause minor problems with document counters + we don't report these errors well to UI, so it might be confusing for now for the customer.

In future work (resilience) we want to make it better

tarekziade (approved):

PR #316, opened 2022-09-15
Add serialization logic to Mongo Connector

artem-shelkovnikov:
Problem faced:
1. MongoDB ids are serialized as objects (for example, { '$oid': '536268a06d2d7019ba000000' })
2. Decimal128 fields are serialized as objects (for example, { "$numberDecimal": "69.00" } )

To fix this, I'm adding a bit of serialization logic that transforms them to something that Elasticsearch can read.

Some small problems still: some MongoDB data types (like Decimal128) are not really supported and are just converted to strings, I don't believe there's a way to keep the type, but is there even a need?


artem-shelkovnikov:
I will follow up with some tests on top

tarekziade (approved):

PR #318, opened 2022-09-16
catch SINGINT/SIGTERM (to be tested)

tarekziade:
## Closes https://github.com/elastic/connectors-ruby/issues/###

needs tests, also not sure it's the proper spot.

when the service runs a ctrl+C or a kill should call shutdown so the service gracefully stops with exit code 0

### DIFF
--- a/lib/connectors_service.rb
+++ b/lib/connectors_service.rb
@@ -15,6 +15,18 @@ class ConnectorsService
   def self.run!
     Utility::Environment.set_execution_environment(App::Config) do
       App::PreflightCheck.run!
+
+      # set up exits signals to shut down gracefully
+      Signal.trap("INT") {
+        App::Dispatcher.shutdown!
+        exit(0)
+      }
+      Signal.trap("TERM") {
+        App::Dispatcher.shutdown!
+        exit(0)
+      }
+
+      # start the dispatcher
       App::Dispatcher.start!
     rescue App::PreflightCheck::CheckFailure => e
       Utility::Logger.error("Preflight check failed: #{e.message}")


artem-shelkovnikov:
This code should live in `App::` namespace in the entry point IMO

wangch079:
> This code should live in `App::` namespace in the entry point IMO

We now have 2 same entry points:
1. [https://github.com/elastic/connectors-ruby/blob/main/lib/connectors_service.rb](https://github.com/elastic/connectors-ruby/blob/main/lib/connectors_service.rb)
2. [App](https://github.com/elastic/connectors-ruby/blob/main/lib/app/app.rb)

I'm thinking of removing the 2nd one. WDYT?

artem-shelkovnikov:
Hmmmm, I'm not sure honestly, I see you've added the 1. one and it looks like you did so to share this logic in the gem, is it correct?

Cause if it's a gem shared logic, I'd say yes, it's nice to keep the lib one, otherwise it IMO should live in the app.

wangch079:
> I see you've added the 1. one and it looks like you did so to share this logic in the gem, is it correct?

I followed the [Make your own GEM guide](https://guides.rubygems.org/make-your-own-gem/#your-first-gem):
>Code for your package is placed within the **lib** directory. The convention is to have one Ruby file with the same name as your gem, since that gets loaded when **require "hola"** is run.

artem-shelkovnikov:
We do have other packages that follow this pattern, but we just do requires of all classes within the module, for example:

https://github.com/elastic/connectors-ruby/blob/main/lib/connectors_utility.rb#L10-L16

I think if we don't use this code to run connectors elsewhere, we really benefit from keeping it in `App::` namespace (because it uses everything from `App::`, and now there's an unexpected dependency between `App::` and `ConnectorService::` namespaces)

artem-shelkovnikov:
Already don

PR #320, opened 2022-09-19
Remove local_configuration

wangch079:
With the decision made to store credentials in clear text in 8.5, the local configuration is no longer needed.

#### Pre-Review Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Reference:
 
 ## Terminology
 
-- `connector client` - specific light-weight connector implementation, open-code. Connector clients can be built by Elastic or community built.
+- `connector client` - specific light-weight connector implementation, open-code. Connector clients can be built by Elastic or Community.
 - `connector service` - the app that runs the asynchronous loop that calls Elasticsearch on a regular basis to check whether syncs need to happen.
 - `connector framework` - framework and libraries used to customize and build connector clients. Basically, it's what this repository is.
 - `connector packages` - a previous version of the connector clients. Refer to the [8.3 branch](https://github.com/elastic/connectors-ruby/tree/8.3) if you're looking for connector packages. Also, read more about them in the [custom connector packages guide](https://www.elastic.co/guide/en/workplace-search/8.4/custom-connector-package.html).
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -54,7 +54,7 @@ In this section:
 The [connector protocol](docs/CONNECTOR_PROTOCOL.md) is document-based and building a connector is language- and tool-agnostic.
 To create an Elastic connector, you need to implement this protocol. At a high level, you need to create an application (the connector service) that can read and write to a specific document in Elasticsearch that represents the connector. This document "registers" the connector with Kibana, keeps the configuration that is made via the Kibana UI, as well as the scheduling configuration and the state information for the service. You need to update the connector state to keep up with the current status of the connector application and of the third-party data source. You also need to log sync jobs to an additional index, so that the history of synchronization tasks would be available. And of course, you need to write the results of the synchronization to the Elasticsearch document index, created for this connector service.
 
-The procedure to properly implement the protocol _without the connector framework_ is as follows:
+The procedures to properly implement the protocol _without the connector framework_ is as follows:
```
mchernyavskaya, line 57:
```suggestion
The procedure to properly implement the protocol _without the connector framework_ is as follows:
```

mchernyavskaya: If the verb is in single mode, then the subject should be in single mode.

mchernyavskaya (approved):
LGTM apart from one nitpicky comment.

README.md
wangch079: Then it should be changed to
>The procedures to properly implement the protocol _without the connector framework_ are as follows

WDYT?

lib/connectors/mongodb/connector.rb
```diff
@@ -72,6 +74,8 @@ def do_health_check
       end
 
       def with_client
+        raise "Invalid value for 'Direct connection' : #{@direct_connection}." unless %w[true false].include?(@direct_connection.to_s.strip.downcase)
```
wangch079, line 77: I didn't validate in `initialize`. In that case, the job runner will fail early and sync job won't be claimed, therefore it won't be surfaced to Kibana.

lib/utility/errors.rb
```diff
@@ -119,8 +119,8 @@ def to_h
   end
 
   class HealthCheckFailedError < StandardError
-    def initialize
-      super('Health check failed for 3rd-party service.')
+    def initialize(msg = nil)
+      super("Health check failed for 3rd-party service: #{msg}")
```
wangch079, line 123: If `direct_connection` is not a valid bool value, it will fail here first
https://github.com/elastic/connectors-ruby/blob/aac87fd740f52e965480af5936167f33589640ad/lib/core/sync_job_runner.rb#L54

before it reaches 
https://github.com/elastic/connectors-ruby/blob/aac87fd740f52e965480af5936167f33589640ad/lib/core/sync_job_runner.rb#L61

I think this change also makes sense even without this use case. A hardcoded error message `Health check failed for 3rd-party service` is not very helpful and actionable.

artem-shelkovnikov (approved):
Looks good to me!

I haven't read deeply into README.md changes, but on the surface they look okay to me

PR #321, opened 2022-09-20
Add more ECS fields

wangch079:
#### Pre-Review Checklist


seanstory (commented):
Thanks, Chenhui! Have you made sure it will handle errors/exceptions in a similar manner to ent-search?

wangch079:
> Have you made sure it will handle errors/exceptions in a similar manner to ent-search?

Hi @seanstory , could you point me to the error/exception handling code in ent-search? I don't see any.

seanstory:
> > Have you made sure it will handle errors/exceptions in a similar manner to ent-search?
> 
> Hi @seanstory , could you point me to the error/exception handling code in ent-search? I don't see any.

Check out https://github.com/elastic/ent-search/blob/main/lib/swiftype/exception_tracking.rb#L18-L33

wangch079:
> Check out https://github.com/elastic/ent-search/blob/main/lib/swiftype/exception_tracking.rb#L18-L33

I understand what you meant. For errors, stacktrace, we should use the [Error fields](https://www.elastic.co/guide/en/ecs/8.4/ecs-error.html) instead of `message`. I did some investigation, but I feel it's too risky to make the change now, we need to support both ecs-logging and non-ecs-logging. 

I would suggest we leave this out of the scope of this PR. I think `connectors-python` doesn't support this either.

artem-shelkovnikov (approved):
Looks good to me, but I think @seanstory has valid concerns that I'm not in-context enough to address, should we get Sean's approval here too?

mchernyavskaya:
I think we should also utilize the `log_exception` method from `exception_tracking.rb` to add the formatted data. 

https://github.com/elastic/connectors-ruby/blob/main/lib/utility/exception_tracking.rb#L28-L31

I'd also override the `error()` method in our logger to check for the instance class (as @wangch079 does here).

I am not sure we **can** add the correct error data via the `logger.error`, since that method doesn't seem to have the exception instance in the parameters at all (correct me if I am wrong). But we can at least add a message if the method called is the `error()`.

Then the method could be changed to something like:

```ruby
def ecs_additional_fields(error_message = nil)
  result = {
    :labels => { :index_date => Time.now.strftime('%Y.%m.%d') },
    :log => { :logger => logger.progname },
    :service => {
      :type => 'connectors-ruby',
      :version => Settings.version
    },
    :process => {
      :pid => Process.pid,
      :name => $PROGRAM_NAME,
      :thread => Thread.current.object_id
    }
  }
  unless error_message.nil?
    result[:error] = {
      :message => error_message
    }
  end
  result
end
```
The [original issue](https://github.com/elastic/enterprise-search-team/issues/2795) however doesn't mention the error fields, so it might indeed be out of scope. I'd suggest to create another issue to take care of error data in ecs fields. WDYT ?

wangch079:
Hi @mchernyavskaya , I also thought of having the message going to `error.message` when `logger.error()` is called.

But I think it also makes sense to keep it in message:
```
{... "log.level": "ERROR", "message": "Something went wrong", ...}
```

Agree that `log_exception` and `capture_exception` method of `exception_tracking.rb` is the right place to pass stacktrace to `error.stacktrace`. But I feel it's too risky and not that necessary to push this change now.

mchernyavskaya (approved):

seanstory:
> Agree that log_exception and capture_exception method of exception_tracking.rb is the right place to pass stacktrace to error.stacktrace. But I feel it's too risky and not that necessary to push this change now.

I'll leave that decision to y'all. Can you file a follow-up ticket if it isn't covered here? CC @danajuratoni @serenachou, I think getting ECS error handling that mimics ent-search will be a high priority from Support's perspective.

wangch079:
>I'll leave that decision to y'all. Can you file a follow-up ticket if it isn't covered here?

As we already pass 8.5 FF. I add it in this PR, see [here](https://github.com/elastic/connectors-ruby/pull/321/files/09d74a32188a5bf52d8649ad5eab99d743bb10e2..a1a2267a0f45c20d20fec3cac4ca5ecca76fc8fa)

artem-shelkovnikov (approved):
It looks okay to me, as long as @seanstory's concerns are addressed - I'm a bit out of context here though.

seanstory (approved):

wangch079:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.5|[](https://github.com/elastic/connectors-ruby/pull/340)|

### Questions ?

PR #323, opened 2022-09-20
Use yaml to register connectors

wangch079:
Use YAML config instead of code to register connectors.

#### Pre-Review Checklist

### DIFF
--- a/lib/connectors/registry.rb
+++ b/lib/connectors/registry.rb
@@ -4,49 +4,53 @@
 # you may not use this file except in compliance with the Elastic License.
 #
 
+require 'active_support/core_ext/class/subclasses'
+require 'connectors'
+
 module Connectors
-  class Factory
-    attr_reader :connectors
-
-    def initialize
-      @connectors = {}
-    end
-
-    def register(name, klass)
-      @connectors[name] = klass
-    end
-
-    def registered?(name)
-      @connectors.has_key?(name)
-    end
-
-    def connector_class(name)
-      @connectors[name]
-    end
-
-    def connector(name, configuration)
-      klass = connector_class(name)
-      if klass.present?
-        return klass.new(configuration: configuration)
+  class Registry
+    class << self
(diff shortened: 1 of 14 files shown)


lib/connectors/registry.rb
```diff
@@ -4,49 +4,53 @@
 # you may not use this file except in compliance with the Elastic License.
 #
 
+require 'active_support/core_ext/class/subclasses'
+require 'connectors'
+
 module Connectors
-  class Factory
-    attr_reader :connectors
+  class Registry
```
artem-shelkovnikov, line 11: I'm not sure about making it static class - we already have problems with singletons and module methods, I think it would be better to keep this change out of PR and discuss it better.

IMO in the end the best we could do is get rid of `Connectors::REGISTRY` and just use one instance of `Connectors::Factory` that's instantiated by App. It's a class that is not used a lot and it can be trivial to pass it around to the places that use it.

lib/connectors/registry.rb
```diff
@@ -4,49 +4,53 @@
...
+        @connectors ||= load_connectors
       end
-      raise "Connector #{name} is not yet registered. You need to register it before use"
-    end
 
-    def registered_connectors
-      @connectors.keys.sort
+      def load_connectors
+        connectors = {}
+        connectors_in_yaml = App::Config.connectors.dup || []
```
artem-shelkovnikov, line 41: Particularly this is the worrying place for me - reference to `App::` namespace, plus `Utility::Logger` - 3 objects in codebase reference each other in a circle

artem-shelkovnikov (changes requested):

wangch079:
Hi @artem-shelkovnikov , In this case, we can postpone this to post-FF or even 8.6. I just drafted another PR to simply [only register `MongoDB` as native connector](https://github.com/elastic/connectors-ruby/pull/325). Let's push that PR before FF.

PR #325, opened 2022-09-20
Only register MongoDB as native connector

wangch079:
Only `MongoDB` will be registered as native connector in 8.5

There's another PR https://github.com/elastic/connectors-ruby/pull/323, which tries to register connectors in YAML, but there are different opinions from @artem-shelkovnikov and I don't want to push it so close to FF. This PR should be sufficient for 8.5

#### Pre-Review Checklist

### DIFF
--- a/lib/connectors/registry.rb
+++ b/lib/connectors/registry.rb
@@ -39,14 +39,6 @@ module Connectors
 
   REGISTRY = Factory.new
 
-  require_relative './example/connector'
-  REGISTRY.register(Connectors::Example::Connector.service_type, Connectors::Example::Connector)
-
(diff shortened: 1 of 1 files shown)


artem-shelkovnikov:
I'm not sure about it either - I don't honestly think it's a big deal for 8.5 that user can accidentally manually create native Gitlab/Example connector - they'll have to really do it themselves, there's no way to do it in UI.

Shall we postpone this change until 8.6 and do it in a proper way?

My concern is that this change will make usage of on-prem connector harder (if anyone wants to use it, of course)

wangch079:
As suggested by @artem-shelkovnikov , close this PR and discuss the solution post 8.5

PR #327, opened 2022-09-21
Move service type check to xxx_registered? methods

wangch079:
#### Pre-Review Checklist

### DIFF
--- a/lib/connectors/crawler/scheduler.rb
+++ b/lib/connectors/crawler/scheduler.rb
@@ -21,6 +21,12 @@ module Connectors
         Utility::ExceptionTracking.log_exception(e, 'Could not retrieve Crawler connectors due to unexpected error.')
         []
       end
+
+      private
+
+      def connector_registered?(service_type)
+        service_type == 'elastic-crawler'
+      end
     end
   end
 end
--- a/lib/core/scheduler.rb
+++ b/lib/core/scheduler.rb
@@ -59,29 +59,39 @@ module Core
     private
 
     def sync_triggered?(connector_settings)
+      unless Connectors::REGISTRY.registered?(connector_settings.service_type)
+        Utility::Logger.info("The service type (#{connector_settings.service_type}) is not supported.")
+        return false
+      end
+
+      unless connector_settings.valid_index_name?
+        Utility::Logger.info("The index name of #{connector_settings.formatted} is invalid.")
+        return false
+      end
+
       unless connector_settings.connector_status_allows_sync?
-        Utility::Logger.info("Connector #{connector_settings.id} is in status \"#{connector_settings.connector_status}\" and won't sync yet. Connector needs to be in one of the following statuses: #{Connectors::ConnectorStatus::STATUSES_ALLOWING_SYNC} to run.")
+        Utility::Logger.info("#{connector_settings.formatted.capitalize} is in status \"#{connector_settings.connector_status}\" and won't sync yet. Connector needs to be in one of the following statuses: #{Connectors::ConnectorStatus::STATUSES_ALLOWING_SYNC} to run.")
 
         return false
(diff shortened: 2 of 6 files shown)


artem-shelkovnikov (approved):
LGTM, thanks for a quick bugfix!

lib/core/scheduler.rb
```diff
@@ -59,37 +59,47 @@ def shutdown
     private
 
     def sync_triggered?(connector_settings)
+      unless Connectors::REGISTRY.registered?(connector_settings.service_type)
```
vidok, line 62: Crawler uses the same scheduler (see [scheduler.rb](https://github.com/elastic/connectors-ruby/blob/main/lib/connectors/crawler/scheduler.rb)) so this is a breaking change. 

I think Crawler connector should be ignored in ConnectorsSettings level.

vidok (changes requested):

lib/core/scheduler.rb
artem-shelkovnikov: Oh wow, I missed it fully :/ Great catch!

wangch079: Hi @vidok , I'm a bit confused here. Do you expect `connectos-ruby` to run sync job for crawler? I assume the answer is no, because don't see `connctor` class in [crawler](https://github.com/elastic/connectors-ruby/tree/main/lib/connectors/crawler).

If this is true, this PR will make `connectors-ruby` ignore crawler connecters because `crawler` is not one of the registered connectors:

https://github.com/elastic/connectors-ruby/blob/afb58a6ee3e39de6cbe1bbf655182ea57faddeeb/lib/connectors/registry.rb#L42-L51

wangch079: Oh, I got it. This is used as a utility gem in ent-search. Great catch @vidok

wangch079: Hi @vidok , will this work? https://github.com/elastic/connectors-ruby/pull/327/commits/e61f78be3ad5aaaed91b006c809abd8a327ad543

vidok: Yes, I think this should work, however it requires some testing.

lib/connectors/crawler/scheduler.rb
```diff
@@ -21,6 +21,12 @@ def connector_settings
         Utility::ExceptionTracking.log_exception(e, 'Could not retrieve Crawler connectors due to unexpected error.')
         []
       end
+
+      private
+
+      def connector_registered?(service_type)
```
vidok, line 27: I can't see where this method is used.

vidok (approved):

lib/connectors/crawler/scheduler.rb
wangch079: It's used in `sync_triggered?`:
https://github.com/elastic/connectors-ruby/blob/3d172ac61b04e6c78d6dd6d515037ad14f883bbb/lib/core/scheduler.rb#L62

`heartbeat_triggered?`:
https://github.com/elastic/connectors-ruby/blob/3d172ac61b04e6c78d6dd6d515037ad14f883bbb/lib/core/scheduler.rb#L129

and `configuration_triggered?`:
https://github.com/elastic/connectors-ruby/blob/3d172ac61b04e6c78d6dd6d515037ad14f883bbb/lib/core/scheduler.rb#L144

It override the method in `Core:Scheduler`:
https://github.com/elastic/connectors-ruby/blob/3d172ac61b04e6c78d6dd6d515037ad14f883bbb/lib/core/scheduler.rb#L149-L156

So for crawler, this method returns `true` as long as it's a crawler connector.

vidok: Got it, thank you!

PR #330, opened 2022-09-21
Drop nokogiri

artem-shelkovnikov:
Dropping a gem that's not used


artem-shelkovnikov:
Not important for 8.5 release, but just a quality of life change

wangch079 (approved):

PR #332, opened 2022-09-21
Cast direct_connection to bool, defaulting empty to false this way

artem-shelkovnikov:
Adding logic to cast `direct_connection` option for MongoDB to boolean.

It's done to unblock testing (we haven't exposed options in UI for Native connectors). 

It also might be a good change to keep in future!

### DIFF
--- a/lib/connectors/mongodb/connector.rb
+++ b/lib/connectors/mongodb/connector.rb
@@ -52,7 +52,7 @@ module Connectors
         @collection = configuration.dig(:collection, :value)
         @user = configuration.dig(:user, :value)
         @password = configuration.dig(:password, :value)
-        @direct_connection = configuration.dig(:direct_connection, :value)
+        @direct_connection = !!configuration.dig(:direct_connection, :value)
       end
 
       def yield_documents


artem-shelkovnikov:
Also see thread:
https://elastic.slack.com/archives/C03GAPTQ2JF/p1663781130986719

lib/connectors/mongodb/connector.rb
```diff
@@ -52,7 +52,7 @@ def initialize(configuration: {})
         @collection = configuration.dig(:collection, :value)
         @user = configuration.dig(:user, :value)
         @password = configuration.dig(:password, :value)
-        @direct_connection = configuration.dig(:direct_connection, :value)
+        @direct_connection = !!configuration.dig(:direct_connection, :value)
```
wangch079, line 55: `!!"foobar"` will also return `true`

RossAllenBell: FYI here is how we do it in ent-search, which is pretty comprehensive: https://github.com/elastic/ent-search/blob/e4768dd6bbb17ddc7e75c272dd35285a79cf41df/config/initializers/string.rb#L80

artem-shelkovnikov: @mchernyavskaya fyi!

artem-shelkovnikov:
Closing in favour of https://github.com/elastic/connectors-ruby/pull/337

lib/connectors/mongodb/connector.rb
mchernyavskaya: I did it in a similar way here:
https://github.com/elastic/connectors-ruby/pull/337
But this way is a bit more permissive - it doesn't check for the correct false value. I think it's OK for us though. We might want to be a bit lax with these settings.

I also wasn't sure it was the best practice to monkey-patch string. But we can do it in connectors too if need be.

PR #335, opened 2022-09-22
Add dependencies in gemspec

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/2753

Read dependencies from Gemfile and add non-test non-development dependencies in gemspec.

#### Pre-Review Checklist


artem-shelkovnikov (approved):
Looks good, we can switch to manual process if needed later, but so far that looks like a good way ahead.

seanstory:
This is clever! 

CC @RossAllenBell and @vidok , shout if there's anything really wrong/non-idiomatic with this pattern. I think I've seen before that dependencies should be explicitly in the gemspec, and that bundler can read them from the gemspec? But y'all probably know best if it's ok to have them in both like this.

wangch079:
Requested review from @RossAllenBell and @vidok before I merge this.

RossAllenBell (approved):
I'm not familiar with this approach. I think the risk is that the `connectors_service` gem ends up requiring a broader set of gem dependencies than it actually needs when included in another project. I would assume there exists a non-test/dev gem in `Gemfile` that is not necessary when using the `connectors_service` gem.

That's probably ok for now.

artem-shelkovnikov:
Agreed to this concern, @wangch079 is there a way to verify which dependencies are declared this way?

What do you say if we instead provide a script that generates lines to add to gemspec file so that we could vet the dependencies manually?

wangch079:
> is there a way to verify which dependencies are declared this way?

Yes, manually. I ran `make build_service`, examined the meta file in the generated gem, and I can see the dependencies added.

> I would assume there exists a non-test/dev gem in `Gemfile` that is not necessary when using the `connectors_service` gem.

This gemspec is made to work for `connectors-ruby`, you see I excluded dependencies in `:test` and `:development`:
https://github.com/elastic/connectors-ruby/blob/48d7f43e8b5aece87de060e1c7fd7c413939849d/connectors_service.gemspec#L16

In the [Gemfile](https://github.com/elastic/connectors-ruby/blob/main/Gemfile), we actually just have a group of `:test`, I added `development` for future-proof. If we introduce another group (say `:foobar`), which we don't want to package into  the gem, we need to update the gemspec to exclude `:foobar` as well.

PR #336, opened 2022-09-22
Update Pull Request Template to include a manual gem testing step

artem-shelkovnikov:
Based on recent issues with introducing breaking changes to `ent-search`, I'm proposing to return the checklist item to check that gems are tested against Enterprise Search

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -16,7 +16,7 @@ items that may help during the review.-->
 #### Pre-Review Checklist
 - [ ] Covered the changes with automated tests
 - [ ] Tested the changes locally
-- [ ] (Elastic Only) Built gems (both `connectors_utility` and `connectors_service`) and tested that Enterprise Search works well with new gem versions
+- [ ] (Elastic Only) Built gems (both `connectors_utility` and `connectors_service`) and included into Enterprise Search and tested that Enterprise Search works well with new gem versions.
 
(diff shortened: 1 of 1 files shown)


seanstory (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,6 +16,7 @@ items that may help during the review.-->
 #### Pre-Review Checklist
 - [ ] Covered the changes with automated tests
 - [ ] Tested the changes locally
+- [ ] (Elastic Only) Built gems and tested that Enterprise Search works well with new gem versions
```
wangch079, line 19:
```suggestion
- [ ] (Elastic Only) Built gems (`make build_utility`) and tested that Enterprise Search works well with new gem versions
```

We now build 2 gems:
1. make build_utility
2. make build_service

In 8.6, I hope we will only have one gem building process, and this checkbox can be removed

artem-shelkovnikov: Thinking of that, we do need both, right?

wangch079: Assuming crawler code will be fully migrated to `connectors-ruby` in 8.6, where else do we still need the utility gem?

wangch079: `connectors_service` gem includes executables, which will start the app (i.e. start the loop to poll jobs)

PR #338, opened 2022-09-22
Fix the remapping confusion of documents id caused by string vs symbol

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/2915

Fixes a problem with id incorrectly generated when bulk inserting entries into Elasticsearch.

At some point MongoDB connector started to return `'id'` field instead of `:id` for the documents it returns.

Thinking about that, we should get rid of symbols and always enforce documents to return data in hashes with strings as keys.

#### Pre-Review Checklist


lib/core/output_sink/es_sink.rb
```diff
@@ -26,7 +26,7 @@ def initialize(index_name, request_pipeline, flush_threshold = 50)
     def ingest(document)
       return if document.blank?
 
-      @operation_queue << { :index => { :_index => index_name, :_id => document[:id], :data => document } }
+      @operation_queue << { :index => { :_index => index_name, :_id => document['id'], :data => document } }
```
wangch079, line 29: Here it still uses `:id`:

https://github.com/elastic/connectors-ruby/blob/ba2b87a3676dee416da48bbdec55cc8f06d03bd9/lib/core/sync_job_runner.rb#L63-L64

It depends on connector implementation in what's inside of `document`, I think it's reasonable to add some validation in the framework level (i.e. connector service)

tarekziade:
> Thinking about that, we should get rid of symbols and always enforce documents to return data in hashes with strings as keys.

Allowing implicit symbols or string lookups using that active directory utility sounds like a recipe for disaster to me. I got bit by this several times. I would be in favor of using strings everywhere, always -- and be explicit about it

artem-shelkovnikov:
Talking to @vidok he also mentioned that symbols produce memory overhead that is never taken back as another reason to not use symbols. I think we should fully switch to strings in the ingestion or any parts that operate on data that we have no control over + stop using `.with_indifferent_access`

lib/core/output_sink/es_sink.rb
artem-shelkovnikov: Great catch, gonna fix it here too!

seanstory:
> Talking to @vidok he also mentioned that symbols produce memory overhead that is never taken back as another reason to not use symbols. I think we should fully switch to strings in the ingestion or any parts that operate on data that we have no control over + stop using `.with_indifferent_access`

TIL. @vidok can you bring this up in the next architecture meeting? I've always avoided strings since I feel they're more prone to typos. ent-search makes gratuitous use of `.with_indifferent_access`, so I'm surprised to hear that it's that impactful.

lib/core/output_sink/es_sink.rb
mchernyavskaya: I *think* with the fix I put in it's not needed to fix it in that place anymore. I've checked locally and don't see any dupes now.

mchernyavskaya (approved):
LGTM

mchernyavskaya:
I've tested this branch with @artem-shelkovnikov 's changes and it wasn't working for me. It was failing with the following error on sync:

```
Failed to index documents into Elasticsearch with an error '{"index":{"_index":"search-test","_id":null,"status":400,"error":{"type":"illegal_argument_exception","reason":"unexpected metadata [_id:10228731] in source"}}}'. 
```

I've fixed it the way I saw fit and pushed my changes to this branch, and *then* I have read this whole conversation. =)
It looks to me now that the way I fixed it (by using symbols again) is not the way we want it to be. However, I don't feel good with taking the opposite approach and switching to all-string hashes. Same as Sean, I always preferred symbols as they felt way less error-prone, and @vidok 's POV makes me wonder. But while it may be a valid concern, it requires a lot of code changes, since we're using [with_indifferent_access](https://github.com/elastic/connectors-ruby/search?q=with_indifferent_access) and [to_sym](https://github.com/elastic/connectors-ruby/search?q=to_sym) in more places in code than this current fix changes, and it's more difficult to locate all the places where we rely on those symbols being present.

I have already approved this PR but am now hesitant to merge. What do you think @seanstory @artem-shelkovnikov @vidok ?

vidok:
> > Talking to @vidok he also mentioned that symbols produce memory overhead that is never taken back as another reason to not use symbols. I think we should fully switch to strings in the ingestion or any parts that operate on data that we have no control over + stop using `.with_indifferent_access`
> 
> TIL. @vidok can you bring this up in the next architecture meeting? I've always avoided strings since I feel they're more prone to typos. ent-search makes gratuitous use of `.with_indifferent_access`, so I'm surprised to hear that it's that impactful.

Sure, I will try to come up with a good example.

mchernyavskaya:
@vidok just pushed another change - do you think it's better now?

vidok (approved):

PR #343, opened 2022-09-23
added the dry collection

tarekziade:
## Closes https://github.com/elastic/connectors-ruby/issues/###

makes sure gem install works

#### Pre-Review Checklist


artem-shelkovnikov (approved):
Looks good, but if it's a bad practice pls revert this and let's make it better.

artem-shelkovnikov:
@tarekziade I've added a comment before the gemfile section to remove the gems when `config` gem is removed

tarekziade:
> Looks good, but if it's a bad practice pls revert this and let's make it better.

I don't know, is it?

We want to be able to cur snapshot releases on main and 8.5 branches, and have that used in ent-search.
If you have a process in place that pushes the gem in a public repo, I can change ent-search so it pulls it instead of vendoring it there.

PR #349, opened 2022-09-26
Change log level to reduce noise in the logs

vidok:
## Closes https://github.com/elastic/enterprise-search-team/issues/2874

It's too noisy in the crawler logs (most probably in connectors service as well) so PR will change log level for a few calls

#### Pre-Review Checklist
- [ ] (Elastic Only) Built gems (both `connectors_utility` and `connectors_service`) and included into Enterprise Search and tested that Enterprise Search works well with new gem versions.

### DIFF
--- a/lib/core/scheduler.rb
+++ b/lib/core/scheduler.rb
@@ -45,7 +45,7 @@ module Core
         Utility::ExceptionTracking.log_exception(e, 'Sync failed due to unexpected error.')
       ensure
         if @poll_interval > 0 && !@is_shutting_down
-          Utility::Logger.info("Sleeping for #{@poll_interval} seconds in #{self.class}.")
+          Utility::Logger.debug("Sleeping for #{@poll_interval} seconds in #{self.class}.")
           sleep(@poll_interval)
         end
       end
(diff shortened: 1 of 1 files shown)


lib/core/scheduler.rb
```diff
@@ -74,14 +74,14 @@ def sync_triggered?(connector_settings)
 
       # Sync when sync_now flag is true for the connector
       if connector_settings[:sync_now] == true
-        Utility::Logger.info("#{connector_settings.formatted.capitalize} is manually triggered to sync now.")
+        Utility::Logger.debug("#{connector_settings.formatted.capitalize} is manually triggered to sync now.")
```
seanstory, line 77: Does this print out a lot? I would think this might still be valuable as `info` if it's only logged once when something is manually triggered.

seanstory (approved):

seanstory:
Thanks for the quick turnaround!

lib/core/scheduler.rb
vidok: Not for Crawler. Agree with your point

tarekziade:
For the second part (testing in ent-search) see https://github.com/elastic/ent-search/pull/6908

PR #357, opened 2022-10-04
Multiple concurrent syncs hanging

mchernyavskaya:
# Closes https://github.com/elastic/enterprise-search-team/issues/2913

Disallowing multiple syncs to prevent them hanging.
I am not sure if this fix can break crawler jobs - needs discussion.

#### Pre-Review Checklist
- [ ] (Elastic Only) Built gems (both `connectors_utility` and `connectors_service`) and included into Enterprise Search and tested that Enterprise Search works well with new gem versions.


lib/core/elastic_connector_actions.rb
```diff
@@ -88,6 +88,38 @@ def claim_job(connector_id)
         job['_id']
       end
 
+      def find_current_job(connector_id)
```
tarekziade, line 91: I like that it's calling the server for this, it's future-proof for multi-nodes environments.
One thing I wonder is about race conditions on the index.
you probably want to check the seq_num here to make sure no one updated the doc in the interim

https://www.elastic.co/guide/en/elasticsearch/reference/current/optimistic-concurrency-control.html

tarekziade: In practice it means that when you start a new sync job, you pass along `_seq_no` from the old state you had, and if you get a rejection you know someone else started a new sync job already, and you back off

lib/core/elastic_connector_actions.rb
```diff
@@ -88,6 +88,38 @@ def claim_job(connector_id)
...
+                ]
+              }
+            },
+            :sort => [
+              'created_at' => { :order => 'desc' }
+            ]
+          }
+        )
+        hits = response['hits']['hits']
+        hits.empty? ? nil : hits[0].with_indifferent_access
```
seanstory, line 120: do we need an `ensure` block to protect against any non-200 responses? What about some null checks in case `[hits][hits]` isn't the structure of the response?

lib/core/scheduler.rb
```diff
@@ -29,7 +30,11 @@ def when_triggered
       loop do
         connector_settings.each do |cs|
           if sync_triggered?(cs)
-            yield cs, :sync
+            if ElasticConnectorActions.find_current_job(cs.id).nil?
+              yield cs, :sync
+            else
+              Utility::Logger.debug("Sync job for #{cs.formatted} is already running, skipping...")
```
seanstory, line 36: would it skip this whole window, or just skip this loop?
Like, let's say jobs should sync every 60 minutes, and a sync takes 61 minutes to finish. Does the next sync start at 62 minutes, or at 120 minutes or at 121 minutes? And does the python framework have the same behavior?

lib/core/elastic_connector_actions.rb
mchernyavskaya: I've added the null checks now. As to the non-200, I am ignoring the 404 right now (treat it like an empty set). I would assume that other error codes should fail the whole method. Which ones do you think we need to catch?

lib/core/scheduler.rb
mchernyavskaya: As to the python one, it's difficult to say. I am not great with py, and as I was looking at the codebase, I couldn't locate any code that would check for the job existence. 🤦 Perhaps I'm missing something because @tarekziade definitely mentioned that this check was there...

lib/core/elastic_connector_actions.rb
seanstory: I'm totally fine with them failing the method. I'm mostly worried that if you get a:
```
{
  "errors": ["this is a very important error message"]
}
```

response that we'll instead throw an NPE (you've fixed that) or just give an empty result, instead of actually acknowledging that the response indicated an error.

Typically I wouldn't nitpick like this, but since it's so late in the dev cycle, I think we should either wait on this PR for another release or have a REALLY high bar for ensuring it doesn't add any risk.

lib/core/scheduler.rb
seanstory: 👍 I agree that this is probably the best option

lib/core/elastic_connector_actions.rb
mchernyavskaya: @seanstory I suspect that the [ship might have sailed on 8.5 anyway](https://github.com/elastic/enterprise-search-team/issues/2913#issuecomment-1266889446), but I will try to make it more bullet-proof in case it has not. =)

lib/core/elastic_connector_actions.rb
mchernyavskaya: Done.

lib/core/elastic_connector_actions.rb
```diff
@@ -72,10 +85,30 @@ def set_configurable_field(connector_id, field_name, label, value)
...
+          :ignore => 404,
+          :refresh => true
+        ).tap do |response|
+          seq_no = response['_seq_no']
+          primary_term = response['_primary_term']
+          sync_in_progress = response.dig('_source', 'last_sync_status') == Connectors::SyncStatus::IN_PROGRESS
+        end
+        if sync_in_progress
+          raise JobAlreadyRunningError.new(connector_id)
+        end
```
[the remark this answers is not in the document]
mchernyavskaya: I think we should create a bug for this case.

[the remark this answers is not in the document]
seanstory: I think @wangch079 is addressing these types of scenarios in his current project. I wouldn't worry too much about trying to make an issue or fix it as a one off, since it'll be covered holistically.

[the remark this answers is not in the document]
danajuratoni: > I think we should create a bug for this case.

@mchernyavskaya  Is there a followup bug for this?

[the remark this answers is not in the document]
mchernyavskaya: @danajuratoni based on @seanstory 's comment, I didn't think that I should create one.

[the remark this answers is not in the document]
mchernyavskaya: @wangch079 actually, a question for you. Can your current work help prevent the situation when the connector that's died or killed forcefully mid-sync will never be able to sync again?

PR #364, opened 2022-10-10
Add an elasticsearch client bulk patch to reduce memory usage

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/3039

Optimisation change - as written in the comment in the PR, Elasticsearch client by default clones the arguments that are passed into it before processing the data.

This PR stops cloning the arguments to reduce the memory usage of bulk function. For example if the size of bulk request is 50MB, then call to `.bulk` will make it 100MB memory until bulk request finishes. Therefore, we don't use the original method, instead the original method is copied into `patched_bulk` method and `.clone` usage is deleted.

#### Pre-Review Checklist

## Release Note

Decrease the amount of memory used by the connector service when sending bulk requests to Elasticsearch.

### DIFF
--- a/lib/utility/es_client.rb
+++ b/lib/utility/es_client.rb
@@ -8,6 +8,10 @@
 
 require 'logger'
 require 'elasticsearch'
+require "elasticsearch/api"
+require "elasticsearch/api/namespace/common"
+require "elasticsearch/api/utils"
+require 'elasticsearch/api/response'
 
 module Utility
   class EsClient < ::Elasticsearch::Client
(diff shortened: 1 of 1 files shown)


wangch079 (commented):
Probably we should consult the client team before we proceed with the change? I think `bulk` is the [recommended method](https://www.elastic.co/guide/en/elasticsearch/client/ruby-api/current/examples.html#ex-bulk) to call and I also think `arguments.clone` is a good practice.

seanstory:
> Probably we should consult the client team before we proceed with the change? I think `bulk` is the [recommended method](https://www.elastic.co/guide/en/elasticsearch/client/ruby-api/current/examples.html#ex-bulk) to call and I also think `arguments.clone` is a good practice.

++. It makes good sense that we'd want to do a `.clone` so that a call to `patched_bulk` didn't side-effect the `arguments` with all those `.delete` operations. Instead, could we just halve our max batch mb size? Or we could file an issue with the ruby client to not use `.delete` but some non-destructive operations?

artem-shelkovnikov:
> arguments.clone is a good practice.

It is, however if it generates overhead and this overhead is huge. For lightweight connector that tries to send 50MB of data to Elasticsearch, it'll generate 50MB more. What if it's a native connectors deployment and several connectors are syncing? Then this overhead keeps growing.

The `arguments.clone` is there to have a sort of thread-safety in case someone decides to modify the `arguments` outside of the caller, however for us it's not the case now - we just generate `arguments`, pass it into `bulk` in a synchronous manner, and then discard the whole object and its items after the docs were ingested. I think in this context it's a good and simple optimization, though I'll add a comment on top of the file to also mention it and warn everyone who attempts to change the code around it.

@wangch079 @seanstory I've spoken to clients team about it here: https://elastic.slack.com/archives/C0D8Y4ZJ9/p1666602665401769

Our current course of action is:
1. We patch the code on our side for now to optimise the memory consumption for now
2. I contribute to Elasticsearch ruby client to add an option to not clone the `arguments` hash if an option is passed to the method
3. After 2. is merged and we use the client, clean up the code on our side and do additional testing to verify there's no memory bloat.

wangch079 (approved):

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.5|[](https://github.com/elastic/connectors-ruby/pull/398)|

### Questions ?

PR #367, opened 2022-10-11
Connector service documentation

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/2688

Update README docs.


docs/RELEASING.md
```diff
@@ -52,4 +54,4 @@ merged into the next PATCH release. New features are always done in Developer pr
 ## Developer preview releases
```
wangch079, line 54: @elastic/ingestion-team , I'm not very sure about the Developer preview? Do we really want to **release** it?

README.md
```diff
@@ -1,307 +1,29 @@
...
 
 > Note: The connector framework is a tech preview feature. Tech preview features are subject to change and are not covered by the support SLA of general release (GA) features. Elastic plans to promote this feature to GA in a future release.
 
 Before getting started, review important information about this feature:
 
-- [Terminology](#terminology)
-- [Getting help and providing feedback](#getting-help-and-providing-feedback)
+- [Terminology](docs/SUPPORT.md)
+- [Getting help and providing feedback](docs/SUPPORT.md)
+- [Understand the connector protocol](link to elastic docs)
```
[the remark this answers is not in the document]
chriscressman: Hey @wangch079 . I'm not sure of the timeline to move that content into the Elastic docs, since I'm behind on native connector docs. I suggest we continue to link to the connector protocol document within the repository for now.

Later, we can create a new page in the Elastic docs and update any links and documents in the Ruby and Python repositories.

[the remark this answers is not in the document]
wangch079: Noted, I will work on the connector protocol and let both Ruby and Python link to it. Thanks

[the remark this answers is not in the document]
chriscressman: Perfect, thank you!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -43,4 +43,3 @@ author(s) can easily look it up.-->
 ## For Elastic Internal Use Only
 - [ ] Considered corresponding documentation changes to [contribute separately](https://github.com/elastic/enterprise-search-pubs#contribute-docs-changes-for-product-changes)
 - [ ] New configuration settings added in this PR follow the [official guidelines](https://github.com/elastic/ent-search/blob/main/doc/enterprise-search-config.md)
-- [ ] [Manual test steps](https://github.com/elastic/connectors-ruby/blob/main/docs/INTERNAL.md#minimal-manual-tests) are successful
```
seanstory, line 46: Why is this going away? Not doing these sort of manual test steps is what recently caused a scramble at FF when Crawler needed to update its usage of the connectors-utility gem, but couldn't update because there'd been some breaking changes left undiscovered.

README.md
```diff
@@ -1,307 +1,29 @@
-# Elastic Enterprise Search connectors
+# Elastic connectors
 
-![logo](logo-enterprise-search.png)
```
seanstory, line 3: Why lose the logo? I love having the branding.

README.md
```diff
@@ -1,307 +1,29 @@
...
 
 **The connector will be operated by an administrative user from within Kibana.**
 
 > Note: The connector framework is a tech preview feature. Tech preview features are subject to change and are not covered by the support SLA of general release (GA) features. Elastic plans to promote this feature to GA in a future release.
 
 Before getting started, review important information about this feature:
 
-- [Terminology](#terminology)
-- [Getting help and providing feedback](#getting-help-and-providing-feedback)
+- [Terminology](docs/SUPPORT.md)
```
seanstory, line 11:
```suggestion
- [Terminology](docs/TERMINOLOGY.md)
```

README.md
```diff
@@ -1,307 +1,29 @@
...
+- [Understand the connector protocol](link to elastic docs)
 
-Build, deploy, and operate your connector:
+Build, deploy, and operate your own connector:
 
-- [Building a connector](#building-a-connector)
-- [Operating the connector](#operating-the-connector)
-- [Moving to production](#moving-to-production)
+- [Building/Deploying a connector](docs/DEVELOPING.md)
+- [Operating the connector](https://www.elastic.co/guide/en/enterprise-search/current/connectors.html)
```
seanstory, line 18:
```suggestion
- [Operating a connector](https://www.elastic.co/guide/en/enterprise-search/current/connectors.html)
```
For consistency with the previous line

docs/CODE_OF_CONDUCT.md
```diff
@@ -1,3 +0,0 @@
-303 See Other
-
-Location: https://www.elastic.co/community/codeofconduct
```
seanstory, line 3: This allows github.com to link directly to our Elastic code of conduct from its UI. We copied this from how other Elastic repos are set up. Can you add this file back?

README.md
```diff
@@ -1,307 +1,29 @@
...
-We welcome contributors to the project. Before you begin, please read the [Connectors Contributor's Guide](docs/CONTRIBUTING.md).
+- [Contribute to the repository](docs/CONTRIBUTING.md)
 
 ## Other guides
 
 - [Code of Conduct](https://www.elastic.co/community/codeofconduct)
-- [Getting Support](docs/SUPPORT.md)
+- [Configuration](docs/CONFIG.md)
 - [Releasing](docs/RELEASING.md)
-- [Developer guide](docs/DEVELOPING.md)
```
seanstory, line 304: Why was this link removed?

EDIT: I see, because it's linked above? I kinda like the idea of us having an index here that links to all our other supporting guides, even if they're mentioned elsewhere. But this is a loosely held opinion.

docs/INTERNAL.md
```diff
@@ -1,45 +0,0 @@
-# Elastic Internal Documentation
```
seanstory, line 1: Are we sure we don't want this anymore? These kinds of guides can be useful for new folks when they join.

docs/RELEASING.md
seanstory: Right now, we don't have a better way to have ent-search depend on SNAPSHOT versions of connectors code. It would be awesome if we could use a github URL for our `ent-search/Gemfile` to depend on connectors, but we've had issues with that approach and the need to have a valid connectors.yml file.

docs/TERMINOLOGY.md
```diff
@@ -0,0 +1,10 @@
+# Terminology
+
+- `connector client` - specific light-weight connector implementation, open-code. Connector clients can be built by Elastic or Community.
+- `native connector` - a connector client built and supported by Elastic, made available by default on Cloud.
```
seanstory, line 4:
```suggestion
- `native connector` - a connector client built and supported by Elastic, made available by default on Elastic Cloud.
```

docs/RELEASING.md
wangch079: Agree that we need the gem in-between release. What I meant here is do we really need to release it to https://rubygems.org/?

.github/PULL_REQUEST_TEMPLATE.md
wangch079: `INTERNAL.md` is gone, and #minimal-manual-tests in `INTERNAL.md` is TBD.

I actually don't see other manual tests necessary besides
1. `make lint test`, which is covered by CI
2. https://github.com/elastic/connectors-ruby/blob/d95b7003caff6c34543ca64214ec4ca04a093426/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L17

>Not doing these sort of manual test steps is what recently caused a scramble at FF when Crawler needed to update its usage of the connectors-utility gem, but couldn't update because there'd been some breaking changes left undiscovered.

Oh, you meant to test the utility gem in ent-search. I can add it here, but I expect this will eventually be gone when crawler code is migrated to this repo completely.

README.md
wangch079: Although it still requires Enterprise Search to run connectors, and it's still documented under the [Enterprise Search guide](https://www.elastic.co/guide/en/enterprise-search/current/connectors.html) (But it's named as `Elastic connectors`), I'm not sure we still want to make it under Enterprise Search. After all, we want to move Enterprise Search out of the critical execution path.

docs/INTERNAL.md
wangch079: I can add it back.

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -191,17 +140,80 @@ In addition to the .elastic-connectors index for documents, we have an index to
     "worker_hostname" : { "type" : "keyword" },
     "indexed_document_count" : { "type" : "integer" },
     "deleted_document_count" : { "type" : "integer" },
-    "index_name": { "type": "keyword" },
```
sphilipse, line 194: Why is index_name removed here? This was added because we don't want to cross-reference the connectors index to get this data, because that can be out of date. That's the same reason the connector document should be added (even though we never actually got around to implementing that).

If connectors aren't currently setting this data, they should.

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -165,23 +116,21 @@ This is our main communication index, used to communicate the connector's config
 }
 ```
 
-### Jobs index
-In addition to the .elastic-connectors index for documents, we have an index to log all jobs run by connectors. This is the .elastic-connectors-sync-jobs index. Each JSON document will have the following structure:
+#### Connector job index `.elastic-connectors-sync-jobs`
+In addition to the connector index `.elastic-connectors`, we have an additional index to log all jobs run by connectors. This is the `.elastic-connectors-sync-jobs` index. Each JSON document will have the following structure:
 ```
 {
-  connector: object; -> Copy of the current connector document
```
sphilipse, line 172: Why is this connector object removed? I know that current connectors aren't generally writing that connector object, but this is something they should start doing.

wangch079: replied in another comment.

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -22,120 +21,72 @@ At this stage, our assumption is that one connector will manage one index, and o
...
+- `in_progress` -> A sync job successfully started.
+- `completed` -> A sync job successfully completed.
+- `failed` -> A sync job failed.
 
-### Elasticsearch mappings for .elastic-connectors:
+#### Elasticsearch mappings for `.elastic-connectors`:
 ```
 "mappings" : {
   "properties" : {
-    "api_key_id" : { "type" : "keyword" },
```
wangch079, line 135: Hi @sphilipse , do you have any concern of removing this `api_key_id`? I see it's still added by Kibana.

sphilipse: Yes, it's used by Kibana to keep track of the API key that has been issued and revoke it if the user issues a new one. Please keep it.

docs/CONNECTOR_PROTOCOL.md
sphilipse: The index_name we should definitely keep, it's very low-effort and provides a good way to filter on index_name jobs rather than connector_id logs. If connectors aren't adding it now, we should fix that (or file an issue to fix it) instead of changing the protocol, I think. 

As for the connector object, you're right that that may require some extra thought and discussion. Given that we're not encryption configuration objects for now, the added risk is significant, I hadn't thought of that. Probably we can remove it for now and revisit it when we add queuing support.

wangch079: Agree, but for a technical preview, we don't need to consider backward-compatibility, and we even don't need to migrate to `v2`.

docs/RELEASING.md
seanstory: 🤷 It doesn't cost us anything. But no, we could remove the rubygems publishing from the `make release_*_dev` targets, since ent-search will just vendor the gems

.github/PULL_REQUEST_TEMPLATE.md
seanstory: 👍 as long as we have a checkbox somewhere that covers testing that the connectors changes don't break ent-search, I've got not problem. :) I agree that this is a short-ish term thing, and that we won't need to do this once Crawler moves, but until then, it would be great to find problems sooner than later.

README.md
seanstory: Ah! Ok this might be a question then for @danajuratoni and @serenachou. I think we still want to market these as being part of Enterprise Search. Elastic Enterprise Search is now supposed to represent pretty much any Elastic business that's _not_ Security or Observability. So it extends well beyond the borders of the Enterprise Search Rails Server. So I don't necessarily think it's _wrong_ to use the Enterprise Search branding here. But I also don't think it would be wrong to switch it with Elastic branding. WDYT?

docs/RELEASING.md
wangch079: Update it to use `make build_service build_utility`, so gems will only be generated in `.gems/` but not pushed to RubyGems.org

README.md
wangch079: >I think we still want to market these as being part of Enterprise Search.

I think you are right. We can remove it when it's removed from the Enterprise Search guide.

I will add it back.

danajuratoni:
Love the split to separate documents, makes it much easier to navigate.
I don't see anything about deploying native connectors on premise. I know there is a bug about "Use a connector" on premise, but running connectors in native mode (no only with the simplified workflow but also multi-config / multi-connector) was supposed to be part of 8.5. 
What is the final status? @wangch079  @tarekziade

tarekziade:
>  I know there is a bug about "Use a connector" on premise, but running connectors in native mode (no only with the simplified workflow but also multi-config / multi-connector) was supposed to be part of 8.5.

For the Python one, there's a flag to make a specific type of connector being native when you run the service. So on-premise, once you deploy a service (as a standalone docker image or directly) alongside an enterprise-search docker that is not running the service, you will end up with the same environment. We should make sure this is documented for 8.5 for each connector, maybe in a specific 'how to deploy native connectors on premise?'

wangch079:
@danajuratoni added how to deploy native/non-native mode on-prem in the [Configuration](https://github.com/elastic/connectors-ruby/blob/8dafa039f3125449728a81ec94b63effd99b0ed5/docs/CONFIG.md) doc

docs/RELEASING.md
```diff
@@ -1,29 +1,27 @@
...
+To release the connector service:
 
-- Set the VERSION file to the new/incremented version on the release branch
-- Make sure all tests and linter pass with `make lint test`
-- PR these changes to the appropriate Connectors release branch
-- Run `make release`
+1. Set the [VERSION](https://github.com/elastic/connectors-ruby/blob/main/VERSION) file to the new/incremented version on the release branch
+2. Make sure all tests and linter pass with `make lint test`
+3. PR these changes to the appropriate connector service release branch
+4. Run `make release_service release_utility`
```
seanstory, line 22:
```suggestion
1. Make sure all tests and linter pass with `make lint test`
2. Run `make release_service release_utility`
3. Set the [VERSION](https://github.com/elastic/connectors-ruby/blob/main/VERSION) file to the new/incremented version on the release branch
4. PR these changes to the appropriate connector service release branch
```

Reordered. We should release the current VERSION (was a snapshot, this makes it a release), then increment the version to the new SNAPSHOT.

seanstory (commented):
This is looking really good. 1 little nit.

docs/RELEASING.md
wangch079: Nice catch @seanstory , I just changed the bullets to numbers without paying attention to the content.

seanstory (approved):

wangch079:
Merged with a follow-up PR: https://github.com/elastic/enterprise-search-team/issues/3111

wangch079:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.5|[](https://github.com/elastic/connectors-ruby/pull/376)|

### Questions ?

PR #369, opened 2022-10-12
Release connector gems of version 8.5.0.5

artem-shelkovnikov:
Bumping VERSION to 8.5.0.4 while releasing the gems


wangch079:
Why bump to `8.5.0.5`, while [the latest release is still `8.5.0.3`](https://rubygems.org/gems/connectors_service)

If it's a typo, we already have a PR to bump to `8.5.0.4`: https://github.com/elastic/connectors-ruby/pull/356

seanstory:
> I've just released 8.5.0.4 and put it in Enterprise Search here:https://github.com/elastic/ent-search/pull/6952, so this PR actually does the bump so that the next build versions would be 8.5.0.5.
>
> I'm honestly not sure at this point if this should be 8.5.0.4 of 8.5.0.5

It should be 8.5.0.5. However, it looks like when you released 8.5.0.4, it didn't create a tag in this git repo, nor did it update rubygems: https://rubygems.org/search?query=connectors_utility. Did you use the `make release*` commands?

Side note, the last scheduled BC is to be cut today. Should this be 8.5.0.4 and 8.5.0.5, or should they be 8.5.1.0 and 8.5.1.2?

artem-shelkovnikov:
Yeah Diana commented in https://github.com/elastic/ent-search/pull/6952 that I should not pull the fixes into last BC for 8.5.0, I'll just close the issue and merge the fixes as a part of 8.5.1.

PR #375, opened 2022-10-18
Allow multi phrase queries

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/2975###

This fixes the issue, that queries under the 'Documents' tabs containing whitespaces weren't possible (also queries with delimiters lead to the same error; in general multi phrase queries). The error appeared actually within lucene and stated, that certain text fields weren't indexed with position data (preventing MultiPhraseQuery). According to the docs https://www.elastic.co/guide/en/elasticsearch/reference/current/index-options.html we need to use 'index_options: positions' to enable phrase queries (https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query-phrase.html). If we don't use 'index_options' explicitly it defaults to 'positions'. I thought it would be good to add it explicitly anyway so that it's clear with one look, what the 'index_options' are set to.

#### Pre-Review Checklist

### DIFF
--- a/lib/utility/elasticsearch/index/mappings.rb
+++ b/lib/utility/elasticsearch/index/mappings.rb
@@ -23,7 +23,7 @@ module Utility
         TEXT_FIELD_MAPPING = {
           type: 'text',
           analyzer: 'iq_text_base',
-          index_options: 'freqs',
+          index_options: 'positions',
           fields: {
             'stem': {
               type: 'text',
@@ -33,18 +33,15 @@ module Utility
               type: 'text',
               analyzer: 'i_prefix',
               search_analyzer: 'q_prefix',
-              index_options: 'docs'
             },
             'delimiter' => {
               type: 'text',
               analyzer: 'iq_text_delimiter',
-              index_options: 'freqs'
             },
             'joined': {
               type: 'text',
               analyzer: 'i_text_bigram',
               search_analyzer: 'q_text_bigram',
-              index_options: 'freqs'
             },
             'enum': {
               type: 'keyword',


timgrein:
@seanstory & @vidok added you both as I'm not sure, wether I'm missing something about these mappings and the implications (didn't found too much in the official docs; don't know if there were specific reasons to use these specific mappings).

vidok (approved):
Looks good to me. 

@timgrein could you please also do a test where you create an app search engine using your search index? 

something like:
1. Create a search index
2. Ingest some data using a connector
3. Create an Engine based on the search index
4. Do some search queries through the engine API (e.g. "curl -X GET http://localhost:3002/api/as/v1/engines/you-engine-name/search?{search_body}"
5. Make sure it works

carlosdelest:
LGTM, though I would like to understand why is match_phrase used. This was not needed in Enterprise Search in the past; could you provide a little more context on what UI piece is using it, and why?

Thanks! 🙏

timgrein:
@carlosdelest If you ingest data via a connector you can visit the 'Documents' tab of this connector and execute arbitrary queries (including multi phrase queries, which search over every field; see screenshot)

image

I cannot really say something about the why as I don't have enough (historical) context 🤔

carlosdelest:
As mentioned in our meeting with @timgrein , this solves the problem for connector-created indices. An App Search created index (or an ES index that is not storing positions) will have the same problem.

I think we should understand more about what the UI is doing and check for additional ways in which the ES query for the documents can be done.

seanstory:
This sounds a lot like: https://discuss.elastic.co/t/query-suggest-not-working-on-app-search-elastic-index-backed-engines/313583/7?u=sean_story, where the community user found that searching on the `.stem` subfield allowed them to execute phrase queries.

I don't think we want to make this mapping change. These mappings were designed to exactly align with App Search, and we should not deviate. In the long term, the ingestion team shouldn't even maintain these mappings, but the Application Data Management team should.

timgrein:
@seanstory Ok, makes sense. Thank you for the context. Then I'll close this PR.

PR #377, opened 2022-10-19
Update connector protocol

wangch079:
## Follow-up PR on https://github.com/elastic/connectors-ruby/pull/367

Changes are:
1. Add missing `)`.
2. Rename `failed` to `error` in `last_sync_status` enum.
3. Describe the `status` of `.elastic-connectors-sync-jobs` more accurately.

PR #378, opened 2022-10-19
[8.5] Update connector protocol (#377)

wangch079:
# Backport

This will backport the following commits from `main` to `8.5`:
 - [Update connector protocol (#377)](https://github.com/elastic/connectors-ruby/pull/377)

### Questions ?


wangch079:
You see so many changes because this backport PR https://github.com/elastic/connectors-ruby/pull/376 is not merged yet.

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -1,5 +1,4 @@
 # Connector Protocol
-**Version 1, Jul 25 2022**
```
tarekziade, line 2: why do we remove this? even though it's versioned I think it would be great to keep the version here and start to increment the version protocol when we do changes that are changing the behavior

wangch079: >even though it's versioned

This is exactly the reason I removed this. 

It's hard for me to predict the date. I have to update this again before I merge the PR

artem-shelkovnikov (approved):

PR #379, opened 2022-10-20
Add ingestion_type field

vidok:
#### Pre-Review Checklist

### DIFF
--- a/docs/CONNECTOR_PROTOCOL.md
+++ b/docs/CONNECTOR_PROTOCOL.md
@@ -127,6 +127,7 @@ In addition to the connector index `.elastic-connectors`, we have an additional 
   error: string; -> Optional error message
   index_name: string; -> The name of the content index
   worker_hostname: string; -> The hostname of the worker to run the sync job
+  ingestion_type: string; -> Enum: [full, partial]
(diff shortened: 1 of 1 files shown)


docs/CONNECTOR_PROTOCOL.md
```diff
@@ -127,6 +127,7 @@ In addition to the connector index `.elastic-connectors`, we have an additional
   error: string; -> Optional error message
   index_name: string; -> The name of the content index
   worker_hostname: string; -> The hostname of the worker to run the sync job
+  ingestion_type: string; -> Enum: [full, partial]
```
artem-shelkovnikov, line 130: My main question here is "what does ingestion_type actually means?".

Is it something that does not have any meaning and only the connector itself gives meaning to it? (e.g. connector defines "full", "incremental" and "apples" syncs and user can just put a different schedule for them, but the connector itself does the different handling of different types by itself)

Or is it something that also modifies the behaviour of the framework? (e.g. `partial` = only upserts, `full` = upserts + deletes).

seanstory:
@vidok did this change ever get made in the protocol? Or can we close this PR?

vidok:
It's not relevant anymore. Closing it. 

Thanks for flagging

PR #380, opened 2022-10-24
tim/allow-strings-and-symbols-for-configurable-fields

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3005##
 
I think it makes sense to not rely on the assumption, that developers implementing a new connector always use symbols for the self_configurable_fields. Therefore I've added .with_indifferent_access to these fields, updated the tests and added another example to the example-connector, so it's more obvious, that both can be used.

#### Pre-Review Checklist

### DIFF
--- a/spec/core/configuration_spec.rb
+++ b/spec/core/configuration_spec.rb
@@ -53,7 +53,7 @@ describe Core::Configuration do
       described_class.update(connector_settings)
     end
 
-    context 'when all configurable fields are set' do
+    context 'when all configurable fields are set with symbols' do
       let(:configuration) { { :foo => { :value => 'bar' } } }
 
       it 'updates status to configured' do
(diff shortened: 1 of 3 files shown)


artem-shelkovnikov:
cc @vidok  - I remember we've started to get rid of `with_indifferent_access` everywhere to speed things up, do you think for this change we need to instead validate that our configurable fields do not use symbols?

timgrein:
Thanks for the hint @artem-shelkovnikov. I think the config check only happens once at startup and if the config is changed, right? I guess that won't be a possible performance issue (respectively, weighs less than a possible inconvenience for connector developers). But happy to hear your opinion also @vidok!

artem-shelkovnikov:
Indeed it's probably not a performance problem any more, but here it's about consistency - I'm not sure about it but it _might_ be a good idea to just drop symbol usage everywhere inside the connectors. Or we can enforce it only for the documents, let's discuss!

timgrein:
Yes, I totally agree on the consistency part 👍 

Just thinking about developers, who are not really used to Ruby and don't know that both options exist and may be confused, why the default values don't appear in the UI. We could either allow both (and therefore make this exception, that `.with_indifferent_access` is explicitly wanted in this case; could add a comment in the code for transparency) or state it clearly in the docs to use only strings + adding your suggested validation, that symbols are not allowed in the configurable fields.

lib/connectors/base/connector.rb
```diff
@@ -19,6 +19,11 @@ def self.display_name
         raise 'Not implemented for this connector'
       end
 
+      # Used as a framework util method, don't override
+      def self.configurable_fields_indifferent_access
+        configurable_fields.with_indifferent_access
+      end
```
artem-shelkovnikov, line 25: I think you need to import the dependency for the method: ` 'active_support/core_ext/hash/indifferent_access'`

spec/core/configuration_spec.rb
```diff
@@ -21,6 +21,7 @@
       allow(connector_settings).to receive(:needs_service_type?).and_return(needs_service_type)
       allow(connector_settings).to receive(:formatted).and_return('')
       allow(connector_class).to receive(:configurable_fields).and_return(configuration)
+      allow(connector_class).to receive(:configurable_fields_indifferent_access).and_return(configuration.with_indifferent_access)
```
artem-shelkovnikov, line 24: Another place with potentially missing `require` statement of  'active_support/core_ext/hash/indifferent_access'

artem-shelkovnikov (approved):
Left a small comment about the require, but looks good apart from that

PR #381, opened 2022-10-25
Termination of connectors shows error

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/2976 ##

This is a first draft, how sync jobs could be updated to show an error, if the app exits unexpectedly (for example by using ctrl-c in the terminal, where the connectors are running). As it's an ongoing discussion, wether to allow mutliple jobs or only one job this PR shouldn't be merged (acts like a reference to be updated with the final implementation, when a decision was made). 

#### Pre-Review Checklist
- [X] Covered the changes with automated tests


timgrein:
related to #357

timgrein:
As #357 is merged now and we can only have one running job per index I've decided to stick with the caching `sync_jobs` mechanism (actual instances of the sync tasks in ruby). I've introduced two hooks, which can accept an arbitrary proc to execute some preparation/shutdown work controlled by the caller. In this specific case it's used for controlling the job cache. This cache is needed to be able to set errors (when f.e. an Interrupt by ctrl-c occurs) for every potential sync job, which is currently running. 

@seanstory Lambdas didn't work somehow with default parameters, so I sticked to proc for now

lib/core/sync_job_runner.rb
```diff
@@ -86,10 +93,13 @@ def do_sync!
         ElasticConnectorActions.complete_sync(@connector_settings.id, job_id, @status.dup)
```
seanstory, line 93: Remind me again why we can't have `rescue *UNEXPECTED_APP_EXITS` in this function, and then just

```
ElasticConnectorActions.complete_sync(@connector_settings.id, job_id, {"error": "Connector service quit unexpectedly.")
raise
```

as its body?

spec/app/dispatcher_spec.rb
```diff
@@ -127,6 +127,21 @@
           it_behaves_like 'logs exception'
         end
 
+        context 'when app exits unexpectedly' do
```
seanstory, line 130: Can you add a test where two SyncJobRunners are running, and both raise an Interrupt at some point, and ensure both get set with the correct failure status?

spec/app/dispatcher_spec.rb
```diff
@@ -127,6 +127,21 @@
...
+            allow(connector_settings).to receive(:index_name).and_return('')
+            allow(job_runner).to receive(:execute).and_raise(Interrupt)
+          end
+
+          it 'shuts down sync jobs' do
+            expect(described_class).to receive(:shutdown_sync_jobs_with_error)
+            expect { described_class.start! }.to_not raise_error
+          end
+        end
+
```
seanstory, line 144: Can you update other specs or add new ones to show that the pre/post hooks are operating correctly and adding/removing items from the cache as you expect?

seanstory (commented):
Good stuff. Left a few comments. Should probably get one of @artem-shelkovnikov or @mchernyavskaya or @wangch079 to give it a review too.

wangch079 (commented):
Hi @timgrein , please see this PR: https://github.com/elastic/connectors-ruby/pull/318

I'm not sure whether `SystemExit` or `Interrupt` will still be raised if the signals are trapped. 

I also see you used a job cache.  Crawler is managed by Crawl Queue Manager (CQM), which is exactly the role of dispatcher. CQM uses a worker array to manage threads, which gives more control than thread pool. Eventually Crawler code will be migrated to connectors-ruby, and I'm thinking whether we want to adopt the Crawler approach.

lib/app/dispatcher.rb
```diff
@@ -86,8 +107,14 @@ def start_sync_task(connector_settings)
         pool.post do
           Utility::Logger.info("Initiating a sync job for #{connector_settings.formatted}...")
           Core::ElasticConnectorActions.ensure_content_index_exists(connector_settings.index_name)
-          job_runner = Core::SyncJobRunner.new(connector_settings)
-          job_runner.execute
+
+          job_instance = Core::SyncJobRunner.new(connector_settings)
+          job_instance_obj_id = job_instance.object_id
```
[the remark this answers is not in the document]
timgrein: Makes sense! I've to either separate the claiming and execution of a job (to get the job_id before execution for caching) or add the job_id as a parameter to both procs, which is passed as `pre_sync_hook`. I go with the 2nd option so the caller doesn't have to worry about claiming the job_id. WDYT?

lib/app/dispatcher.rb
```diff
@@ -77,6 +93,11 @@ def start_polling_jobs!
             Utility::Logger.error("Unknown task type: #{task}. Skipping...")
           end
         end
+      rescue *Utility::UNEXPECTED_APP_EXITS => e
+        error_message = 'Connector service quit unexpectedly.'
+
+        Utility::ExceptionTracking.log_exception(e, error_message)
+        shutdown_sync_jobs_with_error(error_message)
       rescue StandardError => e
         Utility::ExceptionTracking.log_exception(e, 'The connector service failed due to unexpected error.')
```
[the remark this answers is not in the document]
timgrein: yes, you're right 👍

lib/core/sync_job_runner.rb
timgrein: I've tried that, but it doesn't seem to catch the interrupt in these spawned job sync threads, but catches it in the parent thread 🤔 . I'll read up some stuff and do some local experiments. Not 100% sure, how interrupts get handled in Ruby (probably also specific to the ruby implementation?)

artem-shelkovnikov:
I'm not sure about the changes in the PR as we'd like service to handle the cases differently:

1. If the service crashes we want in future (probably in close-ish future) to be able to pick up the job from the same place or even restart the job.
2. In more far future we want to distribute the work between multiple instances/hosts of connectors services, in this case loss of one service should also lead to a retry by a different agent.

Some other little thoughts I had when thinking of the change:

- Connector can anyway crash in such a way, that we won't be able update the job status in Kibana, what do we do with these jobs?
- Interrupt/SystemExit are nice IMO to write to logs some information, but doing blocking network calls in the handler just feel unnatural (purely opinion, it actually might be a good thing to do)

I think based on the points we can try a different direction - not fail the job when they are interrupted, but instead try to pick it up again and run again. This work is not possible now as jobs do not store any metadata of the sync, neither they store the current state of sync. After 8.6 we will have this information and start implementing more resilient job handling. Also with proposal of separation of Producer and Consumer the flow will change.

I think the change should be postponed until later and we should discuss the concept of Job resilience in a bigger group.

timgrein:
@artem-shelkovnikov Agree, I also think there are too many moving parts now, which will affect this change (Crawl Queue Manager coming into connectors-ruby, handle Interrupt/SystemExit on App level PR also currently open). What do we do about https://github.com/elastic/enterprise-search-team/issues/2976? Should we also postpone it until we've more resilient job handling? Then we should probably add a note to this bug, that it's postponed until more resilient job handling is finished.

timgrein:
@artem-shelkovnikov Yes the reason is, that the `sync_job_runner.rb` has an `ensure` block within `do_sync!`, which forces to complete the sync (as no error is set -> job marked is completed). I've fixed it by catching Interrupt, SystemExit with this PR to set the error before we enter the `ensure` block.

timgrein:
After our conversion @artem-shelkovnikov and I came to the conclusion, that it's probably not worth to introduce such a big change for this bug and keep the fix simple as a lot of surrounding pieces will be subject to change in the near future. We now just check, if the sync job finished execution. Otherwise we show an error message, that the sync thread didn't finish execution and further investigation is needed.

lib/core/sync_job_runner.rb
```diff
@@ -83,10 +85,12 @@ def do_sync!
         Utility::Logger.info("Upserted #{@status[:indexed_document_count]} documents into #{@connector_settings.index_name}.")
         Utility::Logger.info("Deleted #{@status[:deleted_document_count]} documents into #{@connector_settings.index_name}.")
 
+        @status[:error] = 'Sync thread didn\'t finish execution. Check connector logs for more details.' unless @sync_finished
```
artem-shelkovnikov, line 88: Looks like this will trigger also in cases when an error happens, because @sync_finished is never set to `true` in `rescue` block.

Curiously it's not caught by the test, I think it's worth to add a test for it:
1. When an StandardError is caught, then this error is properly put to `status[:error]`
2. When no error is caught, but sync is not finished, only then set the error to the message on this line

artem-shelkovnikov, line 88: I think also this line is worth the comment that explains why is has to be done this way - about the interrupts and how they are handled in a multithreaded app

artem-shelkovnikov (commented):
Looks like a simple step forward that fixes the issue!

artem-shelkovnikov (approved):
LGTM!

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.5|[](https://github.com/elastic/connectors-ruby/pull/402)|

### Questions ?

PR #383, opened 2022-10-28
update connector protocol

seanstory:
## part of https://github.com/elastic/enterprise-search-team/issues/3148

This updates the CONNECTOR_PROTOCOL.md document to include up-to-date mappings, as well as the proposed mapping changes for filtering rules.

Related PR: https://github.com/elastic/ent-search/pull/6977


docs/CONNECTOR_PROTOCOL.md
```diff
@@ -82,29 +110,111 @@ This is our main communication index, used to communicate the connector's config
...
     "language" : { "type" : "keyword" },
     "last_seen" : { "type" : "date" },
-    "last_sync_error" : { "type" : "text" },
+    "last_sync_error" : { "type" : "keyword" },
     "last_sync_status" : { "type" : "keyword" },
     "last_synced" : { "type" : "date" },
-    "last_indexed_document_count" : { "type" : "integer" },
-    "last_deleted_document_count" : { "type" : "integer" },
+    "last_indexed_document_count" : { "type" : "long" },
+    "last_deleted_document_count" : { "type" : "long" },
```
seanstory, line 208: These are unfortunate changes, but are necessary because these were missing from the ent-search mapping definition. Elasticsearch therefore dynamically mapped these numeric fields as `long` in 8.5, so in order to avoid a reindex operation, we now need to explicitly define the mapping type as `long` to match.

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -82,29 +110,111 @@ This is our main communication index, used to communicate the connector's config
...
+          }
+        }
+      }
+    },
     "index_name" : { "type" : "keyword" },
     "is_native" : { "type" : "boolean" },
     "language" : { "type" : "keyword" },
     "last_seen" : { "type" : "date" },
-    "last_sync_error" : { "type" : "text" },
+    "last_sync_error" : { "type" : "keyword" },
```
seanstory, line 204: The mapping was previously defined as Keyword, and this document was just out of sync with ent-search and kibana.

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -82,29 +110,111 @@ This is our main communication index, used to communicate the connector's config
...
 ```
 "mappings" : {
+  "_meta" : {
+    "pipeline" : {
+      "default_extract_binary_content" : true,
+      "default_name" : "ent-search-generic-ingestion",
+      "default_reduce_whitespace" : true,
+      "default_run_ml_inference" : true
+    },
+    "version" : "1"
```
wangch079, line 120: What is this `version` used for?

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -82,29 +110,111 @@ This is our main communication index, used to communicate the connector's config
...
+  },
   "properties" : {
     "api_key_id" : { "type" : "keyword" },
     "configuration" : { "type" : "object" },
     "description" : { "type" : "text" },
-    "error" : { "type" : "text" },
+    "error" : { "type" : "keyword" },
```
wangch079, line 126: I think we probably should change [here](https://github.com/elastic/ent-search/blob/e847748e7bb192787985caa7bc5d33bb2e068856/shared_togo/lib/shared_togo/crawler2/connector.rb#L32) instead of in the doc.

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -125,6 +235,28 @@ In addition to the connector index `.elastic-connectors`, we have an additional
   connector_id: string; -> ID of the connector document in .elastic-connectors
   status: string; -> Same enum as last_sync_status above except the null value
   error: string; -> Optional error message
+  filtering: {          -> Filtering rules
```
wangch079, line 238: My 2 cents on whether we should copy the whole connector protocol to job index:
https://docs.google.com/document/d/1E2NPLPbJb6cu_Ir3PznO7AcV1wTLVjh53ma9Q9g4ECc/edit?disco=AAAAi7bhH3U

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -136,16 +268,48 @@ In addition to the connector index `.elastic-connectors`, we have an additional
 #### Elasticsearch mappings for `.elastic-connectors-sync-jobs`:
 ```
 "mappings" " {
+  "_meta" : {
+    "version" : 1
+  },
   "properties" : {
+    "completed_at" : { "type" : "date" },
```
wangch079, line 275: Is it the convention to have the properties sorted?

docs/CONNECTOR_PROTOCOL.md
seanstory: It's a standard thing to put in your index mappings to help version the mappings. Kibana and ent-search have been using this since the beginning.

docs/CONNECTOR_PROTOCOL.md
seanstory: We could, but then we'd need to make a migration to reindex data for all our 8.4 and 8.5 customers. And since it's not really critical that we be able to execute text search queries on this field right now, I don't think this is the time to do that.

docs/CONNECTOR_PROTOCOL.md
seanstory: It is in the Kibana codebase. And that's where we first defined these mappings, so this doc has been alphabetically sorted. I'm not tied to it here, if you'd rather that we re-order them, but keeping them ordered consistently across all our sources makes it easier to copy-paste from one codebase to the next and not have to manage weird diffs.

docs/CONNECTOR_PROTOCOL.md
seanstory: I'm not going to add the whole connector definition as part of this PR. I only need the filtering rules. I'm not opposed to Dmitrii making that change later in this release when he gets to it, if that's the right call then.

docs/CONNECTOR_PROTOCOL.md
wangch079: I think this is not documented as a special field in [metadata field](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-meta-field.html), do you mean it's something implemented by Enterprise Search?

Will the version be increased if the meta is updated (e.g. `default_run_ml_inference` is set to `false`)? Because I don't see it happening in [Kibana](https://github.com/elastic/kibana/blob/73f1878983113d061890a8648f0548ac46823ca5/x-pack/plugins/enterprise_search/server/lib/pipelines/update_default_pipeline.ts#L33-L36).

Or will the version be increased if we add more fields to meta?

docs/CONNECTOR_PROTOCOL.md
wangch079: Does [this](https://github.com/elastic/ent-search/pull/6984#discussion_r1010023401) make sense?

I agree this is not critical to change now.

docs/CONNECTOR_PROTOCOL.md
wangch079: >I'm not opposed to Dmitrii making that change later in this release when he gets to it, if that's the right call then.

We need to take care of backward compatibility because Crawler is GA.

Even if we want to construct the snapshot, do you think it's better to group them? e.g. 
```
connector: {
  filtering: { ... },
  configuration: { ... },
  ...
}
```

docs/CONNECTOR_PROTOCOL.md
wangch079: I personally prefer those fields to be logically ordered. e.g. It's strange to see `completed_at` at the top and `started_at` at the bottom, `indexed_documents_count` and `deleted_documents_count` are supposed to be together.

Hi @sphilipse , any particular reason to sort them alphabetically?

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -40,6 +40,34 @@ This is our main communication index, used to communicate the connector's config
...
+      },
+      advanced_snippet: {   -> used for filtering data from the 3rd party at query time
+        value: object,      -> this JSON object is passed directly to the connector
+        created_at: string, -> when this JSON object was created
+        updated_at: string  -> when this JSON object was last edited
+      },
+      validation: {
+        state: string,      -> one of ["edited", "valid", "invalid"]
+        errors: {
+          ids: string,      -> the id(s) of any rules that are deemed invalid
```
mchernyavskaya, line 64: According to the [implementation doc](https://docs.google.com/document/d/1llPDGzduj8lST0N45v_G4PB0s0RUOBFurOIsBAD-Fyg/edit#heading=h.l2heyce8rz9r), `errors` is an array, and I see an object here.

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -40,6 +40,34 @@ This is our main communication index, used to communicate the connector's config
                            fields
   description: string;  -> the description of the connector
   error: string;        -> Optional error message
+  filtering: {          -> Filtering rules
+    domain: string,     -> what data domain these rules apply to
+    active: {           -> "active" rules are run in jobs. 
+      rules: {
```
mchernyavskaya, line 46: Isn't `rules` here supposed to be an array?

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -40,6 +40,34 @@ This is our main communication index, used to communicate the connector's config
...
+        updated_at: string  -> when the rule was last edited
+      },
+      advanced_snippet: {   -> used for filtering data from the 3rd party at query time
+        value: object,      -> this JSON object is passed directly to the connector
+        created_at: string, -> when this JSON object was created
+        updated_at: string  -> when this JSON object was last edited
+      },
+      validation: {
+        state: string,      -> one of ["edited", "valid", "invalid"]
+        errors: {
```
mchernyavskaya, line 63: According to the [implementation doc](https://docs.google.com/document/d/1llPDGzduj8lST0N45v_G4PB0s0RUOBFurOIsBAD-Fyg/edit#heading=h.l2heyce8rz9r), the `errors` field is supposed to be an array. Here, I see an object. I'd say array makes more sense.

wangch079 (approved):

PR #384, opened 2022-10-31
Revert "Add an elasticsearch client bulk patch to reduce memory usage"

artem-shelkovnikov:
Reverts elastic/connectors-ruby#364

The change was tested in the batch with the other changes and seemed like something that reduces our memory usage significantly.

After thorough testing I wasn't able to find correlation between this change and actual memory usage reduction, thus reverting it.

A little meta about testing:

The assumption was that `.clone` function is a deep cloning function, but documentation states that it's a _shallow_ copy. `arguments` that are passed into the function for us consists only two properties: `body` and `pipeline` where `body` is an array and `pipeline` is a string.

Therefore shallow clone _should not copy the array_ which is the largest size object there. Testing showed that it's indeed the case and amount of memory is approximately the same.

### DIFF
--- a/lib/utility/es_client.rb
+++ b/lib/utility/es_client.rb
@@ -8,10 +8,6 @@
 
 require 'logger'
 require 'elasticsearch'
-require 'elasticsearch/api'
-require 'elasticsearch/api/namespace/common'
-require 'elasticsearch/api/utils'
-require 'elasticsearch/api/response'
 
 module Utility
   class EsClient < ::Elasticsearch::Client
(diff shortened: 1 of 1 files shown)


artem-shelkovnikov:
Ruby code snippet used for testing:

```ruby
require 'objspace'

ObjectSpace.trace_object_allocations_start

require 'elasticsearch'
require 'elasticsearch/api'
require 'elasticsearch/api/namespace/common'
require 'elasticsearch/api/utils'
require 'elasticsearch/api/response'

require 'newrelic_rpm'

$mem_sampler = NewRelic::Agent::Samplers::MemorySampler.new.sampler

def random_text(word_count, max_word_len)
  word_count.times.map { (0...(rand(max_word_len))).map { ('a'..'z').to_a[rand(26)] }.join }.join(" ")
end

def generate_large_test_object(id)
  text = random_text(500, 10)

  obj = {
    :id => id,
    :items => 200.times.map do
      {
        :text => text
      }
    end
  }
end

class ClientWithoutBulkPatch < ::Elasticsearch::Client
  def bulk(arguments = {})
    p "Before: #{ $mem_sampler.get_sample }"
    raise ArgumentError, "Required argument 'body' missing" unless arguments[:body]

    arguments = arguments.clone
    headers = arguments.delete(:headers) || {}

    body   = arguments.delete(:body)

    _index = arguments.delete(:index)

    method = Elasticsearch::API::HTTP_POST
    path   = if _index
               "#{Utils.__listify(_index)}/_bulk"
             else
               "_bulk"
             end
    params = Utils.process_params(arguments)

    if body.is_a? Array
      payload = Elasticsearch::API::Utils.__bulkify(body)
    else
      payload = body
    end

    headers.merge!("Content-Type" => "application/x-ndjson")
    Elasticsearch::API::Response.new(
      perform_request(method, path, params, payload, headers)
    )
    p "After: #{ $mem_sampler.get_sample }"
  end
end

class ClientWithBulkPatch < ::Elasticsearch::Client
  def bulk(arguments = {})
    p "Before: #{ $mem_sampler.get_sample }"
    headers = arguments.delete(:headers) || {}

    body = arguments.delete(:body)

    index = arguments.delete(:index)

    method = ::Elasticsearch::API::HTTP_POST
    path = if index
             "#{Utils.__listify(index)}/_bulk"
           else
             '_bulk'
           end
    params = ::Elasticsearch::API::Utils.process_params(arguments)

    payload = if body.is_a? Array
                ::Elasticsearch::API::Utils.__bulkify(body)
              else
                body
              end

    headers['Content-Type'] = 'application/x-ndjson'
    ::Elasticsearch::API::Response.new(
      perform_request(method, path, params, payload, headers)
    )
    p "After: #{ $mem_sampler.get_sample }"
  end
end

objects_to_send = 20.times.map { generate_large_test_object(50) }

client = ClientWithoutBulkPatch.new(
  :hosts => '127.0.0.1:9200',
  :api_key => 'WDhnX0g0UUJHSDdSa3cxYkZ4Mzk6RTlndGcxdjVSVVNDcF9nWDRPV0lVQQ=='
)

index_name = 'bulk_test'
bulk_operations = objects_to_send.map { |o| { :index => { :_index => index_name, :_id => o[:id], :data => o } }  }

client.bulk(:body => bulk_operations)

ObjectSpace.dump_all(output: File.open('heap.json','w'))
```

wangch079 (approved):

PR #385, opened 2022-10-31
Add filtering capability (by an advanced config) to MongoDB connector

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3150##

This PR introduces the capability to filter a MongoDB collection by an advanced config (rules are already present in the API, implementation will follow in another PR). Therefore find queries and aggregation pipelines are now allowed. The filtering options are currently passed inside a general job_description object (introduced to be able to extend parameters easily later on without polluting the yield_documents API). This PR also keeps the recently introduced [memory optimisation](https://github.com/elastic/connectors-ruby/pull/362) by differentiating skip/limit parameters introduced by filtering (used for overall skipping/limiting) vs skip/limit parameters used for paging (needed for memory optimisation).

Additional notes:
* Includes a new strings utility class (for formatting string arrays)
* Includes a commons utility class (to be able to use an already existing utility method outside of connector settings)

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/connectors-ruby/pull/362

## Release Note
Introduce MongoDB filtering capabilities.


lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
         with_client do |client|
           # We do paging using skip().limit() here to make Ruby recycle the memory for each page pulled from the server after it's not needed any more.
           # This gives us more control on the usage of the memory (we can adjust PAGE_SIZE constant for that to decrease max memory consumption).
           # It's done due to the fact that usage of .find.each leads to memory leaks or overuse of memory - the whole result set seems to stay in memory
           # during the sync. Sometimes (not 100% sure) it even leads to a real leak, when the memory for these objects is never recycled.
-          cursor = client[@collection].find
+          cursor = client[@collection].send(function, *args)
           skip = 0
 
+          options = args[1]
```
timgrein, line 140: Options are the same for retrieval functions (find/aggregate) in MongoDB (also at the same place), therefore we can assume that they could exist as second parameter.

lib/utility/errors.rb
```diff
@@ -32,6 +32,8 @@ def to_h
   end
 
   class ClientError < StandardError; end
+
+  class InvalidFilterConfigError < ClientError; end
```
seanstory, line 36: Why inherit from ClientError and not from StandardError?

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+        return unless filtering_present?(rules, config)
+
+        check_find_and_aggregate_present!(config)
+
+        check_find_and_aggregate_missing!(config)
+      end
+
+      def check_find_and_aggregate_missing!(config)
+        allowed_keys = Utility::Strings.format_string_array(ALLOWED_TOP_LEVEL_FILTER_KEYS, default: ' ')
+        keys_present = Utility::Strings.format_string_array(config.keys, default: ' ')
```
seanstory, line 116: Why is the special formatting necessary? Just doing a `.to_s` on a string array gets you the default behavior I think you're looking for:

```
irb(main):002:0> a = ['foo', 'bar']
=> ["foo", "bar"]
irb(main):003:0> s = a.to_s
=> "[\"foo\", \"bar\"]"
irb(main):004:0> puts s
["foo", "bar"]
=> nil
```

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+
+      private
+
+      def extract_rules(job_description)
+        Utility::Commons.return_if_present(job_description.dig(:filtering, :rules), DEFAULT_RULES)
+      end
+
+      def extract_filter_config(job_description)
+        Utility::Commons.return_if_present(job_description.dig(:filtering, :advanced_config), DEFAULT_ADVANCED_CONFIG)
+      end
```
seanstory, line 92: These functions and their constants could be moved somewhere more generic, as none of this is specific to Mongodb

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+        args = NO_ARGS
+
+        function, args = setup_find(filter_config) if find_present?(filter_config)
+
+        function, args = setup_aggregate(filter_config) if aggregate_present?(filter_config)
+
+        [function, args]
+      end
+
+      def check_filter_config!(rules, config)
```
seanstory, line 106: I've typically seen the `!` suffix used to signal that a side effect occurs in this function, and is used to differentiate when there are two versions of the function - both a (`def foo()` and a `def foo!()`). That doesn't seem to be the case here. Are you trying to signal that the function may raise an error?

seanstory (commented):
Nice work! There's a lot to review here, and I haven't finished going through it yet, but figured I'd give the comments I've made so far. I'll try to come back to this later today or tomorrow for a more thorough review.

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+        return unless filtering_present?(rules, config)
+
+        check_find_and_aggregate_present!(config)
+
+        check_find_and_aggregate_missing!(config)
```
artem-shelkovnikov, line 111: These two methods can actually be replaced by a single simpler method.

lib/connectors/mongodb/connector.rb
artem-shelkovnikov: +1 to that, seems like an unnecessary function that makes reading the code harder

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+      end
+
+      def check_find_and_aggregate_missing!(config)
+        allowed_keys = Utility::Strings.format_string_array(ALLOWED_TOP_LEVEL_FILTER_KEYS, default: ' ')
+        keys_present = Utility::Strings.format_string_array(config.keys, default: ' ')
+
+        find_and_aggregate_missing = !find_present?(config) && !aggregate_present?(config)
+        both_missing_msg = "Only one of #{allowed_keys} is allowed in the filtering object. Keys present: '#{keys_present}'."
+
+        raise Utility::InvalidFilterConfigError.new(both_missing_msg) if find_and_aggregate_missing
```
artem-shelkovnikov, line 121: Problem - error is raised only if `find_and_aggregate_missing` but `both_missing_msg` is generated anyway. It's a small waste of resources, but it also makes reading the method harder.

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+      end
+
+      def check_find_and_aggregate_present!(config)
+        find_and_aggregate_present = find_present?(config) && aggregate_present?(config)
+        both_present_msg = '\'find\' and \'aggregate\' functions cannot be used at the same time for MongoDB. Please drop one from the configuration.'
+
+        raise Utility::InvalidFilterConfigError.new(both_present_msg) if find_and_aggregate_present
```
artem-shelkovnikov, line 128: Problem - error is raised only if `find_and_aggregate_present` but `both_present_msg` is generated anyway. It's a small waste of resources, but it also makes reading the method harder.

lib/connectors/mongodb/connector.rb
artem-shelkovnikov: This method can just verify 2 things:
1. Only one hash key is present
2. This key's value is expected ('find' or 'aggregate')

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+      end
+
+      def call_db_function_on_collection(function, args, &on_doc_serialization)
         with_client do |client|
           # We do paging using skip().limit() here to make Ruby recycle the memory for each page pulled from the server after it's not needed any more.
           # This gives us more control on the usage of the memory (we can adjust PAGE_SIZE constant for that to decrease max memory consumption).
           # It's done due to the fact that usage of .find.each leads to memory leaks or overuse of memory - the whole result set seems to stay in memory
           # during the sync. Sometimes (not 100% sure) it even leads to a real leak, when the memory for these objects is never recycled.
-          cursor = client[@collection].find
+          cursor = client[@collection].send(function, *args)
```
artem-shelkovnikov, line 137: We should not make dynamic calls, this makes reading, maintaining and refactoring the code harder.

`case/switch` might be a better choice here.

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+          # if no overall limit is specified by filtering use -1 to not break ingestion, when no overall limit is specified (found_overall is only increased,
+          # thus can never reach -1)
+          overall_limit = -1
+
+          unless options.nil?
+            # there could be a skip parameter defined for filtering
+            skip = Utility::Commons.return_if_present(options[:skip], 0)
+            # there could be a limit parameter defined for filtering -> used for an overall limit (not a page limit, which was introduced for memory optimization)
+            overall_limit = Utility::Commons.return_if_present(options[:limit], -1)
+          end
```
artem-shelkovnikov, line 152: I'm not sure exposing/allowing skip/limit brings value. While `limit` might make sense for some customer, `skip` does not seem to have any benefit but can provide weird experience if user copy/pasted the query incorrectly.

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
 
-      private
+      def setup_aggregate(advanced_config)
+        aggregate = advanced_config[:aggregate]
+
+        pipeline = aggregate[:pipeline]
+        options = extract_options(aggregate)
+
+        if !pipeline_present?(pipeline) && !options_present?(options)
+          Utility::Logger.warn('\'Aggregate\' was specified with an empty pipeline and empty options.')
```
artem-shelkovnikov, line 187: What will be the actual result of it? Empty result set?

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+      end
+
+      def setup_find(advanced_config)
+        find = advanced_config[:find]
+
+        filter = find[:filter]
+        options = extract_options(find)
+
+        if !filter_present?(filter) && !options_present?(options)
+          Utility::Logger.warn('\'Find\' was specified with an empty filter and empty options.')
```
artem-shelkovnikov, line 202: What will be the actual result of it? Empty result set?

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+        arguments = [pipeline, options]
+
+        [AGGREGATE, arguments]
+      end
+
+      def setup_find(advanced_config)
+        find = advanced_config[:find]
+
+        filter = find[:filter]
+        options = extract_options(find)
```
artem-shelkovnikov, line 199: Can `find` have other options except for `:filter`?

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
           end
         end
       end
 
-      private
+      def setup_aggregate(advanced_config)
+        aggregate = advanced_config[:aggregate]
+
+        pipeline = aggregate[:pipeline]
+        options = extract_options(aggregate)
```
artem-shelkovnikov, line 184: Can `aggregate` have other options than `:pipeline`?

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+        [FIND, arguments]
+      end
+
+      def extract_options(advanced_config)
+        options_present?(advanced_config[:options]) ? advanced_config[:options] : EMPTY_OPTIONS
+      end
+
+      def find_present?(advanced_config)
+        advanced_config[:find].present?
+      end
```
artem-shelkovnikov, line 216: Can be probably safely inlined

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+
+      def filter_present?(filter)
+        !filter.nil? && filter != EMPTY_FILTER
+      end
+
+      def aggregate_present?(advanced_config)
+        advanced_config[:aggregate].present?
+      end
```
artem-shelkovnikov, line 224: Can be probably safely inlined

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+        options_present?(advanced_config[:options]) ? advanced_config[:options] : EMPTY_OPTIONS
+      end
+
+      def find_present?(advanced_config)
+        advanced_config[:find].present?
+      end
+
+      def filter_present?(filter)
+        !filter.nil? && filter != EMPTY_FILTER
+      end
```
artem-shelkovnikov, line 220: 1 usage of function found, can be inlined

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+        end
+
+        arguments = [filter, options]
+
+        [FIND, arguments]
+      end
+
+      def extract_options(advanced_config)
+        options_present?(advanced_config[:options]) ? advanced_config[:options] : EMPTY_OPTIONS
+      end
```
artem-shelkovnikov, line 212: What is contained inside `:options`?

artem-shelkovnikov: It can probably go to the base class

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+        !filter.nil? && filter != EMPTY_FILTER
+      end
+
+      def aggregate_present?(advanced_config)
+        advanced_config[:aggregate].present?
+      end
+
+      def pipeline_present?(pipeline)
+        !pipeline.nil? && pipeline != EMPTY_PIPELINE
+      end
```
artem-shelkovnikov, line 228: 1 usage of function found, can be inlined

lib/core/connector_settings.rb
```diff
@@ -80,20 +83,33 @@ def scheduling_settings
       self[:scheduling]
     end
 
+    def filtering
+      Utility::Commons.return_if_present(@elasticsearch_response[:filtering], DEFAULT_FILTERING)
```
artem-shelkovnikov, line 87: What is `:filtering`?

lib/core/connector_settings.rb
```diff
@@ -80,20 +83,33 @@ def scheduling_settings
...
+    end
+
+    def active_filter_config
+      # assume for now, that first object in filtering array or a filter object itself is the only filtering object
+      filter = filtering.is_a?(Array) ? filtering[0] : filtering
```
artem-shelkovnikov, line 92: Might be a really good idea to warn/raise error if unexpected values that we can transparently handle happens

lib/core/connector_settings.rb
```diff
@@ -19,6 +19,9 @@ class ConnectorSettings
     DEFAULT_REDUCE_WHITESPACE = true
     DEFAULT_RUN_ML_INFERENCE = true
 
+    DEFAULT_FILTERING = {}
+    DEFAULT_ADVANCED_FILTER_CONFIG = {}
```
artem-shelkovnikov, line 23: These should not be constants IMO

lib/core/sync_job_runner.rb
```diff
@@ -59,7 +59,10 @@ def do_sync!
 
         Utility::Logger.debug("#{existing_ids.size} documents are present in index #{@connector_settings.index_name}.")
 
-        @connector_instance.yield_documents do |document|
+        filtering = @connector_settings.filtering
+        job_description = { :filtering => filtering }
```
artem-shelkovnikov, line 63: Why not just pass filtering rules as is?

lib/utility/commons.rb
```diff
@@ -0,0 +1,20 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+module Utility
+  class Commons
```
artem-shelkovnikov, line 10:
```suggestion
module Utility
  class Common
```

lib/utility/strings.rb
```diff
@@ -0,0 +1,21 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+module Utility
+  class Strings
```
artem-shelkovnikov, line 10:
```suggestion
module Utility
  class String
```

artem-shelkovnikov: I understand the naming is coming from `Utility::Errors` but we should rename it to `Utility::Error` in a separate PR :D

lib/utility/strings.rb
```diff
@@ -0,0 +1,21 @@
...
+    class << self
+      def format_string_array(elements = [], default: ' ', delimiter: '\'', separator: ', ')
+        return default if elements.nil? || elements.empty?
+
+        separated_elements = elements.join("#{delimiter}#{separator}#{delimiter}")
+
+        "#{delimiter}#{separated_elements}#{delimiter}"
+      end
```
artem-shelkovnikov, line 18: It's a complex function and there seems to be no need for it (Sean mentioned it above and I agree with it)

artem-shelkovnikov (changes requested):
I've added some comments about things to fix, but before approaching them I'd like to reflect on the amount of work needed to do the `find`/`aggregate` work and amount of edge cases, validation and such it produces versus fully offloading this problem to MongoDB's views feature.

Do we implement something that is absolutely not achievable by MongoDB views? If MongoDB views achieve everything or almost everything, I would actually switch to this - provide guidance to the customers via documentation and leave MongoDB be a really simple connector.

seanstory:
> Do we implement something that is absolutely not achievable by MongoDB views? If MongoDB views achieve everything or almost everything, I would actually switch to this - provide guidance to the customers via documentation and leave MongoDB be a really simple connector.

That's probably something to bring up in [this section of the implementation document](https://docs.google.com/document/d/1llPDGzduj8lST0N45v_G4PB0s0RUOBFurOIsBAD-Fyg/edit?pli=1#heading=h.w7zy4hbynecl). Please make sure to tag in Dana.

lib/connectors/base/connector.rb
```diff
@@ -37,7 +37,19 @@ def initialize(configuration: {})
         @configuration = configuration.dup || {}
       end
 
-      def yield_documents; end
+      def yield_documents(job_description = {}); end
```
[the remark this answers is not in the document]
seanstory: This was guidance from me. Our connectors instances are initialized _before_ we've claimed the `.elastic-connectors-sync-job` record. But it makes sense that a given call to `yield_documents` should be done within the context of a specific job. This will become even more true with the work that @vidok is proposing for the Job Processing changes.

lib/connectors/mongodb/connector.rb
seanstory: This was another suggestion by me 😅.  I disagree that dynamic calls are always wrong. In this case, the advanced config is specifically designed to be tightly coupled to the API of the mongo gem. We really do want to explicitly pass whatever is in that config.

Case/switch might also not be terrible though - previously, the code was like:
```
client[@collection].find.each do |doc|
  ...do a whole bunch of stuff...
end  
```

And I didn't want it to be:
```
if <condition>
  client[@collection].find(params).each do |doc|
    ... a bunch of stuff...
  end
else
  client[@collection].aggregate(params).each do |doc|
      ... a copied codeblock of a bunch of stuff...
   end
 end
 ```
 
 but now that we're using the cursor, it feels more viable to construct that `cursor` in an if/else or case/switch block.

File: lib/connectors/mongodb/connector.rb (hunk 10 above)
Comment (seanstory) [reply lib/connectors/mongodb/connector.rb:152]: I don't want to try to set artificial constraints here. The implementation doc suggested we let the user provide whatever options they want so that they can tune how they fetch data. `Skip` could let the user skip an initial set of records that they don't care to index. `limit` lets them set the batch size as it makes sense for their performance.

File: lib/connectors/mongodb/connector.rb (hunk 13 above)
Comment (seanstory) [reply lib/connectors/mongodb/connector.rb:199]: Only `filter` and `options`: https://www.mongodb.com/docs/ruby-driver/current/api/Mongo/Collection.html#find-instance_method

File: lib/connectors/mongodb/connector.rb (hunk 14 above)
Comment (seanstory) [reply lib/connectors/mongodb/connector.rb:184]: Only `pipeline` and `options`: https://www.mongodb.com/docs/ruby-driver/current/api/Mongo/Collection.html#aggregate-instance_method

File: lib/core/connector_settings.rb (hunk 20 above)
Comment (seanstory) [reply lib/core/connector_settings.rb:87]: See the [implementation document](https://docs.google.com/document/d/1llPDGzduj8lST0N45v_G4PB0s0RUOBFurOIsBAD-Fyg/edit?pli=1#heading=h.l2heyce8rz9r)

File: lib/utility/errors.rb (hunk 2 above)
Comment (timgrein) [reply lib/utility/errors.rb:36]: I thought that an invalid filter config can only come from someone typing in an invalid config in the client. But also happy to update it to extend from StandardError.

File: lib/connectors/mongodb/connector.rb (hunk 3 above)
Comment (timgrein) [reply lib/connectors/mongodb/connector.rb:116]: I thought it would be nice to have more control (omitting brackets for example, showing a space instead of an empty string -> not showing a space/default message could lead to "Some message:.", if no elements are present, which looks a bit weird IMO). But I see the point of just using the built-in `to_s` in favor of a cleaner codebase. I'll drop the class then 👍

File: lib/connectors/mongodb/connector.rb (hunk 4 above)
Comment (timgrein) [reply lib/connectors/mongodb/connector.rb:92]: 👍

File: lib/connectors/mongodb/connector.rb [hunk 28]
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
         @direct_connection = configuration.dig(:direct_connection, :value)
       end
 
-      def yield_documents
+      def yield_documents(job_description = {}, &on_doc_serialization)
```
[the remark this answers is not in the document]
timgrein: It's used in `call_db_function_on_collection`, when a document is serialized. We can also pass it implicitly, but I thought it would be useful to have it explicitly in the code.

lib/connectors/mongodb/connector.rb
timgrein: Yes, but you're totally right. I've looked up an article and it seems like we should only add the `!` suffix, if there's another method existing without the suffix 👍

lib/utility/errors.rb
seanstory: Typically, `ClientError` would be for expressing "I expected a 200, but got a 4xx or 5xx error". So you might have something like:

```
class NotFoundError < ClientError; end
```

to specifically cover 404s. But I expect that `InvalidFilterConfigError` won't have that kind of relationship. So yes, I think inheriting from StandardError is probably better unless I've misunderstood something here.

lib/connectors/mongodb/connector.rb
timgrein: In this case options refers to options, which can be passed to MongoDBs `find` or `aggregate` function, so it should live in the MongoDB connector.

lib/core/connector_settings.rb
timgrein: I think we should keep it concise with the rest of the class, which also specifies simple boolean values (f.e. `DEFAULT_REDUCE_WHITESPACE`) as constants.

lib/core/sync_job_runner.rb
timgrein: Will be removed with the next PR update.

lib/utility/strings.rb
timgrein: Will be removed with the next PR update.

lib/connectors/mongodb/connector.rb
timgrein: 👍

lib/connectors/mongodb/connector.rb
timgrein: 👍

spec/connectors/base/connector_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+      it_behaves_like 'filtering is not present'
+    end
+
+    context '.initialize' do
+      context 'three rules are present' do
+        let(:active_rules) {
+          [
+          { :name => 'rule one' },
+          { :name => 'rule two' },
+          { :name => 'rule three' },
```
seanstory, line 178: Just as a note, rules don't have `:name` properties as we've defined them. I think this test is fine as it is until we get to https://github.com/elastic/enterprise-search-team/issues/3168, but wanted to make sure we're on the same page that this isn't really representative right now.

spec/connectors/base/connector_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+require 'connectors/base/connector'
+
+describe Connectors::Base::Connector do
```
seanstory, line 11: praise: really love the test coverage

Keep this up! 👏

spec/connectors/mongodb/connector_spec.rb
```diff
@@ -292,5 +432,203 @@
         end
       end
     end
+
+    shared_examples_for 'logs a warning on yield documents' do
+      it 'logs a warning' do
```
seanstory, line 437: why?

To make this clear to readers in the future, maybe augment this with an expectation of what the warning log message should be.

spec/core/connector_settings_spec.rb
```diff
@@ -62,6 +62,153 @@
...
+        expect(first_filter[:active][:advanced_config]).to_not be_nil
+      end
+    end
+
+    context 'filtering is not present' do
+      it 'returns default filtering object' do
+        filtering = subject.filtering
+
+        expect(filtering).to_not be_nil
+        expect(filtering.empty?).to eq(true)
```
seanstory, line 96:
```suggestion
        expect(filtering).to be_empty
```

nit: this is slightly more idiomatic. See:

https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/predicate-matchers

spec/core/connector_settings_spec.rb
```diff
@@ -62,6 +62,153 @@
...
+              },
+              :draft => {
+                :active_rules => [],
+                :advanced_config => {}
+              }
+            }
+        }
+      }
+
+      it_behaves_like 'extracts active_filter_config'
```
seanstory, line 162: praise: I'm impressed that you've picked up using shared examples. This is a great tool that we underutilize, IMO

seanstory (commented):
This is shaping up nicely, I think. A lot of these comments are small, but let's chat in the connectors sync about the one more contentious point. Sorry we didn't iron that out before you started working 😅

lib/connectors/base/connector.rb
```diff
@@ -33,8 +32,13 @@ def self.service_type
         raise 'Not implemented for this connector'
       end
 
+      attr_reader :active_rules, :active_filter_config
+
       def initialize(configuration: {})
         @configuration = configuration.dup || {}
+
+        @active_rules = extract_active_rules(@configuration)
```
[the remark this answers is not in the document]
timgrein: Ah I see, then I got a bit confused with the terms. We can talk about it in the connector's sync. Then I would also prefer to pass it as an param to `yield_documents` 👍

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +60,114 @@ def initialize(configuration: {})
...
             view = cursor.skip(skip).limit(PAGE_SIZE)
             view.each do |document|
-              yield serialize(document)
-              found_count += 1
+              yield serialize(document) { on_doc_serialization.call if block_given? }
+
+              found_in_page += 1
+              found_overall += 1
+
+              overall_limit_reached = found_overall == overall_limit
```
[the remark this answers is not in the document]
timgrein: Good point, but I think this will only work, if we also check, that the `overall_limit` is not set -1 (flagging, that there's no overall limit set). Probably `Float::INFINITY` makes more sense semantically as a default value for the overall limit.

lib/utility/errors.rb
timgrein: Makes sense, updated it 👍

lib/connectors/mongodb/connector.rb
```diff
@@ -8,10 +8,24 @@
...
+require 'utility'
 
 module Connectors
   module MongoDB
     class Connector < Connectors::Base::Connector
+
+      ALLOWED_TOP_LEVEL_FILTER_KEYS = %w[find aggregate]
+
+      FIND = 'find'
+      AGGREGATE = 'aggregate'
```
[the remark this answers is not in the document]
timgrein: Done 👍

lib/connectors/mongodb/connector.rb
```diff
@@ -8,10 +8,24 @@
...
+      NO_ARGS = []
```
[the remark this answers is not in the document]
timgrein: Done 👍

lib/connectors/mongodb/connector.rb
```diff
@@ -8,10 +8,24 @@
...
+      DEFAULT_ADVANCED_CONFIG = {}
+      DEFAULT_RULES = []
+
+      EMPTY_PIPELINE = []
+      EMPTY_FILTER = {}
+      EMPTY_OPTIONS = {}
```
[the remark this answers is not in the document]
timgrein: Done 👍

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +70,166 @@ def initialize(configuration: {})
...
+        function, args = setup_db_function_with_args(advanced_filter_config)
+
+        call_db_function_on_collection(function, args, &on_doc_serialization)
+      end
+
+      private
+
+      def extract_rules(job_description)
+        Utility::Commons.return_if_present(job_description.dig(:filtering, :rules), DEFAULT_RULES)
+      end
```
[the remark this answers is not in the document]
timgrein: Lives in the base class now

lib/connectors/mongodb/connector.rb
timgrein: Replaced by `check_find_and_aggregate`.

lib/connectors/mongodb/connector.rb
timgrein: Replaced by `check_find_and_aggregate`. (`invalid_keys_msg` now also only generated, if needed)

lib/connectors/mongodb/connector.rb
timgrein: done 👍

lib/core/sync_job_runner.rb
timgrein: Actually this will probably be removed, when we change the return value by `claim_job` (when we return the full document, we can probably just pass it as is; added a comment to hint at this future change).

lib/utility/strings.rb
timgrein: Done 👍

lib/utility/strings.rb
timgrein: Done 👍

lib/connectors/base/connector.rb
```diff
@@ -33,11 +32,19 @@ def self.service_type
...
+        @job_description = job_description.dup || {}
+
+        filter = get_filter(@job_description[:filtering])
+
+        @rules = Utility::Common.return_if_present(filter[:rules], [])
+        @advanced_filter_config = Utility::Common.return_if_present(filter[:advanced_config], {})
       end
 
-      def yield_documents; end
+      def yield_documents(job_description: {}); end
```
seanstory, line 47:
```suggestion
      def yield_documents; end
```
Now that `@job_description` is available through the constructor, this is not necessary

lib/core/sync_job_runner.rb
```diff
@@ -52,14 +51,18 @@ def do_sync!
...
+
+        connector_instance.do_health_check!
 
         incoming_ids = []
         existing_ids = ElasticConnectorActions.fetch_document_ids(@connector_settings.index_name)
 
         Utility::Logger.debug("#{existing_ids.size} documents are present in index #{@connector_settings.index_name}.")
 
-        @connector_instance.yield_documents do |document|
+        connector_instance.yield_documents(job_description) do |document|
```
seanstory, line 65:
```suggestion
        connector_instance.yield_documents do |document|
```

job description now passed via the constructor

lib/connectors/mongodb/connector.rb
```diff
@@ -56,30 +60,112 @@ def initialize(configuration: {})
...
+
           loop do
-            found_count = 0
+            found_in_page = 0
+
             view = cursor.skip(skip).limit(PAGE_SIZE)
             view.each do |document|
-              yield serialize(document)
-              found_count += 1
+              yield serialize(document) { on_doc_serialization.call if block_given? }
```
seanstory, line 121: I _think_ this statement is redundant. I believe that you can remove all usages of `&on_doc_serialization` in parameter lists, and change this to:

```suggestion
              yield serialize(document)
```

lib/core/sync_job_runner.rb
```diff
@@ -52,14 +51,18 @@ def do_sync!
       begin
         Utility::Logger.debug("Successfully claimed job for connector #{@connector_settings.id}.")
 
-        @connector_instance.do_health_check!
+        # will be replaced when job claiming returns full job description object
```
seanstory, line 54: 👍 I'll tackle this in https://github.com/elastic/enterprise-search-team/issues/3149

seanstory (commented):
It's looking good! Looks like a few spots got missed in the changes after today's meeting, but after those, I think I'll be happy with it. 

@artem-shelkovnikov I don't want to dismiss your "requires changes" PR, but I do think this is getting to a state where it would be ok to merge as a first step for this project. Happy to consider any larger reorganization/cleanup in some of our follow-up PRs. WDYT?

lib/connectors/mongodb/connector.rb
timgrein: Actually, if I remove that statement 4 tests will fail (it seems like you cannot pass an implicit block two times? Or am I missing something obvious?). 

I think we need to pass the block explicitly to the `call_db_function_on_collection` to be able to `yield` it later. One other option would be to move all the code from `call_db_function_on_collection` directly into `yield_documents`. Then we can make use of an implicit block and we don't have to introduce the explicit block. I don't have a strong opinion here, but explicit blocks _always_ create a Proc object, which have a worse performance (source: [[Passing Blocks in Ruby Without &block](https://mudge.name/2011/01/26/passing-blocks-in-ruby-without-block/)]) We could use &Proc.new, but according to this article this option was removed in Ruby 3.0.0, so we probably run into the same issue again, if we use a version >= 3.0.0. 

WDYT?

lib/core/sync_job_runner.rb
timgrein: 👍

artem-shelkovnikov:
> @artem-shelkovnikov I don't want to dismiss your "requires changes" PR, but I do think this is getting to a state where it would be ok to merge as a first step for this project. Happy to consider any larger reorganization/cleanup in some of our follow-up PRs. WDYT?

@seanstory totally agreed, I've missed a bit the discussion in the thread but it looks like a solid PR. I've spoke to @timgrein about the `yield` change, gonna take a look and put my stamp in a bit, sorry for the delay.

artem-shelkovnikov (approved):
LGTM, thanks for the great work!

seanstory (approved):

lib/connectors/registry.rb
```diff
@@ -24,10 +24,10 @@ def connector_class(name)
       @connectors[name]
     end
 
-    def connector(name, configuration)
+    def connector(name, configuration, job_description)
       klass = connector_class(name)
       if klass.present?
-        return klass.new(configuration: configuration)
+        return klass.new(configuration: configuration, job_description: job_description)
```
wangch079, line 30: I think this change breaks many places:
https://github.com/elastic/connectors-ruby/blob/87a9c8aad96b798bf7b45667cb3fac15fa0b0ccf/lib/app/console_app.rb#L170

https://github.com/elastic/connectors-ruby/blob/87a9c8aad96b798bf7b45667cb3fac15fa0b0ccf/lib/core/heartbeat.rb#L22

PR #386, opened 2022-10-31
Add metadata to connector protocol and job index mapping

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/3193

This PR adds generic and connector-specific metadata to connector protocol and job index mapping

### DIFF
--- a/docs/CONNECTOR_PROTOCOL.md
+++ b/docs/CONNECTOR_PROTOCOL.md
@@ -228,6 +228,7 @@ In addition to the connector index `.elastic-connectors`, we have an additional 
   cancelation_requested_at: date; -> The date/time when the cancelation of the job is requested
   canceled_at: date; -> The date/time when the job is canceled
   completed_at: date; -> The data/time when the job is completed
+  configuration: object; -> Connector configuration
   connector_id: string; -> ID of the connector document in .elastic-connectors
   created_at: date; -> The date/time when the job is created
   deleted_document_count: number; -> Number of documents deleted in the job
@@ -259,8 +260,15 @@ In addition to the connector index `.elastic-connectors`, we have an additional 
   indexed_document_volume: number; -> The volume (in bytes) of documents indexed in the job
   last_seen: date; -> Connector writes check-in date-time regularly (UTC)
   metadata: object; -> Connector-specific metadata
+  pipeline: { -> Connector pipeline
+    extract_binary_content: boolean;
+    name: string;
+    reduce_whitespace: boolean;
+    run_ml_inference: boolean;
+  },
   started_at: date; -> The date/time when the job is started
   status: string; -> Job status Enum, see below
+  total_document_count: number; -> Number of documents in the index after the job completes
   trigger_method: string; -> How the job is triggered. Possible values are on-demand, scheduled.
   worker_hostname: string; -> The hostname of the worker to run the job
 }
@@ -285,6 +293,7 @@ In addition to the connector index `.elastic-connectors`, we have an additional 
     "cancelation_requested_at" : { "type" : "date" },
     "canceled_at" : { "type" : "date" },
     "completed_at" : { "type" : "date" },
+    "configuration" : { "type" : "object" },
     "connector_id" : { "type" : "keyword" },
     "created_at" : { "type" : "date" },
     "deleted_document_count" : { "type" : "integer" },
@@ -324,8 +333,17 @@ In addition to the connector index `.elastic-connectors`, we have an additional 
     "indexed_document_volume" : { "type" : "integer" },
     "last_seen" : { "type" : "date" },
     "metadata" : { "type" : "object" },
+    "pipeline" : {
+      "properties" : {
+        "extract_binary_content" : { "type" : "boolean" },
+        "name" : { "type" : "keyword" },
+        "reduce_whitespace" : { "type" : "boolean" },
+        "run_ml_inference" : { "type" : "boolean" }
+      }
+    },
     "started_at" : { "type" : "date" },
     "status" : { "type" : "keyword" },
+    "total_document_count" : { "type" : "integer" },
     "trigger_method" : { "type" : "keyword" },
     "worker_hostname" : { "type" : "keyword" }
   }
--- a/docs/CONNECTOR_PROTOCOL.md
+++ b/docs/CONNECTOR_PROTOCOL.md
@@ -71,13 +71,12 @@ This is our main communication index, used to communicate the connector's config
   index_name: string;   -> The name of the content index where data will be written to
   is_native: boolean;   -> Whether this is a native connector
   language: string;     -> the language used for the analyzer
-  last_seen: date;      -> Connector writes check-in date-time
-                           regularly (UTC)
+  last_seen: date;      -> Connector writes check-in date-time regularly (UTC)
   last_sync_error: string;   -> Optional last job error message
   last_sync_status: string;  -> Status of the last job, or null if no job has been executed
   last_synced: date;    -> Date/time of last job (UTC)
-  last_indexed_document_count: number;    -> How many documents were indexed in the last job
   last_deleted_document_count: number;    -> How many documents were deleted in the last job
+  last_indexed_document_count: number;    -> How many documents were indexed in the last job
   name: string; -> the name to use for the connector
   pipeline: {
     extract_binary_content: boolean; -> Whether the `request_pipeline` should handle binary data
@@ -198,8 +197,8 @@ This is our main communication index, used to communicate the connector's config
     "last_sync_error" : { "type" : "keyword" },
     "last_sync_status" : { "type" : "keyword" },
     "last_synced" : { "type" : "date" },
-    "last_indexed_document_count" : { "type" : "long" },
     "last_deleted_document_count" : { "type" : "long" },
+    "last_indexed_document_count" : { "type" : "long" },
     "name" : { "type" : "keyword" },
     "pipeline" : {
       "properties" : {
@@ -226,10 +225,13 @@ This is our main communication index, used to communicate the connector's config
 In addition to the connector index `.elastic-connectors`, we have an additional index to log all jobs run by connectors. This is the `.elastic-connectors-sync-jobs` index. Each JSON document will have the following structure:
 ```
 {
+  cancelation_requested_at: date; -> The date/time when the cancelation of the job is requested
+  canceled_at: date; -> The date/time when the job is canceled
+  completed_at: date; -> The data/time when the job is completed
   connector_id: string; -> ID of the connector document in .elastic-connectors
-  status: string; -> Job status Enum, see below
+  created_at: date; -> The date/time when the job is created
+  deleted_document_count: number; -> Number of documents deleted in the job
   error: string; -> Optional error message
-  trigger_method: string; -> How the job is triggered. Possible values are on-demand, scheduled.
   filtering: {          -> Filtering rules
     domain: string,     -> what data domain these rules apply to
     rules: {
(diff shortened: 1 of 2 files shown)


docs/CONNECTOR_PROTOCOL.md
```diff
@@ -125,12 +125,19 @@ In addition to the connector index `.elastic-connectors`, we have an additional
...
-  worker_hostname: string; -> The hostname of the worker to run the sync job
-  indexed_document_count: number; -> Number of documents indexed in the sync job
-  deleted_document_count: number; -> Number of documents deleted in the sync job
-  created_at: date; -> The date/time when the sync job is created
-  completed_at: date; -> The data/time when the sync job is completed
+  worker_hostname: string; -> The hostname of the worker to run the job
+  indexed_document_count: number; -> Number of documents indexed in the job
+  indexed_document_volume: number; -> The volume (in bytes) of documents indexed in the job
+  deleted_document_count: number; -> Number of documents deleted in the job
+  metadata: object; -> Connector-specific metadata
```
wangch079, line 134: Is it better to call it `metadata` or `stats`?

seanstory: Both are pretty general terms. Can we come up with something more specific and descriptive? What are some examples that will go in here?

wangch079: It's connector specific. 

e.g. A crawler metadata can be:
1. crawl type
2. max crawl depth
3. crawl duration
4. number of crawled urls
5. number of crawled pages
6. avg response time
7. number of 4xx error
8. number of 5xx error
9. list of domains/seed urls/sitemap urls.

Confluence metadata can be:
1. number of spaces indexed
2. number of pages indexed
3. number of blogs indexed
4. number of attachments indexed
5. current processing space
6. number of 4xx error
7. number of 5xx error
8. etc..

seanstory: Most of these look numeric. What about a `counters` object? And let's use as generic of key names for the counters as possible. Kibana will not care about the difference between "number of crawled pages" and "total number of confluence documents indexed", but will want to display both in the same place, I expect. So `num_documents` might be better.

For the non-numerics:
- "crawl type" could be a generic `job_type` field, and we'll probably eventually want that for other service types as well, as we'll want to display it in Kibana. Keeping the field name consistent for this will make our lives a lot easier.
- "list of domains/seed urls/sitemap urls" this sounds like configuration, which @vidok will be working to add as part of his work, so you probably don't need to worry about here

This kind of design work and iteration might be easier to do in the implementation document, in the future.

sphilipse: I would love to see those properties actually mapped as properties as well, rather than mapping the whole thing as 'object'. That way we could actually execute efficient searches on these properties--things like, "give me all jobs with more than 10 4xx" errors. That's a lot harder if we don't map these.

wangch079: > So `num_documents` might be better.

> "crawl type" could be a generic `job_type` field

> "list of domains/seed urls/sitemap urls" this sounds like configuration

> I would love to see those properties actually mapped as properties as well

Hi @seanstory @sphilipse , the generic stats are defined at the root level of `.elastic-connectors-sync-jobs`, e.g. `indexed_document_count`. All those example fields above are connector-specific, which can be from Crawler, or Confluence, or Dropbox. We have no way to know what are the available fields in advance, and it's up to connector implementation.

>This kind of design work and iteration might be easier to do in the implementation document, in the future.

Yes, it was designed to be a new field `metadata` of type `object. I'm just not sure of the name

> Most of these look numeric. What about a `counters` object?

`counters` implies it should be all numbers? so does `stats`.

seanstory: > `counters` implies it should be all numbers? so does `stats`.

True. I feel that "stats" conflates a bit with system-level information - like memory and cpu usage, threadcount, idle threads, etc, I would expect to find under "stats". "Counters" on the other hand is where I have experience of ETL frameworks tracking how many of X have been seen, and also how much time has been spent on Y. But it's just a suggestion.

sphilipse: > ll those example fields above are connector-specific, which can be from Crawler, or Confluence, or Dropbox. We have no way to know what are the available fields in advance, and it's up to connector implementation.

We have some idea right now, though. We know what metadata our current connectors will report. If we have some way of generalising over that kind of information that would be helpful.

wangch079: >We know what metadata our current connectors will report. If we have some way of generalising over that kind of information that would be helpful.

This is still in [discussion](https://elastic.slack.com/archives/C03GAPTQ2JF/p1667375423236469), but even there is some info we can generalize, we should make it one of the general metadata like `indexed_document_count`, which should reside at the root level. 

Here, the `metadata` is for connector-specific metadata, which can be anything, similar to `configuration` in `.elastic-connectors`. I'm just not sure about the name.

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -258,11 +261,14 @@ In addition to the connector index `.elastic-connectors`, we have an additional
     }
   };
   index_name: string; -> The name of the content index
-  worker_hostname: string; -> The hostname of the worker to run the sync job
-  indexed_document_count: number; -> Number of documents indexed in the sync job
-  deleted_document_count: number; -> Number of documents deleted in the sync job
-  created_at: date; -> The date/time when the sync job is created
-  completed_at: date; -> The data/time when the sync job is completed
+  indexed_document_count: number; -> Number of documents indexed in the job
+  indexed_document_volume: number; -> The volume (in bytes) of documents indexed in the job
```
wangch079, line 265: Hi @sphilipse , do you need `last_indexed_document_volume` in `.elastic-connectors`?

docs/CONNECTOR_PROTOCOL.md
sphilipse: Honestly given all that 'metadata' is probably the best term at this point.

docs/CONNECTOR_PROTOCOL.md
sphilipse: We don't have an immediate use for it.

sphilipse:
One more note: would be great to get the `pipeline` fields in here as well, so we know per job which pipelines to apply (and which have been applied).

wangch079:
>One more note: would be great to get the pipeline block from .elastic-connectors in here as well, so we know per job which pipelines to apply (and which have been applied).

I'm also going to include `configuration` in `.elastic-connectors-sync-jobs`

sphilipse:
> > One more note: would be great to get the pipeline block from .elastic-connectors in here as well, so we know per job which pipelines to apply (and which have been applied).
> 
> I'm also going to include `configuration` in `.elastic-connectors-sync-jobs`

So to clarify both `configuration` and `pipeline` will be in there?

wangch079:
> So to clarify both `configuration` and `pipeline` will be in there?

Yes

sphilipse:
Another question: we had discussed several times having a list of errors encountered during the sync in the sync job documents. Is that no longer the case?

wangch079:
> Another question: we had discussed several times having a list of errors encountered during the sync in the sync job documents. Is that no longer the case?

This is not the plan in the Monitoring project. I think we will just record a single error when a job fails. I know @artem-shelkovnikov is working on error tolerance in the Resilience project, is this in your plan to record all errors encountered during a job execution, Artem?

lib/core/elastic_connector_actions.rb
```diff
@@ -235,12 +235,99 @@ def ensure_connectors_index_exists
           :properties => {
             :api_key_id => { :type => :keyword },
             :configuration => { :type => :object },
-            :error => { :type => :text },
+            :description => { :type => :text },
```
seanstory, line 238: TIL that this function exists. Sorry I missed this for filtering rules. I don't love that we're maintaining this in yet _another_ place, but this isn't the time to change that.

seanstory (approved):

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -252,11 +256,21 @@ In addition to the connector index `.elastic-connectors`, we have an additional
...
   index_name: string; -> The name of the content index
-  worker_hostname: string; -> The hostname of the worker to run the job
   indexed_document_count: number; -> Number of documents indexed in the job
-  deleted_document_count: number; -> Number of documents deleted in the job
-  created_at: date; -> The date/time when the job is created
-  completed_at: date; -> The data/time when the job is completed
+  indexed_document_volume: number; -> The volume (in bytes) of documents indexed in the job
+  last_seen: date; -> Connector writes check-in date-time regularly (UTC)
+  metadata: object; -> Connector-specific metadata
+  pipeline: { -> Connector pipeline
```
vidok, line 263: I wonder if you'd considered to put `pipeline` and other job's configurations option under `properties` object

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -276,7 +290,10 @@ In addition to the connector index `.elastic-connectors`, we have an additional
     "version" : 1
   },
   "properties" : {
+    "cancelation_requested_at" : { "type" : "date" },
```
vidok, line 293: I wonder why you need this property. Could you please share more information about that?

docs/CONNECTOR_PROTOCOL.md
seanstory: Lets not use `properties` just so it doesn't get confusing with Elasticsearch where `properties` is used to describe sub-object fields 😓 

To be fair to Chenhui, he'd argued in another thread to use `connector` as the outer key to group all this stuff under. I'd noted that Sander has expressed wanting to keep things flatter, and I (when I added `filtering`) didn't want to nest it by itself. Chenhui, feel free to blame this on me and say, "I told you so!" :D

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -276,7 +290,10 @@ In addition to the connector index `.elastic-connectors`, we have an additional
     "version" : 1
   },
   "properties" : {
```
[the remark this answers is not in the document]
wangch079: Hi @vidok , were you referring to `_meta`?

We do have job configurations, e.g. `job_reporting_interval` - how often the connector should report metadata. But they are going to be in the `_meta` of `.elastic-connectors`.

docs/CONNECTOR_PROTOCOL.md
wangch079: This is explained above:
>cancelation_requested_at: date; -> The date/time when the cancelation of the job is requested
  canceled_at: date; -> The date/time when the job is canceled

It's the date/time when an operator clicks `Cancel` in Kibana. `canceled_at` is the date/time the job is actually canceled by the connector.

Free feel to propose a new name.

docs/CONNECTOR_PROTOCOL.md
wangch079: It won't bring any performance benefits, eventually, they will be flattened. But I'm still in favor of a grouped object, which makes it much easier to read, especially now all the fields are alphabetically ordered.

artem-shelkovnikov:
> is this in your plan to record all errors encountered during a job execution, Artem?
The plan for now is to move the code from Workplace Search and what it does is - it will generate one error that will internally point I think to the last error encountered. 

We can modify this logic for future, but for now there's gonna be only one error in connectors index, all errors are available via logs if user needs it.

docs/CONNECTOR_PROTOCOL.md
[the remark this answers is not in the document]
vidok: I think this data should be also copied into jobs documents when you create one.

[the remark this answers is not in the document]
vidok: > Hi @vidok , were you referring to `_meta`?

No, I'm referring to `properties`

docs/CONNECTOR_PROTOCOL.md
vidok: Thanks. My questions was about where are you going to use that property? Is it gonna be used in the code? Or maybe in Kibana UI?

wangch079: It's not going to be used in the connector code. I think just for audit purposes, that we know when it's requested. And I think this will also be used in Kibana to display events, until we have events generator, similar to Crawler.

docs/CONNECTOR_PROTOCOL.md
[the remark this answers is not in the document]
wangch079: Hi @vidok , just realized `job_reporting_interval` is hardcoded and won't be made configurable in 8.6, so I think we don't need to do it now, and we can iterate in a later release.

docs/CONNECTOR_PROTOCOL.md
vidok: I would not add it if it's not required. This really sounds like a good use case for logging. 

I wonder if we can retrieve this information from Kibana server log for example.

wangch079: >I would not add it if it's not required.

It's required, see [here](https://docs.google.com/document/d/1mqPTc3Agg7FaUPVZ0YR4_R3M552HjqFwAL1UJvi57Ek/edit#bookmark=id.gd6hz8otbhqh). I meant it's not used by backend code (It's not set by BE, it's not used by BE either).

PR #389, opened 2022-11-01
Implement dynamic elasticsearch data flushing

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/3096

This PR changes the way data is sent into Elasticsearch.

Before we were just ingesting data in batches of same sizes (of size 50), now we have 1 more parameter - maximum payload size of bulk request.

So for example, with current settings the batch is sent to Elasticsearch if the number of operations reaches 500 OR the size of the payload reaches 5MB.

#### Pre-Review Checklist

## Release Note

Improved logic of batching the data when sending it to Elasticsearch - now size of the payloads is taken into account, preventing failed bulk requests because the payload was too big.


lib/utility.rb
```diff
@@ -4,14 +4,15 @@
...
 
+require 'utility/bulk_queue'
+require 'utility/common'
 require 'utility/constants'
 require 'utility/cron'
-require 'utility/common'
+require 'utility/elasticsearch/index/mappings'
+require 'utility/elasticsearch/index/text_analysis_settings'
+require 'utility/environment'
 require 'utility/errors'
```
artem-shelkovnikov, line 14: I've also done sorting, thus the bigger change.

Real change is this:

```
require 'utility/bulk_queue'
```

lib/utility/bulk_queue.rb
```diff
@@ -0,0 +1,85 @@
...
+#
+
+require 'json'
+
+module Utility
+  class BulkQueue
+    class QueueOverflowError < StandardError; end
+
+    # 500 items or 5MB
+    def initialize(count_threshold = 500, size_threshold = 5 * 1024 * 1024)
```
artem-shelkovnikov, line 14: There will be a follow-up to allow overriding the settings via the config

lib/utility/bulk_queue.rb
```diff
@@ -0,0 +1,85 @@
...
+      @buffer << "\n"
+
+      if payload
+        @buffer << payload
+        @buffer << "\n"
+      end
+    end
+
+    def will_fit?(operation, payload = nil)
+      return false if @current_op_count + 1 > @count_threshold
```
artem-shelkovnikov, line 55: This will for now fail when document is too big to ingest. I will provide a follow-up PR that will silently ignore documents that are too big.

artem-shelkovnikov:
I've also tried to keep the change small, but it was pretty hard :/ Most of the changeset is tests though.

spec/core/output_sink/base_sink_spec.rb
```diff
@@ -2,12 +2,162 @@
...
+
+  context 'when all private methods are implemented' do
+    before(:each) do
+      allow(subject).to receive(:do_ingest)
+      allow(subject).to receive(:do_delete)
+      allow(subject).to receive(:do_flush)
+      allow(subject).to receive(:do_serialize)
+    end
+
+    context '#ingest' do
```
vidok, line 24: [nitpick] It's better to use `#describe` when you start testing a method. 

refs: https://www.betterspecs.org

spec/core/output_sink/base_sink_spec.rb
```diff
@@ -2,12 +2,162 @@
...
+
+        it 'calls do_ingest on serialized document' do
+          expect(subject).to receive(:do_ingest).with(document_id, serialized_document)
+
+          subject.ingest(document)
+        end
+      end
+    end
+
+    context '#ingest_multiple' do
```
vidok, line 52: [nitpick] It's better to use #describe when you start testing a method.

refs: https://www.betterspecs.org/

spec/core/output_sink/base_sink_spec.rb
```diff
@@ -2,12 +2,162 @@
...
+      it 'calls ingest on each ingested document' do
+        expect(subject).to receive(:ingest).with(document1)
+        expect(subject).to receive(:ingest).with(document2)
+        expect(subject).to receive(:ingest).with(document3)
+
+        subject.ingest_multiple([document1, document2, document3])
+      end
+    end
+
+    context '#delete' do
```
vidok, line 66: [nitpick] It's better to use #describe when you start testing a method.

refs: https://www.betterspecs.org/

spec/core/output_sink/base_sink_spec.rb
```diff
@@ -2,12 +2,162 @@
...
+
+        it 'calls do_delete with expected id' do
+          expect(subject).to receive(:do_delete).with(deleted_id)
+
+          subject.delete(deleted_id)
+        end
+      end
+    end
+
+    context '#delete_multiple' do
```
vidok, line 88: [nitpick] It's better to use #describe when you start testing a method.

refs: https://www.betterspecs.org/

spec/core/output_sink/base_sink_spec.rb
```diff
@@ -2,12 +2,162 @@
...
+      it 'calls ingest on each ingested document' do
+        expect(subject).to receive(:delete).with(id1)
+        expect(subject).to receive(:delete).with(id2)
+        expect(subject).to receive(:delete).with(id3)
+
+        subject.delete_multiple([id1, id2, id3])
+      end
+    end
+
+    context '#flush' do
```
vidok, line 102: [nitpick] It's better to use #describe when you start testing a method.

refs: https://www.betterspecs.org/

spec/core/output_sink/base_sink_spec.rb
```diff
@@ -2,12 +2,162 @@
...
+      it 'calls do_flush' do
+        expect(subject).to receive(:do_flush)
+
+        subject.flush
+      end
+    end
+
+    context '#ingestion_stats' do
```
vidok, line 110: [nitpick] It's better to use #describe when you start testing a method.

refs: https://www.betterspecs.org/

spec/utility/bulk_queue_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+#
+
+require 'utility/bulk_queue'
+
+describe Utility::BulkQueue do
+  let(:subject) { described_class.new(count_threshold, size_threshold) }
+  let(:count_threshold) { 50 }
+  let(:size_threshold) { 999999999 } # super high default
+
+  context '.add' do
```
vidok, line 14: [nitpick] It's better to use #describe when you start testing a method.

refs: https://www.betterspecs.org/

spec/utility/bulk_queue_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+
+        buffer = subject.pop_all
+
+        expect(buffer).to include(op)
+        expect(buffer).to include(payload)
+      end
+    end
+  end
+
+  context '.will_fit?' do
```
vidok, line 56: [nitpick] It's better to use #describe when you start testing a method.

refs: https://www.betterspecs.org/

spec/utility/bulk_queue_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+        end
+      end
+
+      it 'returns false' do
+        expect(subject.will_fit?(big_operation)).to eq(false)
+      end
+    end
+  end
+
+  context '.pop_all' do
```
vidok, line 102: [nitpick] It's better to use #describe when you start testing a method.

refs: https://www.betterspecs.org/

spec/utility/bulk_queue_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+
+      it 'cleans up the queue' do
+        subject.pop_all
+
+        expect(subject.pop_all).to eq('')
+      end
+    end
+  end
+
+  context '.current_stats' do
```
vidok, line 124: [nitpick] It's better to use #describe when you start testing a method.

refs: https://www.betterspecs.org/

lib/core/output_sink/base_sink.rb
```diff
@@ -9,24 +9,66 @@
...
+      def initialize
+        @ingested_count = 0
+        @ingested_volume = 0
+        @deleted_count = 0
       end
 
-      def ingest_multiple(_documents)
-        raise 'not implemented'
+      def ingest(document)
+        return if document.blank?
```
vidok, line 19: I guess `empty?` would also work because the method expects a JSON object (afaik). 

#lets-remove-active_support :)

lib/utility/bulk_queue.rb
vidok, line 14: Is `count_threshold` a operations number?

lib/core/output_sink/es_sink.rb
```diff
@@ -9,66 +9,53 @@
...
     end
 
-    def ingest(document)
-      return if document.blank?
+    private
 
-      @operation_queue << { :index => { :_index => index_name, :_id => document[:id], :data => document } }
-      flush if ready_to_flush?
-    end
+    def do_ingest(id, serialized_document)
```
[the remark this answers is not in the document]
artem-shelkovnikov: I took this pattern from previous C# experience when a public method of the class is a VERB, and an abstract method it's calling is DO_VERB, e.g:

```C#
class ElasticsearchQueryBase
{
  IEnumerable<ElasticsearchObject> Query(ElasticsearchQuery query) {
    // do something
    return DoQuery(query)
  }
}
```

I'm not sure about the naming, do you have suggestions? I'm not happy with it either at the level that I even thought to switch to composition instead of inheritance and rename methods to something completely different.

spec/core/output_sink/base_sink_spec.rb
artem-shelkovnikov: Good call, but I'll make it in a different PR for all tests in ruby project - they all don't follow this. Also some `#` or `.` in method prefixes don't make much sense, I'll fix that in a separate PR as well.

lib/core/output_sink/base_sink.rb
artem-shelkovnikov: Lol, I'm all up to remove ActiveSupport.

It's good to handle `nil` anyway cause the document that gets here is coming from the connector, and we cannot prevent the author of the connector from calling `yield nil`. I'll update this method to warn that nil/empty object is sent, I think it's a good idea.

lib/utility/bulk_queue.rb
artem-shelkovnikov: Yup, maximum number of operations, will `operation_count_threshold` sound better?

vidok: Yes, I think it's better

lib/core/output_sink/es_sink.rb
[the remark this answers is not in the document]
vidok: Now, I think that the naming issue is follows from inheritance pattern. 
I see a several of issues: 
1. Calling `super` in "#initialize" might be not obvious for anyone who wants to create a new sink. 
2. `Base` sink is not expandable because all you can do is change the methods behaviour
3. Some of the methods (`#delete_multiple`) might not be working for other sinks (I don't know if we anticipate other sinks existence)

I see 2 different ways to solve that: 
`Super simple`: Approve and merge because it's not a blocker :)
`Simple`: Get rid of `Base` sink and move its logic into `EsSink` (`Console` and `Output` sinks will have some duplicated code which is okay)
`Complex`: Apply `Strategy` pattern: 

```ruby
class Sink
  def initialize(sink:)
   @sink = sink
  end

  def ingest(document)
   return if document.blank?

   serialized_document = @sink_type.serialize(document)
   @sink.ingest(document['id'], serialized_document)

   increment_ingested_count
   increment_ingested_volume(serialized_document.bytesize)
 end

 def delete(id)
   return if id.nil?
   
   @sink.delete(id)

   increment_deleted_count
 end
end

......
```

When you use it in connector service
```
sink = Sink.new(sink: EsSink.new)
sink.ingest(document)
```

I like the 3 option the most but it's not a blocker (Assuming you will need to rewrite tests)

lib/core/output_sink/base_sink.rb
artem-shelkovnikov: Ha, never seen it, thanks for sharing! I'll give it a look :)

lib/core/output_sink/es_sink.rb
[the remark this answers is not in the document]
artem-shelkovnikov: @vidok I see, I've attempted to address this with a bit different naming:

```ruby
sink = Sink.new(EsStrategy.new)
sink.ingest(document)
```

I'm not a big fan of this naming though (EsStrategy? meh EsSink? Not sure either). What do you think?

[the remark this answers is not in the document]
vidok: `EsSink` doesn't hurt as much as `EsStrategy` does :), especially if you use named arguments. 

```Ruby
Sink.new(sink: EsSink.new)
```

or 
```Ruby
Sink::Elasticsearch
Sink::Stdout
Sink::Console
```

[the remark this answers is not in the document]
artem-shelkovnikov: I find it a bit weird about creating a Sink that has another Sink inside, maybe the outer sink should be called differently?

```ruby
Ingester.new(sink: Sink::Elasticsearch.new)
```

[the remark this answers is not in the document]
vidok: > I find it a bit weird about creating a Sink that has another Sink inside
agree

> Ingester.new(sink: Sink::Elasticsearch.new)
Yes, why not.

vidok (approved):
great job🥇

lib/core/ingestion/ingester.rb
```diff
@@ -0,0 +1,81 @@
...
+        @deleted_count = 0
+      end
+
+      def ingest(document)
+        unless document&.any?
+          Utility::Logger.warn('Connector attempted to ingest an empty document, skipping')
+          return
+        end
+
+        serialized_document = @sink_strategy.serialize(document)
```
wangch079, line 27: I think we should add a base class for `ConsoleSink` and `EsSink` to make sure `serialize` method is implemented, as well as `ingest` and `delete`

lib/core/ingestion/ingester.rb
```diff
@@ -0,0 +1,81 @@
...
+        @sink_strategy.ingest(document['id'], serialized_document)
+
+        @ingested_count += 1
+        @ingested_volume += serialized_document.bytesize
```
wangch079, line 31: 🤔 You add another layer `Ingester` above `Sink`. 

I think one of the benefits of having ingestion stats in Sink, is that we could have more accurate data. E.g. we could increment @ingested_count when flush is called. But I see you serialize everything (operation, payload) in the queue, so it's not easy to count the docs and volume anyway.

lib/core/ingestion/ingester.rb
```diff
@@ -0,0 +1,81 @@
...
+        raise NotImplementedError
+      end
+
+      def do_flush
+        raise NotImplementedError
+      end
+
+      def do_serialize(_document)
+        raise NotImplementedError
+      end
```
wangch079, line 78: Where are those methods implemented? Don't see any classes inheriting `Ingester`.

lib/core/sync_job_runner.rb
```diff
@@ -23,10 +23,16 @@ class SyncJobRunner
...
+      @ingester = Core::Ingestion::Ingester.new(Core::Ingestion::EsSink.new(connector_settings.index_name, @connector_settings.request_pipeline))
       @connector_class = Connectors::REGISTRY.connector_class(connector_settings.service_type)
       @sync_finished = false
       @sync_error = nil
+      @status = {
+        :indexed_document_count => 0,
+        :deleted_document_count => 0,
+        :indexed_document_volume => 0,
+        :error => nil
+      }
```
wangch079, line 35: I don't think we need this

lib/utility/bulk_queue.rb
```diff
@@ -0,0 +1,85 @@
...
+require 'json'
+
+module Utility
+  class BulkQueue
+    class QueueOverflowError < StandardError; end
+
+    # 500 items or 5MB
+    def initialize(operation_count_threshold = 500, size_threshold = 5 * 1024 * 1024)
+      @operation_count_threshold = operation_count_threshold.freeze
+      @size_threshold = size_threshold.freeze
```
wangch079, line 16: Don't think the `freeze` has any effect.

lib/core/ingestion/es_sink.rb
```diff
@@ -0,0 +1,59 @@
...
+        @request_pipeline = request_pipeline
+        @operation_queue = bulk_queue
+      end
+
+      def ingest(id, serialized_document)
+        index_op = serialize({ 'index' => { '_index' => index_name, '_id' => id } })
+
+        flush unless @operation_queue.will_fit?(index_op, serialized_document)
+
+        @operation_queue.add(
```
wangch079, line 30: If the size of a single doc is over the size threshold, this will throw an overflow error

wangch079 (commented):
Sorry for the late review

lib/core/ingestion/ingester.rb
artem-shelkovnikov: These are leftovers, gonna delete them, thanks for noticing!

lib/core/sync_job_runner.rb
artem-shelkovnikov: It was left here after a rebase, what's up with these lines?

lib/utility/bulk_queue.rb
artem-shelkovnikov: It won't allow modification of variables, or not?

lib/core/ingestion/ingester.rb
artem-shelkovnikov: At this point I feel like dropping ConsoleSink and BaseSink and just leaving the class that does all ingestion operations. ConsoleSink is not used and makes refactoring harder.

lib/core/ingestion/ingester.rb
artem-shelkovnikov: I see, I've overlooked this, I'll address in a separate PR

lib/core/sync_job_runner.rb
wangch079: Previously we use `@status` because we want to `@status[:indexed_document_count] +=1` for every yielded document. But now, we will just use `@ingester.ingestion_stats.merge(:metadata => connector_instance.metadata)`.

I will make the change in my PR.

lib/core/ingestion/es_sink.rb
wangch079: I think oversized doc should be discarded.

lib/utility/bulk_queue.rb
wangch079: Both are integers and I think they are immutable.

PR #390, opened 2022-11-01
Report metadata during job execution

wangch079:
#### Pre-Review Checklist

### DIFF
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -323,6 +323,10 @@ module Core
         end
       end
 
+      def document_count(index_name)
+        client.count(:index => index_name)['count']
+      end
+
       private
 
       def client
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -115,6 +115,7 @@ module Core
           :status => Connectors::SyncStatus::IN_PROGRESS,
           :worker_hostname => Socket.gethostname,
           :created_at => Time.now,
+          :last_seen => Time.now,
           :filtering => convert_connector_filtering_to_job_filtering(connector_record.dig('_source', 'filtering'))
         }
 
@@ -147,7 +148,7 @@ module Core
 
       def update_sync(job_id, metadata)
         body = {
-          :doc => metadata
+          :doc => { :last_seen => Time.now }.merge(metadata)
         }
         client.update(:index => Utility::Constants::JOB_INDEX, :id => job_id, :body => body)
       end
@@ -167,6 +168,7 @@ module Core
           :doc => {
             :status => sync_status,
             :completed_at => Time.now,
+            :last_seen => Time.now,
             :error => error
           }.merge(metadata)
         }
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -115,6 +115,7 @@ module Core
           :status => Connectors::SyncStatus::IN_PROGRESS,
           :worker_hostname => Socket.gethostname,
           :created_at => Time.now,
+          :started_at => Time.now,
           :last_seen => Time.now,
           :filtering => convert_connector_filtering_to_job_filtering(connector_record.dig('_source', 'filtering'))
         }
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -146,14 +146,14 @@ module Core
         update_connector_fields(connector_id, body)
       end
 
-      def update_sync(job_id, status)
+      def update_sync(job_id, metadata)
         body = {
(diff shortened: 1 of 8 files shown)


lib/core/output_sink/es_sink.rb
```diff
@@ -56,6 +56,10 @@ def delete_multiple(ids)
       ids.each { |id| delete(id) }
     end
 
+    def ingestion_stats
+      {}
```
wangch079, line 60: To be implemented by @artem-shelkovnikov  in https://github.com/elastic/connectors-ruby/pull/389

lib/core/sync_job_runner.rb
```diff
@@ -81,22 +87,21 @@ def do_sync!
         # occurred as most of the time the main execution thread is interrupted and we miss this Signal/Exception here
         @sync_finished = true
       rescue StandardError => e
-        @status[:error] = e.message
+        @sync_error = e.message
         Utility::ExceptionTracking.log_exception(e)
-        ElasticConnectorActions.update_connector_status(@connector_settings.id, Connectors::ConnectorStatus::ERROR, Utility::Logger.abbreviated_message(e.message))
```
wangch079, line 86: Hi @mchernyavskaya , `git blame` told me you added this line, but this is covered by `complete_sync` in the `ensure` block.

lib/core/elastic_connector_actions.rb
```diff
@@ -132,22 +132,30 @@ def update_connector_status(connector_id, status, error_message = nil)
         update_connector_fields(connector_id, body)
       end
 
-      def complete_sync(connector_id, job_id, status)
-        sync_status = status[:error] ? Connectors::SyncStatus::ERROR : Connectors::SyncStatus::COMPLETED
+      def update_sync(job_id, metadata)
```
wangch079, line 135: `update_sync` method will only update the job index, and `last_indexed_document_count` and `last_deleted_document_count` are only updated in the connector index in `complete_sync` method. cc @sphilipse

lib/core/sync_job_runner.rb
```diff
@@ -59,11 +56,17 @@ def do_sync!
...
         Utility::Logger.debug("#{existing_ids.size} documents are present in index #{@connector_settings.index_name}.")
 
-        @connector_instance.yield_documents do |document|
+        reporting_cycle_start = Time.now
+        @connector_instance.yield_documents do |document, connector_metadata = {}|
           document = add_ingest_metadata(document)
           @sink.ingest(document)
           incoming_ids << document['id']
-          @status[:indexed_document_count] += 1
+          @connector_metadata = connector_metadata
```
seanstory, line 64: This re-assigns an instance variable every time a single document is emitted. That feels odd to me, that we aren't aggregating that metadata, but totally re-assigning it each time. If we wanted it fresh, would it not make more sense to add a `def metadata` function to the connector, so that we could just call `@connector_instance.metadata` whenever we wanted, rather than making the connector `yield` it with the documents?

lib/core/elastic_connector_actions.rb
```diff
@@ -132,22 +132,30 @@ def update_connector_status(connector_id, status, error_message = nil)
...
+                                :last_deleted_document_count => metadata[:deleted_document_count])
 
         body = {
           :doc => {
             :status => sync_status,
-            :completed_at => Time.now
-          }.merge(status)
+            :completed_at => Time.now,
+            :error => error
+          }.merge(metadata)
```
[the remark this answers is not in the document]
wangch079: the `metadata` passed in should be in the format of
```
{
  :indexed_document_count => 1,
  :indexed_document_volume => 20,
  :deleted_document_count => 0,
  :metadata => {
    ...
  }
}
```

lib/core/sync_job_runner.rb
wangch079: Great suggestion @seanstory . Will do so.

lib/core/sync_job_runner.rb
```diff
@@ -19,16 +19,14 @@ def initialize(service_type, expected_fields, actual_fields)
   end
 
   class SyncJobRunner
+    JOB_REPORTING_INTERVAL = 10
```
wangch079, line 22: As suggested by @sphilipse here: https://github.com/elastic/enterprise-search-team/issues/3198#issuecomment-1305361746, will just make a hardcoded variable in 8.6

lib/core/elastic_connector_actions.rb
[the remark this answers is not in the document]
seanstory: Metadata has a nested `:metadata` key? That's so.... "meta"
:trollface:

seanstory (approved):

lib/core/elastic_connector_actions.rb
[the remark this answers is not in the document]
wangch079: I agree. Naming is hard

PR #403, opened 2022-11-03
Persist filtering in job records

seanstory:
## Closes https://github.com/elastic/enterprise-search-team/issues/3149

looks up the filtering rules in `.elastic-connectors` and saves them into `.elastic-connectors-sync-jobs`.

With latest Kibana, choosing "build a connector", running the example connector, and triggering a sync, you get a `.elastic-connectors-sync-jobs` record like:

```
     {
        "_index": ".elastic-connectors-sync-jobs-v1",
        "_id": "_mEnRIQB5rPmQXyEqZAB",
        "_score": 1,
        "_source": {
          "connector_id": "9WEjRIQB5rPmQXyEWJB2",
          "status": "completed",
          "worker_hostname": "Seans-MacBook-Pro-2.local",
          "created_at": "2022-11-04T19:39:20.699+00:00",
          "filtering": [
            {
              "domain": "DEFAULT",
              "rules": [
                {
                  "created_at": "2022-11-04T19:34:37.926Z",
                  "field": "_",
                  "id": "DEFAULT",
                  "order": 0,
                  "policy": "include",
                  "rule": "regex",
                  "updated_at": "2022-11-04T19:34:37.926Z",
                  "value": ".*"
                }
              ],
              "advanced_snippet": {
                "created_at": "2022-11-04T19:34:37.926Z",
                "updated_at": "2022-11-04T19:34:37.926Z",
                "value": {}
              },
              "warnings": []
            }
          ],
          "indexed_document_count": 3,
          "completed_at": "2022-11-04T19:39:21.315+00:00",
          "deleted_document_count": 0,
          "error": null
        }
      },
```

#### Pre-Review Checklist

- [x] Built gems (both `connectors_utility` and `connectors_service`) and included into Enterprise Search and tested that Enterprise Search works well with new gem versions. Instruction can be found [here](https://docs.google.com/document/d/10KJOIhe4sauDul8iWeV9Cn-_3uPWa76qG8SwYk6BCAA/edit)
  - PR: https://github.com/elastic/ent-search/pull/7012

### DIFF
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -129,7 +129,7 @@ module Core
             'domain' => filtering_domain['domain'],
             'rules' => filtering_domain.dig('active', 'rules'),
             'advanced_snippet' => filtering_domain.dig('active', 'advanced_snippet'),
-            'warnings' => [] # TODO in https://github.com/elastic/enterprise-search-team/issues/3174
+            'warnings' => [] # TODO: in https://github.com/elastic/enterprise-search-team/issues/3174
           }
         end
       end
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -88,7 +88,7 @@ module Core
         seq_no = nil
         primary_term = nil
         sync_in_progress = false
-        client.get(
+        connector_record = client.get(
           :index => Utility::Constants::CONNECTORS_INDEX,
           :id => connector_id,
           :ignore => 404,
@@ -114,11 +114,24 @@ module Core
           :connector_id => connector_id,
           :status => Connectors::SyncStatus::IN_PROGRESS,
           :worker_hostname => Socket.gethostname,
-          :created_at => Time.now
+          :created_at => Time.now,
+          :filtering => convert_connector_filtering_to_job_filtering(connector_record.dig('_source', 'filtering'))
         }
-        job = client.index(:index => Utility::Constants::JOB_INDEX, :body => body)
 
-        job['_id']
+        client.index(:index => Utility::Constants::JOB_INDEX, :body => body)
+      end
+
+      def convert_connector_filtering_to_job_filtering(connector_filtering)
+        return [] unless connector_filtering
+        connector_filtering = [connector_filtering] unless connector_filtering.is_a?(Array)
+        connector_filtering.each_with_object([]) do |filtering_domain, job_filtering|
+          job_filtering << {
+            'domain' => filtering_domain['domain'],
+            'rules' => filtering_domain.dig('active', 'rules'),
+            'advanced_snippet' => filtering_domain.dig('active', 'advanced_snippet'),
+            'warnings' => [] # TODO in https://github.com/elastic/enterprise-search-team/issues/3174
+          }
+        end
       end
 
       def update_connector_status(connector_id, status, error_message = nil)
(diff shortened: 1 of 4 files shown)


lib/core/elastic_connector_actions.rb
```diff
@@ -114,11 +114,24 @@ def claim_job(connector_id)
...
-        job['_id']
+        client.index(:index => Utility::Constants::JOB_INDEX, :body => body)
+      end
+
+      def convert_connector_filtering_to_job_filtering(connector_filtering)
+        return [] unless connector_filtering
+        connector_filtering = [connector_filtering] unless connector_filtering.is_a?(Array)
+        connector_filtering.each_with_object([]) do |filtering_domain, job_filtering|
+          job_filtering << {
+            'domain' => filtering_domain['domain'],
```
artem-shelkovnikov, line 129: Why is `domain` not in `active.rules`?

lib/core/elastic_connector_actions.rb
```diff
@@ -114,11 +114,24 @@ def claim_job(connector_id)
...
+          :created_at => Time.now,
+          :filtering => convert_connector_filtering_to_job_filtering(connector_record.dig('_source', 'filtering'))
         }
-        job = client.index(:index => Utility::Constants::JOB_INDEX, :body => body)
 
-        job['_id']
+        client.index(:index => Utility::Constants::JOB_INDEX, :body => body)
+      end
+
+      def convert_connector_filtering_to_job_filtering(connector_filtering)
```
artem-shelkovnikov, line 124: Nit: this should not be a public method - maybe a can be moved to Utility namespace?
Later we can provide different abstraction over it, I feel like `ElasticConnectorActions` becomes too complex - initial idea over it was to provide a domain interface over Elasticsearch, so just lightweight wrapper over the index. I'll think a bit about how to make it simpler.

lib/core/elastic_connector_actions.rb
artem-shelkovnikov: From what I remember `active` is the current rules - to not mix up rules in progress of editing and currently used rules?

artem-shelkovnikov: Actually even more specific question - what is `domain`?

artem-shelkovnikov (approved):
Looks good overall, left a couple questions

lib/core/elastic_connector_actions.rb
seanstory: See the "Notes" section in the [`In .elastic-connectors`](https://docs.google.com/document/d/1llPDGzduj8lST0N45v_G4PB0s0RUOBFurOIsBAD-Fyg/edit?pli=1#heading=h.l2heyce8rz9r). The `domain` was originally added with Crawler in mind. Crawler has different sets of crawl rules for each domain, so we needed to make a list/array structure rather than just an Object under `filtering`. I kept the key name as `domain` because it's more general meaning (like in DDD or DSL) works here - we want to define the "domain" in which these rules apply.

I see a not-too-distant future where we have several domains for MySQL so that we can define different configs and filtering rules for different databases/tables. Would also make it easier for us to support multiple Mongo collections in a single connector in the future.

lib/core/elastic_connector_actions.rb
seanstory: I had the same thought. However, I like being able to unit test just this bit of logic on its own, without having to mock Elasticsearch queries and responses. I've been told by mentors before that it's an anti-pattern to make a method private and then to write unit tests for it where you exercise access to a method that you shouldn't have direct access to. I know I _can_ do this with `send()`, but I'm hesitant to. WDYT?

artem-shelkovnikov: Makes sense to me, thanks for the explanation :)

PR #408, opened 2022-11-07
Add filtering rule features to ruby connectors

seanstory:
## part of https://github.com/elastic/enterprise-search-team/issues/3254

- adds the `features` key to the connectors mapping 
- adds logic to the Configuration task to set the `features`
  - allows ruby connectors to override features if they want, but by default, all Ruby connectors will support filtering rules

#### Pre-Review Checklist

## Related Pull Requests

- https://github.com/elastic/ent-search/pull/7027
- https://github.com/elastic/kibana/pull/144748


lib/core/configuration.rb
```diff
@@ -24,8 +24,10 @@ def update(connector_settings, service_type = nil)
             return
           end
           configuration = connector_class.configurable_fields_indifferent_access
+          features = connector_class.kibana_features.each_with_object({}) { |feature, hsh| hsh[feature] = true }
           doc = {
-            :configuration => configuration
+            :configuration => configuration,
+            :features => features
```
wangch079, line 30: Ideally, the configuration task will only be executed for non-native connectors, which means `features` will be empty for native connectors, assuming [FE won't do anything](https://elastic.slack.com/archives/C03GAPTQ2JF/p1667840735215359?thread_ts=1667839458.895959&cid=C03GAPTQ2JF):
>The connector would have to write that featureset to its connector document, the frontend has no knowledge of which framework is being used

@sphilipse , in my opinion, FE should add `features` for native connectors. Because we hardcode native connector definition in Kibana. We know it's MySQL, and we know it requires `url`, `port`, etc.. to connect to a remote source, we should also know it's supported by Python connector and filtering rules are not supported in 8.6

seanstory: Kibana is currently the system that sets native connectors to CONFIGURED? Where does that happen? Why isn't that being handled by the connector framework?

wangch079: When operators select `MongoDB` or `MySQL`, Kibana set the `service_type` and `configuration`, it also set `status` to `needs_configuration`. Kibana will set it to `configured` when operators save configurations in UI. cc @sphilipse

seanstory: Gotcha. Makes sense that we could hardcode this for Native connectors. Might oneday let our native connectors have a different feature set from our Build A Connectors. I added a checkbox to https://github.com/elastic/enterprise-search-team/issues/3254 to make sure I add this to Kibana.

mchernyavskaya (approved):
LGTM

PR #410, opened 2022-11-08
[Mongo] add "simple" pre-processing filtering logic

mchernyavskaya:
## Closes https://github.com/elastic/enterprise-search-team/issues/3168 

Created a base rules parser for filtering rules and a mongo-specific implementation.

#### Pre-Review Checklist

### DIFF
--- a/lib/connectors/mongodb/mongo_rules_parser.rb
+++ b/lib/connectors/mongodb/mongo_rules_parser.rb
@@ -24,7 +24,7 @@ module Connectors
         when 'regex'
           parse_regex(rule, field, value)
         else
-          raise "Unknown operator: #{rule}"
+          raise "Unknown operator: #{op}"
         end
       end
 
--- a/lib/connectors/mongodb/mongo_rules_parser.rb
+++ b/lib/connectors/mongodb/mongo_rules_parser.rb
@@ -1,0 +1,72 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+require 'connectors/base/simple_rules_parser'
+
+module Connectors
+  module MongoDB
+    class MongoRulesParser < Connectors::Base::SimpleRulesParser
+      def parse_rule(rule)
+        field = rule[:field]
+        value = rule[:value]
+        op = rule[:rule]&.to_s
+        case op
+        when 'Equals'
+          parse_equals(rule, field, value)
+        when '>'
+          parse_greater_than(rule, field, value)
+        when '<'
+          parse_less_than(rule, field, value)
+        when 'regex'
+          parse_regex(rule, field, value)
+        else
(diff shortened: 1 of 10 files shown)


lib/connectors/mongodb/mongo_rules_parser.rb
```diff
@@ -0,0 +1,72 @@
...
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+require 'connectors/base/simple_rules_parser'
+
+module Connectors
+  module MongoDB
+    class MongoRulesParser < Connectors::Base::SimpleRulesParser
```
seanstory, line 12: 👏 this is really clean. I like it.

spec/connectors/mongodb/mongo_rules_parser_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  context 'with multiple rules' do
+    let(:rules) do
+      [
+        { field: 'foo', value: 'bar1', policy: 'include', rule: 'Equals' },
+        { field: 'foo', value: 'bar2', policy: 'exclude', rule: '>' }
+      ]
+    end
+    it 'parses rules as and' do
+      result = subject.parse
+      expect(result).to match({ '$and' => [{ 'foo' => 'bar1' }, { 'foo' => { '$lte' => 'bar2' } }] })
```
seanstory, line 78: This is a good example of a rule that should cause us to issue warnings and _not_ use these for pre-filtering when we get to https://github.com/elastic/enterprise-search-team/issues/3174. Because it doesn't make any sense to AND rules where the first asserts a strict value of `foo`, and the second asserts a weaker range.

For this PR, this is great through. Really nice test coverage.

seanstory (approved):
I like it!

mchernyavskaya:
@seanstory what about documenting? Do we need to create a page in `enterprise-search-pubs`? Or do we already have one? Or do we add something to README?

timgrein (approved):
lgtm! I really like, that you've directly introduced the SimpleRulesParser base class 👏  We could follow up with another example connector implementing some simple rules after merge.

lib/connectors/mongodb/mongo_rules_parser.rb
```diff
@@ -0,0 +1,80 @@
...
+      def parse_rule(rule)
+        field = rule[:field]
+        value = rule[:value]
+        unless value.present?
+          raise "Value is required for field: #{field}"
+        end
+        unless field.present?
+          raise "Field is required for rule: #{rule}"
+        end
```
artem-shelkovnikov, line 23: This code can probably go to the base class

mchernyavskaya: I am thinking that it's probably connector-specific. As in, we might have sources where such conditions would make sense.

seanstory:
>  Do we need to create a page in enterprise-search-pubs? Or do we already have one? Or do we add something to README?

I haven't made issues yet for documentation, but have checkboxes for them in the epic. I think we'll knock all the docs out in one go.

PR #412, opened 2022-11-08
First pass at post-processing rules

seanstory:
## Closes https://github.com/elastic/enterprise-search-team/issues/3169

**Very much still a draft.**

first pass at adding post-processing filtering rules.
- [ ] should unify with Maryna's modeling of simple rules
- [ ] should unify with Tim's introduction of the core/filtering module
- [ ] should unify with bean vs Hashie::Mash vs Hash
- [ ] should unify strings vs symbols vs indifferent
- [ ] needs tests

#### Pre-Review Checklist


seanstory:
@timgrein @mchernyavskaya I expect your PRs will get in before this one, but wanted to put up what I have early for visibility.

lib/core/filtering/post_process_result.rb
```diff
@@ -0,0 +1,32 @@
...
+      def initialize(document, matching_rule)
+        @document = document
+        @matching_rule = matching_rule
+      end
+
+      def policy
+        matching_rule.policy
+      end
+
+      def include?
```
mchernyavskaya, line 23: Oops I [named these a bit differently](https://github.com/elastic/connectors-ruby/blob/main/lib/connectors/base/simple_rules_parser.rb#L39-L45). =) Should we sync up?

lib/core/filtering/simple_rule.rb
```diff
@@ -0,0 +1,64 @@
...
+      def initialize(rule_hash)
+        @policy = rule_hash.fetch(POLICY)
+        @field = rule_hash.fetch(FIELD)
+        @rule = rule_hash.fetch(RULE)
+        @value = rule_hash.fetch(VALUE)
+        @id = rule_hash.fetch(ID)
+      end
+
+      def match?(document)
+        return true if id == DEFAULT_RULE_ID
```
mchernyavskaya, line 40: I think it's not fool-proof. If one manipulates the data in `.elastic_connectors` one can assign a value `DEFAULT` to anything and therefore get inconsistent results. But I guess it's their problem if they tamper with data.

lib/core/filtering/simple_rule.rb
```diff
@@ -0,0 +1,64 @@
...
+        doc_value = document[field]
+        case rule
+        when 'equals'
+          doc_value == value
+        when 'starts_with'
```
mchernyavskaya, line 45: Aha so these are going to be lowercase and with an underscore? That's something I haven't deduced from the implementation document, so I guess simple rules should be fixed accordingly.

lib/core/filtering/simple_rule.rb
```diff
@@ -0,0 +1,64 @@
...
+        @rule = rule_hash.fetch(RULE)
+        @value = rule_hash.fetch(VALUE)
+        @id = rule_hash.fetch(ID)
+      end
+
+      def match?(document)
+        return true if id == DEFAULT_RULE_ID
+        doc_value = document[field]
+        case rule
+        when 'equals'
```
mchernyavskaya, line 43: Should we perhaps also allow `=` and/or `eq` to mean `equals`? Or is it too confusing?

lib/core/filtering/simple_rule.rb
```diff
@@ -0,0 +1,64 @@
...
+          doc_value == value
+        when 'starts_with'
+          doc_value.starts_with?(value)
+        when 'ends_with'
+          doc_value.ends_with?(value)
+        when 'contains'
+          doc_value.include?(value)
+        when 'regex'
```
mchernyavskaya, line 51: Actually, I thought that only `starts_with` and `ends_with` were to be covered in post-processing (by the way, why not in pre-processing?). Now I'm not sure what we have in pre-processing and what in post-processing?

lib/core/filtering/post_process_engine.rb
```diff
@@ -0,0 +1,37 @@
...
+      DEFAULT_DOMAIN = 'DEFAULT'
+
+      def initialize(job_description)
+        @ordered_rules = ordered_rules(job_description['filtering'])
+      end
+
+      def process(document)
+        @ordered_rules.each do |rule|
+          if rule.match?(document)
+            return PostProcessResult.new(document, rule)
```
mchernyavskaya, line 24: So what if the document is matched by more than one rule? Simple case:
* We have two rules: [ include :: color :: red], [ exclude :: name :: apple ]
* A document with { name: apple, color: red } comes in
* First matching rule is include, and we return a result immediately
* As a result, the document is indexed

Second case, and we reorder the rules:
* We have two rules: [ exclude :: name :: apple ], [ include :: color :: red]
* A document with { name: apple, color: red } comes in
* First matching rule is exclude, and we return a result immediately
* As a result, the document is not indexed

But both cases should have brought the same result right?

lib/core/filtering/post_process_result.rb
seanstory: nah, I'll just adopt your approach. Since your changes merged already, I think it will be easy for me to rebase on main and then refactor what I have here to unify with your changes.

lib/core/filtering/simple_rule.rb
seanstory: > I guess it's their problem if they tamper with data

That's my feeling too. Crawler accomplished this by not actually _storing_ the default rule, so it couldn't get messed up. But that meant that you couldn't actually see what the rules were from looking at ES, which we need to be able to do now. We'll definitely need to rely more on Kibana to not allow breaking this ID or getting the ordering wrong, and hope that most people use the UI, or _carefully_ use programmatic access.

lib/core/filtering/simple_rule.rb
seanstory: Yeah that was something I probably could have covered in the implementation document and I didn't. I don't really care how we serialize them as long as we're consistent. CC @sphilipse who will also need to align. @mchernyavskaya I'm happy to take the approach from your PR, though I may make small changes to make sure the casing is consistent across all rules.

lib/core/filtering/simple_rule.rb
seanstory: I'd prefer human-readable words over operators. However I wasn't consistent with that when I defined `<` and `>`. @zombieFox @sphilipse do you have strong opinions/designs for this for the UI? I think just aligning the UI and the backend makes sense.

lib/core/filtering/simple_rule.rb
seanstory: You're right, we could do this with regex for mongo. I didn't see an explicit "starts with" operator in their docs so I just left it, but maybe we should support that. Though I'm also fine to not scope creep that at this point.

lib/core/filtering/post_process_engine.rb
seanstory: Nope, ordering is very important and intentional for post-processing. This follows the behavior we've established and hardened in Crawl Rules. The UI lets you reorder rules. In the first example you provided, this would be good if you wanted to include red apples but not green apples, but all colors of pear (green AND red).

For pre-processing, order doesn't really matter, and it can't, if you're just ANDing all the clauses together. That discrepency is why we need to do our best effort to detect if simple rules cause conflicting pre-processing clauses. Again using your top example, if there were a third rule `[include :: name :: .*]`, we could detect that this conflicts with `[ exclude :: name :: apple ]`, and that we therefore shouldn't use _either_ for preprocessing.

lib/core/filtering/simple_rule.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+
+      class Rule
+        REGEX = 'regex'
+        EQUALS = 'equals'
+        STARTS_WITH = 'starts_with'
+        ENDS_WITH = 'ends_with'
+        CONTAINS = 'contains'
+        LESS_THAN = '<'
+        GREATER_THAN = '>'
```
seanstory, line 32: @mchernyavskaya I tried to go through and replace a lot of symbol and string usage with these constants. Hopefully that should make it easy for us to change them across the board as we iron out how we want these serialized in JSON.

timgrein:
I've a general question:
If a connector developer implements _some_ simple rules in pre-processing, will the framework be responsible for excluding these in post-processing or does the developer need to flag those as already implemented and the `PostProcessingEngine` will handle the rest?

Example: For data source A regex is not supported (and therefore not in pre-processing), but the rest of the rules is supported (and a developer implements these rules in pre-processing for performance gains) -> Regex support will be covered in post-processing?

lib/core/filtering/simple_rule.rb
mchernyavskaya: I actually prefer the snake_case since what we have is essentially an enum. We could just pull main in and refactor it in your branch.

lib/core/filtering/simple_rule.rb
mchernyavskaya: I'd propose these options:
1. `equals`, `greater_than` and `less_than`
2. `eq`, `gt` and `lt` (my preference - short and clear)
3. `=`, `>` and `<`
What do you think @seanstory @zombieFox @sphilipse ?

seanstory:
Good question, @timgrein. The current design is that all Simple Rules will be used for post-processing, irregardless of if they were also used for pre-processing. This probably isn't necessary, I'm struggling to come up with a good example of when you'd want a rule applied in both. However, it does simplify our logic a good bit, and shouldn't increase performance overhead too much. But let's keep this thought in mind for 8.7, as it might be a good improvement to make.

lib/utility/logger.rb
```diff
@@ -23,7 +24,7 @@ def level=(log_level)
       end
 
       def logger
-        @logger ||= Settings[:ecs_logging] ? EcsLogging::Logger.new(STDOUT) : ::Logger.new(STDOUT)
+        @logger ||= defined?(::Settings) && ::Settings[:ecs_logging] ? EcsLogging::Logger.new(STDOUT) : ::Logger.new(STDOUT)
```
seanstory, line 27: IDK how this wasn't causing tests to fail all over the place. `Settings` doesn't seem to be defined in our test scope. Is that expected, @artem-shelkovnikov @wangch079? Maryna and I were stumped.

artem-shelkovnikov: I'm not quite sure - I've never had a failing test that failed due to this line.

I feel that it's better to keep the code as it is right now and see the tests - maybe mock more stuff in tests if they fail

spec/core/filtering/post_process_engine_spec.rb
```diff
@@ -0,0 +1,193 @@
...
+
+    context 'default rule in wrong spot' do
+      before(:each) do
+        rules.unshift(Core::Filtering::SimpleRule::DEFAULT_RULE.to_h)
+      end
+      it_behaves_like 'included', 'test1'
+      it 'does not have the default rule in the ordered rules' do
+        expect(subject.rules.map(&:id)).to_not include(Core::Filtering::SimpleRule::DEFAULT_RULE_ID)
+      end
+      context 'no match' do
```
artem-shelkovnikov, line 187:
```suggestion
      end
      
      context 'no match' do
```

spec/core/filtering/simple_rule_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+  let(:document) do
+    {
+      'str' => 'foo',
+      'int' => 123,
+      'time' => time_now,
+      'datetime' => date_now,
+      'bool' => false
+    }
+  end
+  let(:date_now) { DateTime.now.new_offset(0) }
```
artem-shelkovnikov, line 37: Curious - why `new_offset(0)`?

spec/core/filtering/simple_rule_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+  let(:time_now) { Time.now }
+
+  shared_examples_for 'a match' do
+    it 'matches' do
+      expect(subject.match?(document)).to be
```
artem-shelkovnikov, line 42: Does this line check for truthy-ness?

lib/core/filtering/simple_rule.rb
```diff
@@ -0,0 +1,144 @@
...
+        'rule' => 'regex',
+        'value' => '.*',
+        'id' => SimpleRule::DEFAULT_RULE_ID
+      )
+
+      def match?(document)
+        return true if id == DEFAULT_RULE_ID
+        doc_value = document[field]
+        return false if doc_value.nil?
+        coerced_value = coerce(doc_value)
```
artem-shelkovnikov, line 73: Why in the code below the `doc_value` is sometimes compared with `value` and sometimes with `coerced_value`?

Also what is a `coerced_value`?

artem-shelkovnikov (commented):
Looks good overall, but left a couple of questions about the code that I could not understand

lib/core/filtering/post_process_engine.rb
```diff
@@ -0,0 +1,38 @@
...
+
+require 'core/filtering'
+
+module Core
+  module Filtering
+    class PostProcessEngine
+      attr_reader :rules
+
+      def initialize(job_description)
+        @rules = ordered_rules(job_description['filtering'])
```
mchernyavskaya, line 17: There's now a utility function in main that does this extraction (in `Utility::Filtering`). I think we should use that. Or alternatively, create a wrapper around the job entity with the methods we need, like it's been done with connectors config.

lib/core/filtering/post_process_result.rb
```diff
@@ -0,0 +1,27 @@
...
+
+module Core
+  module Filtering
+    class PostProcessResult
+      attr_reader :document, :matching_rule
+
+      def initialize(document, matching_rule)
+        @document = document
+        @matching_rule = matching_rule
+        Utility::Logger.debug("Document '#{document['id']}' matched filtering rule: #{matching_rule.id}. It will be #{matching_rule.policy}ed")
```
mchernyavskaya, line 19: ⛏️ Nitpicking and I like the intention, but it'll produce something like `includeed` or `excludeed`. 

```suggestion
        Utility::Logger.debug("Document '#{document['id']}' matched filtering rule: #{matching_rule.id}. It will be #{matching_rule.policy}d")
```

lib/utility/logger.rb
mchernyavskaya: @artem-shelkovnikov note that it wasn't `::Settings` - that's a change from @seanstory - it was `Settings` before. Which makes me think it was some kind of a global object/variable introduced (maybe?) by our app configuration gem, but we couldn't find anything like this, we only found the [Config global variable](https://github.com/elastic/connectors-ruby/blob/main/lib/app/config.rb#L124). So we weren't sure what that top-level `Settings` object is, and what initializes it.

mchernyavskaya (changes requested):

lib/core/filtering/simple_rule.rb
```diff
@@ -0,0 +1,144 @@
...
+#
+
+# frozen_string_literal: true
+
+require 'utility/logger'
+
+module Core
+  module Filtering
+    class SimpleRule
+      POLICY = 'policy'
```
vidok, line 14: I don't mind using constants as hash keys but I wonder why this is so in you case.

lib/utility/logger.rb
wangch079: Yes, it is a global object introduced by Ruby Config ([ref](https://github.com/rubyconfig/config#accessing-the-settings-object)):

>After installing the gem, `Settings` object will become available globally

lib/core/filtering/post_process_result.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+
+      def is_include?
+        matching_rule.is_include?
+      end
```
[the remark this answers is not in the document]
seanstory: The rule's policy _is_ the result of matching, since a `PostProcessResult` is defined by a document and its `matching_rule`. I could have just had the PostProcessEngine return a tuple, but I thought this would me more clear. 😅 

Would it be better if I named this function `should_include?`?

spec/core/filtering/post_process_engine_spec.rb
```diff
@@ -0,0 +1,193 @@
...
+
+    context 'no matches' do
+      let(:document) { { 'foo' => 'baz' } }
+      it_behaves_like 'included', Core::Filtering::SimpleRule::DEFAULT_RULE_ID
+    end
+
+    context 'document with symbol keys' do
+      let(:document) { { :foo => 'bar' } }
+      it_behaves_like 'excluded', 'test'
+    end
```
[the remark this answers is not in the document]
seanstory: no, it's not ignored. Here, the `it_behaves_like 'excluded', 'test'` is validating that the `test` Filtering Rule matches on this document, even though the rule has `'field' => 'foo'` and the document has the key `:foo` not `'foo'`.

spec/core/filtering/post_process_engine_spec.rb
```diff
@@ -0,0 +1,193 @@
...
+  let(:test_value) { 'bar' }
+  let(:test_rule_id) { 'test' }
+  let(:test_operator) { Core::Filtering::SimpleRule::Rule::EQUALS }
+  let(:test_rule) do
+    {
+      Core::Filtering::SimpleRule::ID => test_rule_id,
+      Core::Filtering::SimpleRule::FIELD => test_field,
+      Core::Filtering::SimpleRule::VALUE => test_value,
+      Core::Filtering::SimpleRule::POLICY => Core::Filtering::SimpleRule::Policy::EXCLUDE,
+      Core::Filtering::SimpleRule::RULE => test_operator
```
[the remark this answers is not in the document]
seanstory: You're not wrong. I constantized them because there were a few threads where the values seemed like they might be in-flux (`=` vs `equals` vs `Equals`, `<` vs `lt` vs `less_than`). I figured using constants would make it easier to refactor.

spec/core/filtering/post_process_engine_spec.rb
```diff
@@ -0,0 +1,193 @@
...
+      let(:document) { { 'foo' => { 'bar' => 'baz' } } }
+      it_behaves_like 'included', Core::Filtering::SimpleRule::DEFAULT_RULE_ID
+
+      context 'when contains rule is used' do
+        let(:test_operator) { Core::Filtering::SimpleRule::Rule::CONTAINS }
+        it_behaves_like 'excluded', 'test'
+      end
+
+      xcontext 'when the rule has nested field' do
+        # TODO: maryna points out this is inconsistent between pre and post processing for mongo, should ask Dana.
```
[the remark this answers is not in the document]
seanstory: Ah sorry, this was a TODO comment for me. I hope to remove it before merge.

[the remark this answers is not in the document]
seanstory: https://elastic.slack.com/archives/C03GAPTQ2JF/p1668442238289029

spec/core/filtering/simple_rule_spec.rb
seanstory: When I was running the tests, it was using my timezone (`-06:00`). That made it hard for me to write an `ends_with` test that would work on other people's machines. The `new_offset(0)` forces it to be GMT (`-00:00`).

Honestly though, IDK why our spec_helper isn't forcing this already. 🤔

spec/core/filtering/simple_rule_spec.rb
seanstory: Yep! `be` and `be_truthy` are equivalent, AIUI.

lib/core/filtering/simple_rule.rb
seanstory: It's your fault. I was trying to avoid symbols, but also avoid risking typos 😅 

See also: https://github.com/elastic/connectors-ruby/pull/412#discussion_r1021744506
WDYT?

lib/core/filtering/simple_rule.rb
```diff
@@ -0,0 +1,144 @@
...
+        @policy = rule_hash.fetch(POLICY)
+        @field = rule_hash.fetch(FIELD)
+        @rule = rule_hash.fetch(RULE)
+        @value = rule_hash.fetch(VALUE)
+        @id = rule_hash.fetch(ID)
+        @rule_hash = rule_hash
+      rescue KeyError => e
+        raise "#{e.key} is required"
+      end
+      def self.from_args(id, policy, field, rule, value)
```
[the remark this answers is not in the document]
seanstory: Screen Shot 2022-11-14 at 11 04 21 AM

Looks like my editor is now making it _look_ like I have newlines where I don't 😠

mchernyavskaya (approved):

mchernyavskaya:
Approving this to ease into the rules validation code; review changes will be addressed by @seanstory in a follow-up PR.

seanstory:
Since we're tight on time and this PR lays a foundation for Maryna's current ticket, I'm going to merge this PR as-is, and follow-up on all the review comments in a subsequent PR. @vidok @artem-shelkovnikov I promise I will address all your comments.

lib/core/filtering/post_process_result.rb
seanstory: Good catch!

lib/connectors/mongodb/mongo_rules_parser.rb
```diff
@@ -8,29 +8,30 @@
...
         case op
-        when 'Equals'
-          parse_equals(rule, field, value)
-        when '>'
-          parse_greater_than(rule, field, value)
-        when '<'
-          parse_less_than(rule, field, value)
-        when 'regex'
-          parse_regex(rule, field, value)
+        when Core::Filtering::SimpleRule::Rule::EQUALS
```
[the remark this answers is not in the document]
seanstory: I've reverted to using string literals in the follow up PR, except for the `Rule` and `Policy` enum classes

lib/core/filtering/simple_rule.rb
seanstory: I've switched to using string literals in the follow-up PR

spec/core/filtering/post_process_engine_spec.rb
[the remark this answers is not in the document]
seanstory: I switched to using string literals for these keys

PR #414, opened 2022-11-09
Ensure, that the is_shutting_down check is also executed, if an exception was raised

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3248 ##

This PR fixes the problem, that we don't exit the scheduler loop, if we've a reoccurring error (f.e. an authorization error) while `is_shutting_down` is set to true. By moving the break condition to the ensure block, we always check for `is_shutting_down`, even if an reoccurring error is raised.

#### Pre-Review Checklist


timgrein:
Have to check, if it also fixes [#2420](https://github.com/elastic/enterprise-search-team/issues/2420)

artem-shelkovnikov (approved):
Great catch!

PR #415, opened 2022-11-09
Make tests more consistent

artem-shelkovnikov:
As discussed in https://github.com/elastic/connectors-ruby/pull/389, some of our tests are not following best practices.

Changes implemented here:

- Change `context` to `describe` if the context is mentioning a particular method
- Double check `#method` and `.method` - see if they actually call class or instance methods and fix it
- Group methods that are not under any `describe` to actually be under `describe '#method'` or `describe '.method'`


mchernyavskaya (approved):
Great job!

PR #416, opened 2022-11-10
Add Job Entity

wangch079:
#### Pre-Review Checklist

### DIFF
--- a/lib/core/connector_job.rb
+++ b/lib/core/connector_job.rb
@@ -106,6 +106,7 @@ module Core
     def reload
       es_response = ElasticConnectorActions.get_job(id)
       raise ConnectorJobNotFoundError.new("Connector job with id=#{id} was not found.") unless es_response[:found]
+      # TODO: remove the usage of with_indifferent_access. get_id method is expected to return a hash
       @elasticsearch_response = es_response.with_indifferent_access
       @connector = nil
     end
--- a/lib/core/connector_job.rb
+++ b/lib/core/connector_job.rb
@@ -98,7 +98,7 @@ module Core
       @connector ||= ConnectorSettings.fetch_by_id(connector_id)
     end
 
-    def connector!
+    def reload_connector!
       @connector = nil
       connector
     end
--- a/lib/core/connector_job.rb
+++ b/lib/core/connector_job.rb
@@ -62,12 +62,12 @@ module Core
       status == Connectors::SyncStatus::CANCELING
     end
 
-    def connector_id
-      self[:connector_id]
-    end
-
     def connector_snapshot
       self[:connector]
+    end
+
+    def connector_id
+      connector_snapshot[:id]
     end
 
     def index_name
--- a/lib/core/connector_job.rb
+++ b/lib/core/connector_job.rb
@@ -1,0 +1,134 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+require 'active_support/core_ext/hash/indifferent_access'
+require 'connectors/sync_status'
+require 'core/elastic_connector_actions'
+require 'utility'
+
+module Core
+  class ConnectorJob
+    DEFAULT_PAGE_SIZE = 100
+
+    # Error Classes
+    class ConnectorJobNotFoundError < StandardError; end
+
+    def self.fetch_by_id(job_id)
+      es_response = ElasticConnectorActions.get_job(job_id)
+
+      raise ConnectorJobNotFoundError.new("Connector job with id=#{job_id} was not found.") unless es_response[:found]
+      new(es_response)
+    end
+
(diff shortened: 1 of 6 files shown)


lib/core/elastic_connector_actions.rb
```diff
@@ -52,6 +52,10 @@ def get_connector(connector_id)
         client.get(:index => Utility::Constants::CONNECTORS_INDEX, :id => connector_id, :ignore => 404).with_indifferent_access
       end
 
+      def get_job(job_id)
+        client.get(:index => Utility::Constants::JOB_INDEX, :id => job_id, :ignore => 404).with_indifferent_access
```
vidok, line 56: Do we really need to use `with_indifferent_access` here?

lib/core/connector_job.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    def connector!
+      @connector = nil
+      connector
+    end
+
+    def reload
+      es_response = ElasticConnectorActions.get_job(id)
+      raise ConnectorJobNotFoundError.new("Connector job with id=#{id} was not found.") unless es_response[:found]
+      @elasticsearch_response = es_response.with_indifferent_access
```
vidok, line 109: Do we really need to use `with_indifferent_access` here?

lib/core/connector_job.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    def pipeline
+      connector_snapshot[:pipeline]
+    end
+
+    def connector
+      @connector ||= ConnectorSettings.fetch_by_id(connector_id)
+    end
+
+    def connector!
```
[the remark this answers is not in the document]
wangch079: We have a similar pattern used [here](https://github.com/elastic/ent-search/blob/83049929bc91571e81fe09e249199592ed2f9180/frito_pie/app/models/frito_pie/content_source.rb#L792-L799), but I didn't mean the existing is correct.

Method ending with `!` in ruby conventionaly means it modifies the object, which is not the case here. What I want to imply from the method name is 1) it's reloaded 2) it still returns a connetor object. `connector!` doesn't serve 1) very well, and `reload_connector!` doesn't imply 2).

lib/core/connector_job.rb
wangch079: This is a defensive action, because we use symbol keys everywhere in the class.

lib/core/connector_job.rb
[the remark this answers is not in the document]
vidok: > Method ending with ! in ruby conventionaly means it modifies the object

Agree, however it should be used only with verbs like (`pop!`, `shift!`, etc). `reload_connector!` is based on ActiveRecord pattern where all models have `reload!` method that does a new query to a database to retrieve fresh data. The `verb+!` has to describe what exactly will happen with the object when you call it. 

`authorization_details!` also has an incorrect signature from my point of view.

lib/core/connector_job.rb
vidok: `with_indifferent_access` adds more work for GC but also it makes access by key operation much slower. 

here is some [benchmarks](https://docs.google.com/presentation/d/1EKSoFGIvIATWpP2kN9WC7nzmeFQD_D6ukmUNOPkVsWc/edit#slide=id.g16453f538f0_0_0)

`@elasticsearch_response` is not exposed outside of the class so you can use strings safely. If you really want to use keys than, you can call `transform_keys(&:to_sym)` to make keys symbols.

lib/core/elastic_connector_actions.rb
vidok: Actually, it does. 

See my [slides](https://docs.google.com/presentation/d/1EKSoFGIvIATWpP2kN9WC7nzmeFQD_D6ukmUNOPkVsWc/edit#slide=id.g16453f538f0_0_0)

lib/core/connector_job.rb
wangch079: Resolve another conversation to keep the discussion in one place.

> `@elasticsearch_response` is not exposed outside of the class so you can use strings safely.

This makes sense.

Another reason to use `with_indifferent_access`: The Elasticsearch client returns an [Elasitcsearch::API::Response](https://github.com/elastic/elasticsearch-ruby/blob/main/elasticsearch-api/lib/elasticsearch/api/response.rb) object, which
>behaves like a Hash, except when status or headers are called upon it, in which case it returns the original object's status and headers.

I think it's favorable to remove the dependencies of third-party lib when Ruby hash is enough (though `ActiveSupport::HashWithIndifferentAccess` is from `active_support`, I think we probably should just use `.to_h`)

I remembered the usage of `with_indifferent_access` was also discussed somewhere else. I suggest we address it in a separate issue if we want to (`with_indifferent_access` is also used in other methods of `ElasticConnectorActions`)

vidok: Yeah, I thought all folks agreed not to add new `with_indifferent_access` usage after my presentation. It's okay if we leave old code unchanged for some time but I'd love to get rid of it. 

I'm okay to merge it if you leave a #TODO to remove it later.

wangch079: Added `TODO` in multiple places.

vidok (approved):
👍🏻

PR #421, opened 2022-11-10
Ignore large documents when ingesting the data

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/3097

Adding simple logic to ignore documents that are bigger than allowed. If such document is encountered, it's ignored and a `warn` statement is produced.

#### Pre-Review Checklist

## Release Note

Now connectors ignore the documents that are too large without attempting to ingest them.

### DIFF
--- a/lib/core/ingestion/ingester.rb
+++ b/lib/core/ingestion/ingester.rb
@@ -11,8 +11,10 @@ require 'utility/logger'
 module Core
   module Ingestion
     class Ingester
-      def initialize(sink_strategy)
+      def initialize(sink_strategy, max_allowed_document_size = 5 * 1024 * 1024)
         @sink_strategy = sink_strategy
+        @max_allowed_document_size = max_allowed_document_size
+
         @ingested_count = 0
         @ingested_volume = 0
         @deleted_count = 0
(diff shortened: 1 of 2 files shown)


artem-shelkovnikov:
Follow-up plan:

1. Unite `Ingester` and `EsSink` into single class - there's no more need to have a sink hierarchy
2. Make sure that maximum allowed document size is always less than maximum queue volume
3. Allow to set the thresholds via a config file

lib/core/ingestion/ingester.rb
```diff
@@ -11,8 +11,10 @@
 module Core
   module Ingestion
     class Ingester
-      def initialize(sink_strategy)
+      def initialize(sink_strategy, max_allowed_document_size = 5 * 1024 * 1024)
```
wangch079, line 14: Will this threshold be related to the queue size? Are you planning to allow them configured separately?

artem-shelkovnikov: It's a part of the mentioned follow-up:

> Make sure that maximum allowed document size is always less than maximum queue volume

For first iteration I believe they can just be the same. If later we'll need different numbers, we can allow for it, but will we ever actually need it? Probably not.

wangch079 (approved):

PR #422, opened 2022-11-10
Use utility function for filter extraction.

timgrein:
Replace the check, if filtering is an array with the available utility function.

#### Pre-Review Checklist


mchernyavskaya (approved):
LGTM

PR #423, opened 2022-11-10
[DRAFT] Handle filtering array and filtering object for the all sync jobs/connector implemenations

timgrein:
## Relates to https://github.com/elastic/enterprise-search-team/issues/3303 ##

This PR handles the correct filter extraction on a sync job level, so that the different connector and filtering validation implementations do not need to handle correct filter extraction by themselves.

#### Pre-Review Checklist


lib/core/sync_job_runner.rb
```diff
@@ -49,6 +49,9 @@ def do_sync!
       job_record = ElasticConnectorActions.claim_job(@connector_settings.id)
       job_description = job_record['_source']
       job_id = job_record['_id']
+
+      # ES can return an array or an object
+      job_description['filtering'] = Utility::Filtering.extract_filter(job_description['filtering'])
```
artem-shelkovnikov, line 54: I think this code should happen much earlier, otherwise other code places that could use the method `claim_job` will receive a different data structure inside `filtering` key.

Can you move this logic into `claim_job`?

vidok: I agree with Artem however, I'm not sure if this should happen in `claim_job`. It looks specific for `job_runner`. 

I would keep `job_description` immutable to make sure other devs can rely on the defined index mapping. I think `claim_job` has to only claim jobs and does not do anything else ([SR principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)).   

`validate_filtering` is a perfect candidate for this logic encapsulation

artem-shelkovnikov: For this particular case it looks like a Elasticsearch-specific workaround, based purely on the expectations that Elasticsearch stores every field as an Array.

I've also seen a PR that introduces a `Job` entity - might be worth to put this logic there, into the Job class?

artem-shelkovnikov: I thought the whole code that accesses `filtering` expects only one record inside, thus my assumption was that the `Job` has to handle this array -> single entry mapping internally. If it's not the case then yes, only validation-specific logic should do this.

timgrein: Sorry missed those comments! Actually this happens now in the generic `self.validate_filtering` method, **but** if someone overrides this in a specific connector we need to rely on a correct implementation. We could also wrap it into a method, which shouldn't be overridden (we did something similar to `configurable_fields`) and call this method during a validation/sync job passing the extracted filter object to the generic (or specific overridden implementation)

WDYT?

artem-shelkovnikov: Sorry I can't wrap my head around it, can you provide a code snippet as an example?

Also can you resolve the merge conflicts before we further discuss the PR? I believe some code in `main` also added `job_description.extracted_filtering` or something similar.

timgrein: `# Do not override
def self.validate_filtering_with_filter_extraction(filtering = {})
  # nothing to validate
  return { :state => Core::Filtering::ValidationStatus::VALID, :errors => [] } unless filtering.present?

  filter = Utility::Filtering.extract_filter(filtering)

  validate_filtering(filter)
end

def self.validate_filtering(filter = {})
  advanced_snippet = filter.dig(:advanced_snippet, :value)

  snippet_validator_instance = advanced_snippet_validator.new(advanced_snippet)

  snippet_validator_instance.is_snippet_valid?
end`

If we structure it like this a connector developer can just override `self.validate_filtering` without the need to worry about calling `Utility::Filtering.extract_filter(filtering)`. `self.validate_filtering_with_filter_extraction` should probably be named different, just as a small illustration 👍  WDYT? 

@vidok or @artem-shelkovnikov do you know [sorbet](https://sorbet.org/)? @seanstory pointed me to that and I think it's pretty cool, if there are no significant performance implications. Could probably avoid a lot of headaches and we would be able to mark methods as [final](https://sorbet.org/docs/final) (like `self.validate_filtering_with_filter_extraction`).

wangch079 (commented):
It seems this PR only adds specs to `SyncJobRunner`.

And I see the new test examples are basically testing all types of `filtering` that `Utility::Filtering.extrat_filter` can accept and what is expected to return.

I feel this kind of tests are more appropriate to happen in [spec/utility/filtering_spec.rb](https://github.com/elastic/connectors-ruby/blob/main/spec/utility/filtering_spec.rb)

And for `spec/core/sync_job_runner_spec.rb`, what it cares about is what can be returned by `Utility::Filtering.extrat_filter`, `{}` and `{ ...extracted rule... }`. If there's no logic difference to handle the extracted rule, you can simply add a before block:
```ruby
allow(Utility::Filtering).to receive(:extract_filter).with(filtering).and_return({})
```

timgrein:
> It seems this PR only adds specs to `SyncJobRunner`.
> 
> And I see the new test examples are basically testing all types of `filtering` that `Utility::Filtering.extrat_filter` can accept and what is expected to return.
> 
> I feel this kind of tests are more appropriate to happen in [spec/utility/filtering_spec.rb](https://github.com/elastic/connectors-ruby/blob/main/spec/utility/filtering_spec.rb)
> 
> And for `spec/core/sync_job_runner_spec.rb`, what it cares about is what can be returned by `Utility::Filtering.extrat_filter`, `{}` and `{ ...extracted rule... }`. If there's no logic difference to handle the extracted rule, you can simply add a before block:
> 
> ```ruby
> allow(Utility::Filtering).to receive(:extract_filter).with(filtering).and_return({})
> ```

@wangch079 Definitely right, will keep that in mind! I've talked with Artem and we'll probably go a different route to handle this filter extraction (we'll use a dedicated "data" object (like a Java bean/POJO), which can handle a filtering array containing multiple filter objects or a single filter object respectively). I'll come back to this, when filtering validation works for 8.6.

lib/core/sync_job_runner.rb
vidok: I know sorbet, and it solves a slightly different problem than we do. If someone wants to override the method nothing will stop them. Also, to use schema validation you need to operate with sorbet's `T::Struct` which will impact the code base. 

For schema validation, you can have a look at https://dry-rb.org/gems/dry-validation/1.8/schemas/ which is super easy to extend if someone wants to override the method.

PR #425, opened 2022-11-14
Only return supported connectors in native mode

wangch079:
This PR made the following changes:
1. When running in native mode, only supported native connectors will be returned when querying ES
2. When running in single mode, an additional check is added in the pre-flight check to make sure the service type in config is supported

#### Pre-Review Checklist

### DIFF
--- a/lib/app/preflight_check.rb
+++ b/lib/app/preflight_check.rb
@@ -23,6 +23,7 @@ module App
         check_es_connection!
         check_es_version!
         check_system_indices!
+        check_single_connector!
       end
 
       private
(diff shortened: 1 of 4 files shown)


lib/app/preflight_check.rb
```diff
@@ -59,6 +60,16 @@ def check_system_indices!
...
 
+      #-------------------------------------------------------------------------------------------------
+      # Ensures the connector is supported when running in non-native mode
+      def check_single_connector!
+        if App::Config.native_mode
+          Utility::Logger.info('Skip single connector check for native mode.')
+        elsif !Connectors::REGISTRY.registered?(App::Config.service_type)
+          fail_check!("The service type #{App::Config.service_type} is not supported. Terminating...")
+        end
+      end
```
artem-shelkovnikov, line 71: Do you mind adding a test for this method?

artem-shelkovnikov (approved):
LGTM, a couple tests are missing for the added logic

PR #426, opened 2022-11-14
Refactor sinks to be more simple and robust

artem-shelkovnikov:
## Follow-up for https://github.com/elastic/connectors-ruby/pull/389

This PR refactors the sink-related code to make it more simple and work better. Some highlights:

- Ingester and EsSink are united in a single EsSink class
- Tests for Ingester and EsSink are united in a single file with minor modification
- (bugfix) Now EsSink properly reports the stats of ingestion - before if document was added to sink but not flushed, the sink will still report this document as already ingested/deleted. Now the sink properly reports when the document was actually ingested into Elasticsearch

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/connectors-ruby/pull/389

- [x] Built gems (both `connectors_utility` and `connectors_service`) and included into Enterprise Search and tested that Enterprise Search works well with new gem versions. Instruction can be found [here](https://docs.google.com/document/d/10KJOIhe4sauDul8iWeV9Cn-_3uPWa76qG8SwYk6BCAA/edit)


artem-shelkovnikov:
cc @wangch079 @seanstory

spec/core/ingestion/es_sink_spec.rb
```diff
@@ -16,165 +17,348 @@
...
+        end
       end
-      let(:another_document) do
-        {
-          :id => 2,
-          :text => 'work work!'
-        }
+
+      context 'when ingested document is not empty' do
+        context('when bulk queue still has capacity') do
```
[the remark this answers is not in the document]
artem-shelkovnikov: whoops, my mistake

spec/core/ingestion/es_sink_spec.rb
```diff
@@ -16,165 +17,348 @@
...
+        let(:document) { {} }
+
+        it 'does not add document to the queue' do
+          expect(bulk_queue).to_not receive(:add)
+
+          subject.ingest(document)
+        end
 
-        subject.ingest(document[:id], serialized_document)
+        it 'produces a warning' do
```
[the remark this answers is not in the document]
artem-shelkovnikov: Do you mean a warning or a test for it?

lib/core/ingestion/es_sink.rb
```diff
@@ -11,49 +11,108 @@
...
+# While we don't actually allow to output our data anywhere except
+# Elasticsearch, we still want to be able to do so sometime in future.
+#
+# This class should stay simple and any change to the class should be careful
+# with the thought of introducing other sinks in future.
 module Core
   module Ingestion
     class EsSink
-      def initialize(index_name, request_pipeline, bulk_queue = Utility::BulkQueue.new)
+      def initialize(index_name, request_pipeline, bulk_queue = Utility::BulkQueue.new, max_allowed_document_size = 5 * 1024 * 1024)
```
seanstory, line 24: I suggest adding the `5*1024*1024` to the `constants.rb` file

lib/core/ingestion/es_sink.rb
```diff
@@ -11,49 +11,108 @@
...
+          :indexed_document_count => 0,
+          :deleted_document_count => 0,
+          :indexed_document_volume => 0
+        }
       end
 
-      def ingest(id, serialized_document)
-        index_op = serialize({ 'index' => { '_index' => index_name, '_id' => id } })
+      def ingest(document)
+        unless document&.any?
```
seanstory, line 46:
```suggestion
        unless document&.present?
```

Nitpick: `.any?` feels like something to use on a collection, but "document" is singular. Technically it works for a hash, but reads kinda funny IMO

lib/core/ingestion/es_sink.rb
```diff
@@ -11,49 +11,108 @@
...
 
       def flush
         data = @operation_queue.pop_all
         return if data.empty?
 
         @client.bulk(:body => data, :pipeline => @request_pipeline)
+
+        @completed[:indexed_document_count] += @queued[:indexed_document_count]
+        @completed[:deleted_document_count] += @queued[:deleted_document_count]
+        @completed[:indexed_document_volume] += @queued[:indexed_document_volume]
```
seanstory, line 100: Should we actually grab these values from the response to the `bulk` request? Hypothetically, some of these operations might fail, in which case we probably shouldn't count them as successes.

lib/core/sync_job_runner.rb
```diff
@@ -24,7 +24,7 @@ class SyncJobRunner
 
     def initialize(connector_settings)
       @connector_settings = connector_settings
-      @ingester = Core::Ingestion::Ingester.new(Core::Ingestion::EsSink.new(connector_settings.index_name, @connector_settings.request_pipeline))
+      @ingester = Core::Ingestion::EsSink.new(connector_settings.index_name, @connector_settings.request_pipeline)
```
seanstory, line 27:
```suggestion
      @sink = Core::Ingestion::EsSink.new(connector_settings.index_name, @connector_settings.request_pipeline)
```

Nitpick: any reason to keep the `ingester` terminology?

lib/core/ingestion/es_sink.rb
artem-shelkovnikov: It'll actually be addressed in a follow-up PR to introduce this as a configuration provided by YAML file, I'll do it there!

lib/core/ingestion/es_sink.rb
artem-shelkovnikov: `.present` is a Rails method, I had a comment from someone to try to not use Rails method here.

Another option would be to use `.empty?` method but it'll require flipping the condition that'll look uglier probably:

```
unless document || document.empty? == false
```

or 

```
if document.nil? || document.empty? # this looks pretty okay, gotta use it!
```

lib/core/ingestion/es_sink.rb
artem-shelkovnikov: That is possible, though right now the logic is such that if `bulk` fails, then we fail the sync fully.

Once we add ingestion resilience, this will have to change to the logic you described. We can do it now too, but I don't feel I have capacity for it currently

lib/core/sync_job_runner.rb
artem-shelkovnikov: That is a bad merge, thanks for noticing!

spec/core/ingestion/es_sink_spec.rb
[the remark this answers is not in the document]
wangch079: I meant test.

lib/core/ingestion/es_sink.rb
wangch079: Actually, it's from [activesupport](https://rubygems.org/gems/activesupport), which, is extracted from Rails.

We have `activesupport` as a gem dependency, if you want to use it. But I don't have preference between `any?` and `present?`

lib/core/ingestion/es_sink.rb
wangch079: It's definitely more accurate, but the [response of the Bulk API ](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-api-response-body) doesn't return the volume, we certainly can have an id-to-volume map, and calculate the the total volume based on the response, which I think it's a bit overkill.

wangch079 (approved):
LGTM

lib/core/ingestion/es_sink.rb
seanstory: @vidok I think it may have been you who discourages activesupport usage. Why is that? I find these types of helpers really valuable, and since the majority of ruby users are rails users, the StackOverflow snippets we encounter often assume active_support is available. Why avoid it?

PR #427, opened 2022-11-14
Add sync jobs producer and consumer

vidok:
## Closes https://github.com/elastic/enterprise-search-team/issues/3250

This PR will introduce a new way of performing sync jobs. Instead of checking `connector.sync_now` flag the connector service will be pulling jobs from `.elastic-connectors-sync-jobs` index.  

Also, in this PR I addressed graceful shutdown. 

image

You can see in the picture above that when the service received `kill -s SIGTERM` it started the shutdown process but waited until the job was completed. 

#### Pre-Review Checklist
- [x] Covered the changes with automated tests (unit tests will be added later)


lib/connectors/sync_status.rb
```diff
@@ -26,7 +26,7 @@ class SyncStatus
       ERROR
     ]
 
-    PENDING_STATUES = [
```
artem-shelkovnikov, line 29: Lol, nice typo

vidok: Yeah, I was thinking to leave it as an Easter Egg :)

lib/app/dispatcher.rb
```diff
@@ -68,7 +68,11 @@ def start_polling_jobs!
         scheduler.when_triggered do |connector_settings, task|
           case task
           when :sync
-            start_sync_task(connector_settings)
+            # update connector metadata
+            # @TODO rename #claim_job because it doesn't claim a job
+            Core::ElasticConnectorActions.claim_job(connector_settings.id)
```
wangch079, line 73: I'm introducing some changes in PR https://github.com/elastic/connectors-ruby/pull/428.

`Core::ElasticConnectorActions` is a wrapper to access ES. IMO, it shouldn't contain any business logic, e.g. `claim_job`, `complete_sync`. It should only contain methods like `update_job`, `create_job`.

I didn't touch `claim_job`, as I think this is more related to the decoupling work. I will eventually remove `complete_sync` method. Whenever a job completes, it should call `job.done!` or `job.error!`, where the `job` is a job entity.

lib/core/connector_settings.rb
```diff
@@ -114,6 +114,16 @@ def valid_index_name?
...
     end
 
+    def ready_for_sync?
+      Connectors::REGISTRY.registered?(service_type) &&
+        valid_index_name? &&
+        connector_status_allows_sync?
+    end
+
+    def running?
+      @elasticsearch_response[:_source][:last_sync_status] == Connectors::SyncStatus::IN_PROGRESS
```
wangch079, line 124: I'm OK with this now, but in long term, we probably should not rely on this, but on `.elastic-connectors-sync-jobs`.

lib/core/elastic_connector_actions.rb
```diff
@@ -157,28 +157,25 @@ def claim_job(connector_id)
...
-          :status => Connectors::SyncStatus::IN_PROGRESS,
-          :worker_hostname => Socket.gethostname,
-          :created_at => Time.now,
-          :started_at => Time.now,
-          :last_seen => Time.now,
-          :connector => {
-            :id => connector_id,
-            :filtering => convert_connector_filtering_to_job_filtering(connector_record.dig('_source', 'filtering'))
+          status: Connectors::SyncStatus::PENDING,
+          worker_hostname: Socket.gethostname,
```
wangch079, line 165: `worker_hostname` should only be added when it's claimed. It's the machine which executes the job, not the one which enqueues the job.

lib/core/elastic_connector_actions.rb
```diff
@@ -157,28 +157,25 @@ def claim_job(connector_id)
...
+          created_at: Time.now,
+          started_at: Time.now,
```
wangch079, line 167: `started_at` is also the time when it's claimed.

lib/core/elastic_connector_actions.rb
```diff
@@ -157,28 +157,25 @@ def claim_job(connector_id)
...
+          last_seen: Time.now,
+          connector: {
+            id: connector_id,
+            filtering: convert_connector_filtering_to_job_filtering(connector_settings.filtering)
```
wangch079, line 171: We should also add `configuration`, `pipeline`, `index_name`, `language` and `service_type`

lib/core/jobs/consumer.rb
```diff
@@ -0,0 +1,120 @@
...
+            if @running.false?
+              Utility::Logger.info('Shutting down the loop')
+              break
+            end
+
+            sleep(POLL_INTERVAL)
+            Utility::Logger.info('Getting registered connectors')
+
+            # load active connectors settings
+            connectors = ready_for_sync_connectors
```
wangch079, line 64: Don't feel this is necessary. We should guard this at producer.

wangch079 (commented):
Not an official review. Just want to quick go through you implementation to avoid potential conflicts

lib/core/elastic_connector_actions.rb
vidok: Good point

lib/core/elastic_connector_actions.rb
vidok: yep, this PR is draft so don't consider is complete

lib/app/dispatcher.rb
vidok: > Core::ElasticConnectorActions is a wrapper to access ES

What exactly it should be wrapping from your point of view?

lib/core/connector_job.rb
```diff
@@ -87,11 +94,11 @@ def configuration
...
 
     def filtering
-      connector_snapshot[:filtering]
+      @elasticsearch_response[:_source][:filtering]
     end
 
     def pipeline
-      connector_snapshot[:pipeline]
+      @elasticsearch_response[:_source][:pipeline]
     end
```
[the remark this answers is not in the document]
vidok: I don't see any points of loading connector doc to retrieve `:pipeline` because it's already a part of the job document

lib/core/connector_settings.rb
```diff
@@ -114,6 +114,16 @@ def valid_index_name?
       index_name&.start_with?(Utility::Constants::CONTENT_INDEX_PREFIX)
     end
 
+    def ready_for_sync?
+      Connectors::REGISTRY.registered?(service_type) &&
```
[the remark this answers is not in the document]
vidok: I disagree because with a new setup anyone can index a job document into the sync job index we need to make sure that the connector is ready to pick up the job.

lib/app/dispatcher.rb
artem-shelkovnikov: When I originally introduced `ElasticConnectorActions` its goal was to have a "contract in actions". What I mean by that:

Connectors protocol defines "entities" - Jobs, Connectors - and their properties. However, it's only a part of the protocol, another part of protocol is "actions": e.g. "start a sync", "complete a sync", "update connector configurable fields" - atomic actions that make the protocol actually work. It was done so that it was easier to see how the code inside the framework changes the state of the system - instead of freeform "es_client.update" it does for example:

```
job_response = ElasticConnectorActions.create_job(connector.id)

job_claim_response = ElasticConnectorActions.claim_job(connector.id)

if job_claim_response['success']
  # do stuff
  ElasticConnectorActions.update_running_job_state(job.id, state: RUNNING, indexed_doc_count: 55, deleted_doc_count: 15)

  # finish
  ElasticConnectorActions.finish_job_successfully(job.id)
end
```

However with time this class started to have more generic methods like "update_connector" and "update job" that made the original approach not work that well any more.

So TLDR: this class was supposed to encapsulate connectors protocol "actions", but not provide any domain logic inside at all. It should have just wrapped Elasticsearch client with more "smart" methods.

lib/app/dispatcher.rb
```diff
@@ -28,6 +28,13 @@ class << self
       def start!
         running!
         Utility::Logger.info("Starting connector service in #{App::Config.native_mode ? 'native' : 'non-native'} mode...")
+
+        # set exit hook
+        Kernel.at_exit { shutdown! }
```
artem-shelkovnikov, line 33: Should it be in the `app.rb` itself? Feels a bit more natural there.

lib/app/dispatcher.rb
```diff
@@ -68,7 +77,11 @@ def start_polling_jobs!
         scheduler.when_triggered do |connector_settings, task|
           case task
           when :sync
-            start_sync_task(connector_settings)
+            # update connector metadata
+            # @TODO rename #claim_job because it doesn't claim a job
```
artem-shelkovnikov, line 81:
```suggestion
            # @TODO rename #claim_job because it doesn't claim a job
```

lib/app/dispatcher.rb
```diff
@@ -132,6 +129,25 @@ def start_filter_validation_task(connector_settings)
...
+      def start_consumers!
+        @consumer = Core::Jobs::Consumer.new(
+          poll_interval: POLL_INTERVAL,
+          termination_timeout: TERMINATION_TIMEOUT,
+          min_threads: MIN_THREADS,
+          max_threads: MAX_THREADS,
+          max_queue: MAX_QUEUE
+        )
+
+        @consumer.subscribe!(index_name: Utility::Constants::JOB_INDEX)
```
artem-shelkovnikov, line 142: The method is using plural, but there's one consumer.

This is an extension point, right? At some point you'll have multiple consumers?

lib/core/connector_job.rb
```diff
@@ -67,7 +74,7 @@ def connector_snapshot
...
 
     def connector_id
-      connector_snapshot[:id]
+      @elasticsearch_response[:_source][:connector][:id]
```
artem-shelkovnikov, line 77: Do you want to `dig` for nullability checks or it's okay to raise an exception if something's missing?

lib/core/connector_job.rb
```diff
@@ -111,6 +118,33 @@ def reload
...
     end
 
+    def with_lock
+      seq_no = nil
+      primary_term = nil
+
+      response = ElasticConnectorActions.get_job(id)
+
+      yield response, seq_no, primary_term
+    end
```
artem-shelkovnikov, line 128: So it's not a real lock, it's optimistic concurrency, right? Method name is a bit confusing

lib/core/connector_job.rb
```diff
@@ -111,6 +118,33 @@ def reload
...
+
+    def process!
```
artem-shelkovnikov, line 130: So it's the actual job claiming, right? It does not actually run the sync?

lib/core/connector_settings.rb
artem-shelkovnikov, line 124: Might be worth to add some sort of a killswitch to also not consider jobs that are stuck/orphaned, though probably not important for now.

lib/core/jobs/consumer.rb
```diff
@@ -0,0 +1,120 @@
...
+            pending_jobs = Core::ConnectorJob.pending_jobs(connectors_ids: connectors.keys)
+            Utility::Logger.info("Number of pending jobs: #{pending_jobs.size}")
+
+            pending_jobs.each do |job|
+              connector_settings = connectors[job.connector_id]
+
+              pool.post do
+                Core::ElasticConnectorActions.ensure_content_index_exists(connector_settings.index_name)
+                job_runner = Core::SyncJobRunner.new(connector_settings, job)
+                job_runner.execute
```
artem-shelkovnikov, line 81: Just to double check, is there a log line somewhere that says that connector [MONGO] picked up job [12safkjslat1]? If no, nice to add here

lib/core/jobs/consumer.rb
```diff
@@ -0,0 +1,120 @@
...
+        # reset pool
+        @pool = nil
+      end
+
+      private
+
+      def start_loop!
+        Utility::Logger.info("Starting a new consumer for #{@index_name} index")
+
+        Thread.new do
```
artem-shelkovnikov, line 51: Can the thread actually exit and die somehow? I see you rescue StandardError, but Exceptions can cause it to die and stop reporting. While Exceptions probably indicate that connector won't actually run at all, it might be valuable to keep receiving errors and trying again so that user in Cloud would be able to see errors in their deployment instead of seeing it as running but not producing proper logs.

lib/core/jobs/consumer.rb
```diff
@@ -0,0 +1,120 @@
...
+          max_queue: @max_queue,
+          fallback_policy: :abort,
+          idletime: @idle_time
+        )
+      end
+
+      # @TODO replace it with the direct API call
+      def scheduler
+        App::Dispatcher.send(:scheduler)
+      end
```
artem-shelkovnikov, line 111: Hmmm why not pass scheduler into initializer?

lib/app/dispatcher.rb
vidok: Got it. I think we can label this discussion as tech-debt

lib/core/connector_job.rb
vidok: It's okay to have a null-pointer issue here because I trust documents at this code level. If the document doesn't have `connector.id` then we can't process it and the job should fail.

lib/core/connector_job.rb
vidok: Yes, you're right. 

I must say the method name is misleading :). 

would `with_concurrency_control` work? any ideas?

lib/core/connector_job.rb
vidok: Yeah, it changes the job statuses. The method naming comes from https://github.com/state-machines/state_machines which might be a bit confusing. After your question, I also find it weird. 

Do you think I can rename it to something like `job.make_running!` or `job.running!`. open for any ideas :)

lib/core/connector_settings.rb
vidok: I would not change it until(if) we stop relying on connectors document to determine the job execution state.

lib/app/dispatcher.rb
vidok: Yes, we might be running multiple consumers. I will rename it though

lib/core/connector_job.rb
artem-shelkovnikov: We agreed on `job.make_running!`

lib/core/jobs/consumer.rb
vidok: I will follow up with a different PR to address this problem. 

`TimerTask` from ruby-concurrent is a good option.

spec/core/elastic_connector_actions_spec.rb
```diff
@@ -601,165 +601,165 @@ def filter_validation_result(domain, state, errors)
...
-        expect(es_client).to receive(:index).with(
-          :index => jobs_index,
-          :body => hash_including(:connector => hash_including(:filtering => [job_filtering])),
-          :refresh => true
-        )
-        described_class.claim_job(connector_id)
-      end
-    end
-  end
+  # describe '#claim_job' do
```
artem-shelkovnikov, line 604: As discussed with @vidok, these tests will be returned back after this PR is merged.

lib/core/jobs/consumer.rb
```diff
@@ -0,0 +1,114 @@
...
+              rescue StandardError => e
+                Utility::ExceptionTracking.log_exception(e, "Sync job for #{connector_settings.formatted} failed due to unexpected error.")
+              end
+            end
+          rescue StandardError => e
+            Utility::ExceptionTracking.log_exception(e, 'The consumer group failed')
+          end
+        end
+
+        @running.make_true
```
wangch079, line 94: I think this should be moved to the top of `start_loop!`:
```ruby
def start_loop!
  unless @running.make_true
    Utility::Logger.warn("A job consumer is already started.")
    return
  end
  ...

Otherwise, calling `start_loop!` twice maybe able to start 2 threads.
```

lib/core/jobs/consumer.rb
wangch079: The consumer in my mind:
```ruby
connector_id = App::Config.native_mode ? nil : App::Config.connector_id
# pending_jobs will call fetch_native_connectors to get supported native connector ids.
# I don't feel necessary to pass in scheduler, as it won't reduce number of ES query.
pending_jobs = Core::ConnectorJob.pending_jobs(connector_id)
pending_jobs.each do |job|
  unless job.connector
    Utility::Logger.warn("couldn't found connector")
    job.error!
    next
  end
  # I'm not sure about this, we have no way to update `index_name` connector, 
  # and once connector in ready for sync states, it won't go back to created or needs_configuration
  # even we need this check, how do we handle the job?
  unless job.connector.ready_for_sync
    Utility::Logger.warn("connector is not ready for sync")
    next
  end

  pool.post do
    SyncJobRunner.new(job.connector, job).execute
  end
end
```

PR #428, opened 2022-11-14
Add Job cleanup task

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/3199

This PR adds a Job cleanup task, which runs by default every 5 mins, to 
1. delete all the orphaned jobs and their content indices. An orphaned job is defined as a job whose associated connector is deleted.
2. mark stuck jobs as error. A stuck job is defined as a job which has seen no update for 1 minute. The task will only process stuck jobs of supported connectors.

#### Pre-Review Checklist


lib/core/job_cleanup.rb
```diff
@@ -0,0 +1,64 @@
...
+require 'utility/logger'
+
+module Core
+  class JobCleanUp
+    class << self
+
+      def execute(connector_id = nil)
+        # clean up orphaned jobs
+        Utility::Logger.info("Start cleaning up orphaned jobs for #{connector_id ? "connector #{connector_id}" : 'native connectors'}...")
+        orphaned_jobs = connector_id ? orphaned_jobs_for_single_connector(connector_id) : orphaned_jobs_for_native_connectors
```
seanstory, line 19: if a non-native connector is only cleaning up orphaned jobs that it is directly associated with, does that not mean that an operator needs to:
1. delete a connector
2. wait a bit
3. stop their connector instance

in that particular order? If they instead:

1. stop their connector instance
2. delete the index/connector

the orphaned job record will never be cleaned.

lib/core/connector_job.rb
```diff
@@ -27,12 +28,38 @@ def self.pending_jobs(page_size = DEFAULT_PAGE_SIZE)
...
+                        [connector_id]
+                      else
+                        ConnectorSettings.fetch_native_connectors.map(&:id)
+                      end
+      query = {
+        bool: {
+          filter: [
+              { terms: { 'connector.id': connector_ids } },
+              { terms: { status: Connectors::SyncStatus::ACTIVE_STATUSES } },
+              { range: { last_seen: { lte: "now-#{STUCK_THRESHOLD}s" } } }
```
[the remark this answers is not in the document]
wangch079: >How frequently do we expect `last_seen` to be updated?

Every 10 seconds
https://github.com/elastic/connectors-ruby/blob/b3cc1332879a38930a272a63f8c6be1847578204/lib/core/sync_job_runner.rb#L24

>30 seconds doesn't feel like a very long time.
@danajuratoni suggested 1 minute [here](https://github.com/elastic/enterprise-search-team/issues/3199#issuecomment-1314068902), and I will make the change.

>Especially if two server clocks are not perfectly synced.

Good point.

Note, both should be configurable in `_meta` of  `.elastic-connectors`, but we deferred it post 8.6

lib/core/job_cleanup.rb
```diff
@@ -0,0 +1,64 @@
...
+require 'core'
+require 'utility/logger'
+
+module Core
+  class JobCleanUp
+    class << self
+
+      def execute(connector_id = nil)
+        # clean up orphaned jobs
+        Utility::Logger.info("Start cleaning up orphaned jobs for #{connector_id ? "connector #{connector_id}" : 'native connectors'}...")
```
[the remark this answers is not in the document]
wangch079: Good point, will change.

lib/core/job_cleanup.rb
wangch079: That is true.

Ideally, when an operator deletes a connector in Kibana, all the associated jobs should be deleted. Is this the case in 8.6? cc @sphilipse 

I explicitly stated in the implementation doc, that a connector service should only clean up jobs, the service types of which are supported. But thinking twice, we should probably just delete all orphaned jobs, regardless of their service types. It does no harm in both product-level (no orphaned jobs will be left as long as there's at least one connector service running) and engineering-level (we use `terms` query and `delete_by_query` so we don't need to worry about race condition) cc @danajuratoni @tarekziade

lib/core/job_cleanup.rb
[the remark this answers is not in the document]
wangch079: Oh, I thought you were referring to this `info`
https://github.com/elastic/connectors-ruby/blob/6849bb57eb4e7cdd5ba477502b485bf0cc605f95/lib/core/job_cleanup.rb#L36-L38

which I think makes sense to change to `debug`

The log `Start cleaning up orphaned jobs` and `Start cleaning up stuck jobs` indicates a cleanup starts, similar to 
https://github.com/elastic/connectors-ruby/blob/b3cc1332879a38930a272a63f8c6be1847578204/lib/app/dispatcher.rb#L30

lib/core/job_cleanup.rb
```diff
@@ -0,0 +1,65 @@
...
+        end
+
+        marked_count = 0
+        stuck_jobs.each do |job|
+          job.error!('The job has not seen any update for some time.')
+          Utility::Logger.debug("Successfully marked job #{job.id} as error.")
+
+          job_id = job.id
+          job = ConnectorJob.fetch_by_id(job_id)
+          Utility::Logger.warn("Could not found job by id #{job_id}") if job.nil?
```
vidok, line 53: I wonder in what cases `job.id` is not presented.

lib/core/job_cleanup.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        # delete content indicies in case they are re-created by sync job
+        content_indices = orphaned_jobs.map(&:index_name).compact.uniq
+        ElasticConnectorActions.delete_indices(content_indices) if content_indices.any?
+        result = ConnectorJob.cleanup_jobs(orphaned_jobs)
+        Utility::Logger.error("Error found when deleting jobs: #{result['failures']}") if result['failures']&.any?
+        Utility::Logger.info("Successfully deleted #{result['deleted']} out of #{result['total']} orphaned jobs.")
+      end
+
+      def cleanup_stuck_jobs(connector_id = nil)
```
vidok, line 38: I think the method name is misleading because the code doesn't clean up the jobs but marks them as failed.

lib/core/job_cleanup.rb
wangch079: When the job is somehow deleted after `job.error!`.

`ConnectorJob.fetch_by_id` can return `nil`, so I want to be more defensive.

lib/core/job_cleanup.rb
wangch079: Agree. Will rename to `process_orphaned_jobs` and `process_stuck_jobs`

lib/core/job_cleanup.rb
vidok: I still don't get it. If I look at `ConnectorJob.stuck_jobs` implementation I see that you query the sync jobs index. This makes me think that all job entities have to have `id` because all Elastic documents have `id`.

lib/core/job_cleanup.rb
```diff
@@ -0,0 +1,65 @@
...
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+require 'core'
+require 'utility/logger'
+
+module Core
+  class JobCleanUp
```
vidok, line 13: Would it be possible to change the class name to `CleanUpJob` to follow the convention where the main noun is at the end of the name like: 

```
ConnectorJob
ElasticConnectorActions
Utility::Logger
SyncJobRunner
```

Another idea is to use `Jobs` namespaces that I created, in that case, it can look like `Core::Jobs::CleanUpJob`. I think it's a good place for all jobs from my point of view.

wangch079: I'm not a fan of using `verb` as a class name. What do you think of `JobCleaner`?

lib/core/job_cleanup.rb
wangch079: https://github.com/elastic/connectors-ruby/blob/61fd7d3a53e94ee995dadebb77aea49c09185dd6/lib/core/job_cleanup.rb#L48-L53

I'm not worrying about `job_id` in line 51 being `nil`. Line 53 is for the rare case `ConnectorJob.fetch_by_id(job_id)` returns `nil`, i.e. after `job.error!` in line 48, somehow the job doc is deleted. Because `fetch_by_id` can return `nil`, I want to be more defensive. In practice, this probably will never happen.

lib/app/dispatcher.rb
```diff
@@ -70,6 +71,17 @@ def scheduler
                        end
       end
 
+      def job_cleanup_timer
```
artem-shelkovnikov, line 74: I'm not sure if we want more threads, this is +2 threads effectively to the process, one of which does read/write operations on Elasticsearch, potentially producing more concurrency problems. Can we also do this in main thread, since these operations are lightweight?

lib/core/connector_job.rb
```diff
@@ -34,12 +36,33 @@ def self.pending_jobs(connectors_ids: [], page_size: DEFAULT_PAGE_SIZE)
...
+      connector_ids = ConnectorSettings.fetch_all_connectors.map(&:id)
+      query = { bool: { must_not: { terms: { 'connector.id': connector_ids } } } }
+      fetch_jobs_by_query(query, page_size)
     end
 
-    def self.stuck_jobs(_page_size = DEFAULT_PAGE_SIZE)
-      []
+    def self.delete_jobs(jobs)
+      query = { terms: { '_id': jobs.map(&:id) } }
+      ElasticConnectorActions.delete_jobs_by_query(query)
```
artem-shelkovnikov, line 47: I don't think we should ever delete anything. Mark as failed with cause "orphaned"?

lib/core/connector_job.rb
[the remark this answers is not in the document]
artem-shelkovnikov: I'd put 5 minutes as a reasonable default TBH

lib/core/job_cleanup.rb
artem-shelkovnikov: CleanUp is a "noun as adjective" here IMO

lib/core/job_cleanup.rb
```diff
@@ -0,0 +1,65 @@
...
+        Utility::Logger.info('Start cleaning up orphaned jobs...')
+        orphaned_jobs = ConnectorJob.orphaned_jobs
+        if orphaned_jobs.empty?
+          Utility::Logger.info('No orphaned jobs found. Skipping...')
+          return
+        end
+
+        # delete content indicies in case they are re-created by sync job
+        content_indices = orphaned_jobs.map(&:index_name).compact.uniq
+        ElasticConnectorActions.delete_indices(content_indices) if content_indices.any?
```
artem-shelkovnikov, line 32: I'm confused - I don't expect a class called "JobCleanUp" to delete content indices. Maybe should be renamed to something like "OrhanedIndicesCleanUp" and done separately from Job clean up?

lib/core/connector_job.rb
artem-shelkovnikov: Oh, right, these are jobs that are there while connector record has been deleted, disregard this comment.

lib/core/job_cleanup.rb
```diff
@@ -0,0 +1,65 @@
...
+        result = ConnectorJob.delete_jobs(orphaned_jobs)
+        Utility::Logger.error("Error found when deleting jobs: #{result['failures']}") if result['failures']&.any?
+        Utility::Logger.info("Successfully deleted #{result['deleted']} out of #{result['total']} orphaned jobs.")
+      end
+
+      def process_stuck_jobs(connector_id = nil)
+        Utility::Logger.info("Start cleaning up stuck jobs for #{connector_id ? "connector #{connector_id}" : 'native connectors'}...")
+        stuck_jobs = ConnectorJob.stuck_jobs(connector_id)
+        if stuck_jobs.empty?
+          Utility::Logger.info('No stuck jobs found. Skipping...')
```
artem-shelkovnikov, line 42:
```suggestion
          Utility::Logger.debug('No stuck jobs found. Skipping...')
```

lib/core/job_cleanup.rb
```diff
@@ -0,0 +1,65 @@
...
+        # delete content indicies in case they are re-created by sync job
+        content_indices = orphaned_jobs.map(&:index_name).compact.uniq
+        ElasticConnectorActions.delete_indices(content_indices) if content_indices.any?
+        result = ConnectorJob.delete_jobs(orphaned_jobs)
+        Utility::Logger.error("Error found when deleting jobs: #{result['failures']}") if result['failures']&.any?
+        Utility::Logger.info("Successfully deleted #{result['deleted']} out of #{result['total']} orphaned jobs.")
+      end
+
+      def process_stuck_jobs(connector_id = nil)
+        Utility::Logger.info("Start cleaning up stuck jobs for #{connector_id ? "connector #{connector_id}" : 'native connectors'}...")
```
artem-shelkovnikov, line 39:
```suggestion
        Utility::Logger.debug("Start cleaning up stuck jobs for #{connector_id ? "connector #{connector_id}" : 'native connectors'}...")
```

lib/core/job_cleanup.rb
```diff
@@ -0,0 +1,65 @@
...
+          Utility::Logger.info('No orphaned jobs found. Skipping...')
+          return
+        end
+
+        # delete content indicies in case they are re-created by sync job
+        content_indices = orphaned_jobs.map(&:index_name).compact.uniq
+        ElasticConnectorActions.delete_indices(content_indices) if content_indices.any?
+        result = ConnectorJob.delete_jobs(orphaned_jobs)
+        Utility::Logger.error("Error found when deleting jobs: #{result['failures']}") if result['failures']&.any?
+        Utility::Logger.info("Successfully deleted #{result['deleted']} out of #{result['total']} orphaned jobs.")
```
artem-shelkovnikov, line 35: I think it'll make it better if we log here anything only if something was really deleted, to not produce too many logs.

lib/core/job_cleanup.rb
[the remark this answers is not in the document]
seanstory: 👍 sounds good

lib/app/dispatcher.rb
wangch079: I don't quite understand. This method starts a `TimerTask`, which will run one more thread, not two.

The main thread loops through `.elastic-connectors`, and the job cleanup is targeting `.elastic-connectors-sync-jobs`.

lib/core/connector_job.rb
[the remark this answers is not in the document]
wangch079: The heartbeat for a job is 10 seconds. If a job is stuck, it will take 5 mins to be identified. What should be the interval to run the clean-up task (the current default is 5 mins). 

With the above, a stuck job will only be cleaned worst case 10 mins. I will leave this to PM @danajuratoni :

How long we should wait to mark a job as stuck - m mins.
How often we should run the clean-up job? - n mins.

Ideally, a job will be marked as stuck after m mins without updating. In reality, it will be from m mins (best case) to (m+n) mins (worst case)

lib/core/job_cleanup.rb
wangch079: This was discussed [here](https://elastic.slack.com/archives/C03GAPTQ2JF/p1667984997829019)

If a connector is deleted during job execution, the content index may be recreated, before the connector service stops execution.

That's why we want to make sure the content indices of orphaned jobs are deleted as well.

lib/core/job_cleanup.rb
wangch079: You are the 2nd one who suggests I change it to `debug`.

As orphaned/stuck jobs are rare, I think it makes sense. 

I will leave below as it is, so that we will know how many jobs are processed if any.

https://github.com/elastic/connectors-ruby/blob/64d612512fa214ea5758322705f5609435aaaa4f/lib/core/job_cleanup.rb#L34-L35

https://github.com/elastic/connectors-ruby/blob/64d612512fa214ea5758322705f5609435aaaa4f/lib/core/job_cleanup.rb#L61

lib/core/connector_job.rb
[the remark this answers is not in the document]
danajuratoni: As we don't have any data atm how often stuck jobs would occur and the impact of our checks, I tend to be more conservative with the configured times. The important thing is to adjust them when we have more insights.
From a product perspective, I'd suggest to use 1 minute for stuck jobs and 5 minutes for the cleanup job and make sure in 8.7 we have telemetry to capture related data.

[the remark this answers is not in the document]
artem-shelkovnikov: If you run MongoDB connector against our cloud instance sometimes new batch of docs take 3 minutes to get back. In reality I see some connectors hang for up to 5 minutes for some weird systems that throttle by keeping open connection and sending no data, thus I proposed 5 minutes.

Rrunning job clean up every hour or 24 hours might be good enough, but it's not a strong opinion.

lib/core/job_cleanup.rb
artem-shelkovnikov: We probably need to revisit this logic ASAP when cancellation is available and not allow to delete indices that have ongoing work, that'll make this code redundant and logic a bit simpler.

If people would want to do manual deletion of jobs/indices, they can break it, but that's okay IMO

lib/core/job_cleanup.rb
artem-shelkovnikov: What my previous comment meant is "I like CleanupJob" a bit more, but it's opinion

lib/app/dispatcher.rb
artem-shelkovnikov: TimerTask as far as I remember is actually 2 threads: 1 monitoring thread that checks if the function inside the timer should be triggered + 1 executor thread that actually runs the callback specified in the timer.

> The main thread loops through .elastic-connectors, and the job cleanup is targeting .elastic-connectors-sync-jobs.

Indeed, my comment was mostly from perspective of reduction of complexity of the system - now we have operations happening in multiple threads while in reality we can have only 1 main thread for producer that does all the logic about connectors/jobs faciliation + 1 thread per sync process.

wangch079: I re-read the [TimerTask doc](https://ruby-concurrency.github.io/concurrent-ruby/1.1.4/Concurrent/TimerTask.html), and you are right:
>When a `TimerTask` is launched it starts a thread for monitoring the execution interval. The `TimerTask` thread does not perform the task, however. Instead, the `TimerTask` launches the task on a separate thread.

However, I'm not so concerned about the race condition.
1. Orphaned jobs will be simply deleted.
2. Stuck Jobs will be updated. With proper `stuck_threshold` configured, most of, if not all, the stuck jobs are really in stuck, while the job consumer only process pending jobs (jobs in `pending` or `suspended` status).

wangch079: In the future, the cleanup task can be added to job consumer. My concern is job consumer is required to be run much more frequently than cleanup, unless we have a place to record when the last clean up runs.

wangch079:
>because this solution looks like a temporary one that needs to be revisited as soon as possible (once cancellation is properly working and is available in UI particularly.

@artem-shelkovnikov , Job cancelation is also part of the Monitoring project and will be delivered in 8.6. Even with job cancelation, I think the job clean up task is still required. We want to make sure there's no orphaned jobs, we want to be able to identify stuck jobs given our async nature of system, and we will also (maybe) add logic to delete stale jobs post 8.6.

artem-shelkovnikov:
My main concern here is that as a user I want to be able to immediately create an index with the same name that I just deleted. With current approach this is not possible and I believe we can make it better with simplifying our code base as well.

wangch079:
>My main concern here is that as a user I want to be able to immediately create an index with the same name that I just deleted. With current approach this is not possible and I believe we can make it better with simplifying our code base as well.

@artem-shelkovnikov 100% agree. When a user deletes a connector, I don't Kibana will delete the jobs but I think it should.

But these are not related to the job clean up task. The connector and job index are open for modification via Dev Tools and Elasticsearch API, and the job clean up task serves a last resort to make sure our indices are in a healthy state.

artem-shelkovnikov (approved):
I'm putting approve here but it took me a lot of time to think about the problems that are addressed by this PR, and I think current way of addressing needs to be improved in future.

Concerns:

1. Action of index deletion can affect users in unpredictable ways when something goes wrong. For example - user deleted an index 'search-a', connectors framework crashed, user created same index 'search-a' again, connectors framework restarted, found orphaned job and deleted this index. As a result user has in UI the fact that index 'search-a' has connector attached to it, but data was deleted. It'll be recreated later, sure, but it can cause weird things to happen with data/UI.
2. I'm not sure if we even want connectors framework to do any index management - we removed ways to create connectors index, job index, and expect Kibana to do index creation/management for us. I think that best we can do is - not allow deletion of index if connectors are running against it, let user manage it, or even a different way of approaching same problem. Current solution feels fragile to me.

So TLDR - it addresses some problems that we have now with deletion-recreation of the index, however we need to follow up soon IMO.

lib/core/job_cleanup.rb
vidok: Okay, here is my concern: if the code handles issues based on `In practice, this probably will never happen.` statement then we put ourselves in a situation where we just can't see the issues. From my point of view if we expect a specific behavior from our system then we trust it and if `neven happen` **does happen** then we should be able to see exceptions and we should fix them.

So I don't get why the job document can disappear after `job.error!` and if it really does, I'd like to fix that because it shouldn't :)

lib/core/job_cleanup.rb
wangch079: Oh, I didn't notice this. I think at least @vidok and @artem-shelkovnikov are voting `CeanUpJob`, I'm not sure about the 👍 from @seanstory , but I'm going to rename it.

I also feel it makes sense to put it under the module `Jobs`. Should we name it `Cleaner`, as consistent with `Producer` and `Consumer`? Or stick with `CleanUp` (or `Cleanup`, so the file name of `clean_up.rb` vs `cleanup.rb`)?

PR #429, opened 2022-11-14
Return nil instead of raising error if the connector/job is not found

wangch079:
`verify_by_id` method should return `nil` instead of raising an error when the connector or job is not found.

#### Pre-Review Checklist


artem-shelkovnikov:
What is the problem that is being solved in the PR?

I see that now connector won't generate an error (if it did), but is it a correct way to go?

wangch079:
> What is the problem that is being solved in the PR?

I'm working on job cancellation and job clean-up tasks. At any time, a job or a connecter can be non-existing. With the current implementation (raising error), I need to catch it whenever I reload connecotr/job.

lib/core/connector_job.rb
```diff
@@ -62,16 +63,36 @@ def canceling?
...
+      Connectors::SyncStatus::ACTIVE_STATUSES.include?(status)
+    end
+
+    def terminated?
+      Connectors::SyncStatus::TERMINAL_STATUSES.include?(status)
+    end
+
     def connector_snapshot
-      self[:connector]
+      self[:connector] || {}
```
artem-shelkovnikov, line 87: Why default to an empty object instead of nil?

lib/core/connector_job.rb
```diff
@@ -62,16 +63,36 @@ def canceling?
...
     end
 
     def connector_id
       connector_snapshot[:id]
     end
 
     def index_name
-      connector_snapshot[:configuration]
+      connector_snapshot[:index_name]
     end
```
artem-shelkovnikov, line 96: I see, so `connector_snapshot` defaults to empty object cause of these lines.

I think it's just making code/debugging harder: the code that tries to get id/index_name for non-existing connector snapshot will return empty values, while in reality it should not have attempted to access connector snapshot id/index name.

I think it's worth to only return `connector_snapshot` from the job and let the user of it handle nulls or return a wrapper class that encapsulates some of its logic.

That is a non-blocking comment though, this refactor can be done later.

lib/core/connector_job.rb
```diff
@@ -133,5 +148,24 @@ def self.fetch_jobs_by_query(query, page_size)
...
       results
     end
+
+    def initialize(es_response)
+      # TODO: remove the usage of with_indifferent_access. The initialize method should expect a hash argument
+      @elasticsearch_response = es_response.with_indifferent_access
+    end
+
+    def terminate!(status, error = nil, ingestion_stats = {}, connector_metadata = {})
+      ingestion_stats ||= {}
```
artem-shelkovnikov, line 158: What's the case when ingestion stats are empty?

lib/core/connector_job.rb
```diff
@@ -133,5 +148,24 @@ def self.fetch_jobs_by_query(query, page_size)
...
+      ingestion_stats[:total_document_count] = ElasticConnectorActions.document_count(index_name)
```
artem-shelkovnikov, line 159: This part looks a bit scary - it should already come from ingestion stats, right?

lib/core/elastic_connector_actions.rb
```diff
@@ -507,31 +507,15 @@ def ensure_job_index_exists
...
-        end
+        update_doc_fields(Utility::Constants::CONNECTORS_INDEX, connector_id, doc, seq_no, primary_term)
+      end
+
+      def update_job_fields(job_id, doc = {}, seq_no = nil, primary_term = nil)
+        update_doc_fields(Utility::Constants::JOB_INDEX, job_id, doc, seq_no, primary_term)
       end
 
       def document_count(index_name)
+        client.indices.refresh(:index => index_name)
```
artem-shelkovnikov, line 518: Does this guarantee immediate refresh? Or just starts a refresh?

lib/core/single_scheduler.rb
```diff
@@ -20,7 +20,7 @@ def initialize(connector_id, poll_interval, heartbeat_interval)
 
     def connector_settings
       connector_settings = Core::ConnectorSettings.fetch_by_id(@connector_id)
-      [connector_settings]
+      [connector_settings].compact
```
artem-shelkovnikov, line 23: That place might be worth a test

lib/core/connector_job.rb
artem-shelkovnikov, line 158: This method is not tested and I believe it should be - it has some logic about defaulting `intestion_stats`, `total_document_count`, conditional `cancelled_at` and `metadata` assignments.

artem-shelkovnikov (approved):
I don't want to block further development, so approving the PR, but I've left some comments to address when you have time - in this or a follow-up PR

lib/core/connector_job.rb
wangch079: Resolve this conversation, and reply in another comment.

lib/core/connector_job.rb
wangch079: I added `|| {}` defensively.

The snapshot data (e.g. `configuration`, `pipeline`, etc..) will be used in job execution, and they are expected to be there (the last item of https://github.com/elastic/enterprise-search-team/issues/3283#issue-1442211849). What I will do is something like this:
```ruby
  pipeline = job.pipeline || connector_settings.pipeline
```

lib/core/connector_job.rb
wangch079: When `job.terminate!('garbage', nil, nil, nil)`, which shouldn't happen, just be defensive.

lib/core/connector_job.rb
wangch079: No, ingestion stats only report how many docs are ingested by this job. This is how many docs are in the connector at the end of the job, requested by @sphilipse . And that's why it's only updated when `terminate!` is called (i.e. terminal status like `complete`, `canceled` or `error`)

lib/core/elastic_connector_actions.rb
wangch079: Yes. From the [Refresh API doc](https://www.elastic.co/guide/en/elasticsearch/reference/8.5/indices-refresh.html#refresh-api-desc):

>Refresh requests are synchronous and do not return a response until the refresh operation completes.

lib/core/single_scheduler.rb
wangch079: Will add test soon

lib/core/connector_job.rb
wangch079: Will add test soon

spec/core/connector_job_spec.rb
```diff
@@ -37,4 +69,93 @@
       expect(expected_job_ids).to eq(actual_job_ids)
     end
   end
+
+  # TODO: #done! example should be made as a shared example and tested by #done!, #cancel! and #error!, with only status and error set differently in each example.
```
wangch079, line 73: Hi @artem-shelkovnikov , as `#done!`, `#cancel!` and `#error!` call the same `terminate!` method, most of the test cases (i.e. the current `describe '#done!'`) should be grouped as a shared example, with the only additional test is then when `#cancel!` is called, `canceled_at` is also updated.

I tried to work on the shared example, but it doesn't work. basically, I have difficulty passing proc to the shared example, e.g.  `subject.done!(ingestion_stats, connector_metadata)`. I also tried to put it in `before` block it doesn't work either.

It cost me about one hour, and i decided to add a `TODO` and proceed.

PR #430, opened 2022-11-14
First pass at [Ruby][Mongo] validate simple filter rules #3174

mchernyavskaya:
## Closes https://github.com/elastic/enterprise-search-team/issues/3174

Basic high-level validation of simple filtering rules. It has a lot of edge cases and we can't cover all, so it's a best-effort attempt to provide at least some rough coverage.

The PR also includes pre-processing of `start_with` and `end_with` rules in mongo (the [first PR for simple rules](https://github.com/elastic/connectors-ruby/pull/410) didn't cover those as they were specified as post-processing in the implementation document, but since technically it's just a kind of a regex, I didn't see why we can't do it in pre-processing for Mongo).

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/connectors-ruby/pull/410 - simple filtering rules
* https://github.com/elastic/connectors-ruby/pull/412 - post-processing rules


lib/core/filtering/simple_rule.rb
```diff
@@ -34,26 +36,50 @@ class Rule
...
+        @value = SimpleRule.flex_fetch(rule_hash, VALUE)
+        @id = SimpleRule.flex_fetch(rule_hash, ID)
+        @order = SimpleRule.flex_fetch(rule_hash, ORDER, 0)
         @rule_hash = rule_hash
       rescue KeyError => e
-        raise "#{e.key} is required"
+        raise "#{e.key} is required: #{e.message}"
+      end
+
+      def self.flex_fetch(hash, key, default_value = nil)
```
seanstory, line 56: I believe @vidok [said this will kill kittens](https://elastic.slack.com/archives/C03GAPTQ2JF/p1668507180031719?thread_ts=1668448379.617669&cid=C03GAPTQ2JF)

lib/core/filtering/simple_rule.rb
```diff
@@ -34,26 +36,50 @@ class Rule
...
-        @id = rule_hash.fetch(ID)
+        @policy = SimpleRule.flex_fetch(rule_hash, POLICY)
+        unless @policy == Policy::INCLUDE || @policy == Policy::EXCLUDE
+          raise "Invalid policy '#{policy}' for rule '#{rule_hash}'"
+        end
+        @field = SimpleRule.flex_fetch(rule_hash, FIELD)
+        @rule = SimpleRule.flex_fetch(rule_hash, RULE)
+        @value = SimpleRule.flex_fetch(rule_hash, VALUE)
+        @id = SimpleRule.flex_fetch(rule_hash, ID)
+        @order = SimpleRule.flex_fetch(rule_hash, ORDER, 0)
```
seanstory, line 50: If we're going to add `order` here, I suggest we require it, and not provide a default value. Otherwise we risk non-deterministic ordering in specs if we don't have an explicit order field defined and they all get `:order => 0` from this default.

lib/connectors/base/simple_rules_parser.rb
```diff
@@ -11,23 +11,144 @@
 
 module Connectors
   module Base
+    class FilteringRulesValidationError < StandardError; end
```
seanstory, line 14: should this go in errors.rb?

lib/connectors/base/simple_rules_parser.rb
```diff
@@ -11,23 +11,144 @@
...
     class SimpleRulesParser
+
+      include Core::Filtering
+
+      attr_reader :rules
+
       def initialize(rules)
-        @rules = (rules || []).map(&:with_indifferent_access).filter { |r| r[:id] != 'DEFAULT' }.sort_by { |r| r[:order] }
+        begin
+          sorted = (rules || []).map { |r| SimpleRule.new(r) }.filter { |r| r.id != 'DEFAULT' }.sort_by(&:order)
```
seanstory, line 24: I like this change. I'd avoided it because of the `order` requirement, but you've solved that now. 
Should we unify this sorting with the one done by the PostProcessingEngine to DRY up our logic?

lib/connectors/base/simple_rules_parser.rb
```diff
@@ -11,23 +11,144 @@
...
+        end
+
+        # # check for overlapping ranges
+        # ranges = field_rules.filter { |r| r[:rule] == '>' || r[:rule] == '<' }
+        # ranges.each_with_index do |r, i|
+        #   next if i == ranges.size - 1
+        #   next_r = ranges[i + 1]
+        #   if r[:value] == next_r[:value]
+        #     raise FilteringRulesValidationError.new("Contradicting rules for field: #{field}. Can't have overlapping ranges.")
+        #   end
```
seanstory, line 79: should this block be removed?

lib/connectors/base/simple_rules_parser.rb
```diff
@@ -11,23 +11,144 @@
...
+          raise FilteringRulesValidationError, "Invalid rule: #{e.message}"
+        end
+        @rules = validate(sorted)
       end
 
       def parse
-        merge_rules(@rules.map do |rule_hash|
-          rule = Core::Filtering::SimpleRule.new(rule_hash)
-          unless rule.is_include? || rule.is_exclude?
-            raise "Unknown policy: #{rule.policy}"
```
seanstory, line 23: Did we lose the `"unknown policy"` check?

lib/connectors/mongodb/mongo_rules_parser.rb
```diff
@@ -16,24 +16,22 @@ class MongoRulesParser < Connectors::Base::SimpleRulesParser
...
-        unless value.present?
-          raise "value is required for field: #{field}"
-        end
-        unless field.present?
-          raise "field is required for rule: #{rule}"
-        end
         op = rule.rule
         case op
-        when Core::Filtering::SimpleRule::Rule::EQUALS
+        when SimpleRule::Rule::EQUALS
```
seanstory, line 21: How did you shorten these? I thought you could only omit the fully-qualified prefix if these were in the same module?

spec/core/filtering/simple_rule_spec.rb
```diff
@@ -39,7 +39,7 @@
 
   shared_examples_for 'a match' do
     it 'matches' do
-      expect(subject.match?(document)).to be
+      expect(subject.match?(document)).to be_truthy
```
seanstory, line 42: :P

lib/connectors/base/simple_rules_parser.rb
mchernyavskaya: Yeah it should, it's a WIP to document the thought process. Frankly, the ranges feel so cumbersome to validate... there's just too many cases.

lib/connectors/mongodb/mongo_rules_parser.rb
mchernyavskaya: I did `include Core::Filtering` in the class. =) (See row 18 above for the `lib/connectors/base/simple_rules_parser.rb`).

lib/core/filtering/simple_rule.rb
mchernyavskaya: My thinking was that if we don't have order, then the ordering probably doesn't matter (like it doesn't now for pre-processing) and we can safely assume some default value. But you're probably right.

lib/core/filtering/simple_rule.rb
mchernyavskaya: Yeah if we do it in a monkey patch. =) But here, we don't.
Frankly, I don't see a good way to do it with key types unknown. Moreover, even the 
`hash[key] || hash.fetch(key.to_sym, nil) || hash.fetch(key.to_s, nil)` won't work properly in the case where the key = boolean false (learned the hard way, therefore this ugly code).

spec/core/filtering/simple_rule_spec.rb
mchernyavskaya: yeah IDK what it's supposed to mean but it failed for me 😢

lib/connectors/base/simple_rules_parser.rb
mchernyavskaya: No, the individual rule checks went into rule init.

lib/core/filtering/simple_rule.rb
seanstory: When resolving merge conflicts, I convinced myself we didn't need this. We can just rely on string keys.

seanstory:
@mchernyavskaya I failed to get this ready to merge. It too me too long to figure out how to resolve the merge conflicts (I couldn't rebase your branch and I still don't really understand why), and I'm just out of time. I think we'll just need to save this issue for 8.7 :(

mchernyavskaya:
@seanstory no worries. Taking out of the DM: I feel like I'm going into a rabbit hole with these filtering rules (and specifically, with validation). I'm second-guessing this pre and post processing model more and more, so maybe that's something we should revisit.

artem-shelkovnikov:
Checked the PR and spoke to @mchernyavskaya about it. It indeed feels like a rabbit hole - number of possible combinations of invalid rules is potentially really high, the complexity of validating the rules is humongous.

I think same problem can be probably approached from a different perspective - let users enter any rules they want but provide good logging and tracing to see what query was generated (maybe even why) and why the expected data was not ingested.

It's probably good to validate each individual rule though just for sanity purposes - for example that RANGE filter is actually [X; Y], not [-INF, X], [Y; +INF]

mchernyavskaya:
@artem-shelkovnikov yep - I also proposed something like this [in Ingestion Sync agenda](https://docs.google.com/document/d/1GlB-_GLx36VobV-kQXPyhFjH6VynT1e8T1Y045xZ6Vo/edit#bookmark=id.5ptdt2tatt1t) - want it at least to be up for a discussion.

seanstory:
I believe we can close this now, as it was implemented in a separate PR

PR #432, opened 2022-11-15
Start sync when scheduling is enabled and properly configured and there's no sync before

wangch079:
This is a bug reported in [slack](https://elastic.slack.com/archives/C03GAPTQ2JF/p1668460076978669). An initial sync will be triggered even operator never manually clicks `Sync` button or schedule a sync job.

This PR fixes the bug, so that the initial job will be triggered, only when operators explicitly enable and configure a schedule.

#### Pre-Review Checklist

### DIFF
--- a/lib/core/scheduler.rb
+++ b/lib/core/scheduler.rb
@@ -119,7 +119,7 @@ module Core
         return true
       end
 
-      next_trigger_time = cron_parser.next_time(last_synced)
+      next_trigger_time = cron_parser.next_time(Time.parse(last_synced))
 
       # Sync if next trigger for the connector is in past
       if next_trigger_time < Time.now
--- a/lib/core/scheduler.rb
+++ b/lib/core/scheduler.rb
@@ -112,14 +112,12 @@ module Core
         return false
       end
 
-      # We want to sync when sync never actually happened
-      last_synced = connector_settings[:last_synced]
-      if last_synced.nil? || last_synced.empty?
-        Utility::Logger.info("#{connector_settings.formatted.capitalize} has never synced yet, running initial sync.")
-        return true
+      last_synced = begin
+        Time.parse(connector_settings[:last_synced])
+      rescue StandardError
+        Time.now
       end
-
-      next_trigger_time = cron_parser.next_time(Time.parse(last_synced))
+      next_trigger_time = cron_parser.next_time(last_synced)
 
       # Sync if next trigger for the connector is in past
       if next_trigger_time < Time.now
--- a/lib/core/scheduler.rb
+++ b/lib/core/scheduler.rb
@@ -90,13 +90,6 @@ module Core
         return false
       end
 
-      # We want to sync when sync never actually happened
-      last_synced = connector_settings[:last_synced]
-      if last_synced.nil? || last_synced.empty?
-        Utility::Logger.info("#{connector_settings.formatted.capitalize} has never synced yet, running initial sync.")
-        return true
-      end
-
       current_schedule = scheduling_settings[:interval]
 
       # Don't sync if there is no actual scheduling interval
@@ -117,6 +110,13 @@ module Core
       unless cron_parser
         Utility::Logger.error("Unable to parse sync schedule for #{connector_settings.formatted}: expression #{current_schedule} is not a valid Quartz Cron definition.")
         return false
+      end
+
+      # We want to sync when sync never actually happened
+      last_synced = connector_settings[:last_synced]
+      if last_synced.nil? || last_synced.empty?
+        Utility::Logger.info("#{connector_settings.formatted.capitalize} has never synced yet, running initial sync.")
+        return true
       end
 
       next_trigger_time = cron_parser.next_time(Time.parse(last_synced))
--- a/lib/core/scheduler.rb
+++ b/lib/core/scheduler.rb
@@ -112,11 +112,13 @@ module Core
         return false
       end
 
-      last_synced = begin
-        Time.parse(connector_settings[:last_synced])
-      rescue StandardError
-        Time.now
+      # We want to sync when sync never actually happened
+      last_synced = connector_settings[:last_synced]
+      if last_synced.nil? || last_synced.empty?
+        Utility::Logger.info("#{connector_settings.formatted.capitalize} has never synced yet, running initial sync.")
+        return true
       end
+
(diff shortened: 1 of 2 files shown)


artem-shelkovnikov:
I thought we need to remove this lines to let connector run on schedule or on sync_now?

With the line moved it looks like it will trigger out of schedule for the first time.

wangch079:
> I thought we need to remove this lines to let connector run on schedule or on sync_now?
> 
> With the line moved it looks like it will trigger out of schedule for the first time.

@artem-shelkovnikov , did you mean remove below?

https://github.com/elastic/connectors-ruby/blob/19276befefb9e74aea29541e6083663bc82bd56d/lib/core/scheduler.rb#L115

artem-shelkovnikov:
> @artem-shelkovnikov , did you mean remove below?
@wangch079 yes, cause it'll still start a sync immediately after user turns on the scheduling, but we should not?

lib/core/scheduler.rb
```diff
@@ -119,6 +112,13 @@ def sync_triggered?(connector_settings)
         return false
       end
 
+      # We want to sync when sync never actually happened
+      last_synced = connector_settings[:last_synced]
+      if last_synced.nil? || last_synced.empty?
```
seanstory, line 117:
```suggestion
      if scheduling_settings[:interval].present? && (last_synced.nil? || last_synced.empty?)
```

I think this is already logically implied, but I think it's easier to reason about if we make this explicit.

seanstory (approved):

wangch079:
>yes, cause it'll still start a sync immediately after user turns on the scheduling, but we should not?

@artem-shelkovnikov , it will start a sync immediately after user turns on the scheduling and **configures a scheduling** (although they couldn't enable scheduling without configuring one via UI). 

We want to start a sync immediately, because we need to calculate the next run from `last_synced`. But I just realized, our schedule is like `run the sync every day at 16:00`, not `run the sync every 10 minutes`. 

I will make the `last_synced` default to now.

lib/core/scheduler.rb
wangch079: With the latest commit https://github.com/elastic/connectors-ruby/commit/59f526ff017e38da596851800bc7f4508f328552, this is not needed.

lib/core/scheduler.rb
```diff
@@ -119,7 +112,12 @@ def sync_triggered?(connector_settings)
         return false
       end
 
-      next_trigger_time = cron_parser.next_time(Time.parse(last_synced))
+      last_synced = begin
+        Time.parse(connector_settings[:last_synced])
+      rescue StandardError
```
vidok, line 117: Do I get it right that you handling a case when `connector_settings[:last_synced]` is empty?

If yes, then I have a question. Do you also want to replace the current `connector_settings[:last_synced]` value by `Time.now` if it's stored in the incorrect format? e.g. 

```Ruby
irb(main):004:0> Time.parse('foo')
        7: from /Users/vidok/.rbenv/versions/2.6.9/bin/irb:23:in `<main>'
        6: from /Users/vidok/.rbenv/versions/2.6.9/bin/irb:23:in `load'
        5: from /Users/vidok/.rbenv/versions/2.6.9/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        4: from (irb):4
        3: from (irb):4:in `rescue in irb_binding'
        2: from /Users/vidok/.rbenv/versions/2.6.9/lib/ruby/2.6.0/time.rb:375:in `parse'
        1: from /Users/vidok/.rbenv/versions/2.6.9/lib/ruby/2.6.0/time.rb:194:in `make_time'
ArgumentError (no time information in "foo")
```

mchernyavskaya (approved):
LGTM

lib/core/scheduler.rb
wangch079: I didn't think of this scenario, but I think it's a good practice.

`last_synced` is set by the connector service, not operators, so I don't expect any surprise. 

Of course, this can be tampered via Dev Tools or ES API. If that happens, I think it also makes sense to use `Time.now` as a fallback. cc @danajuratoni

spec/core/scheduler_spec.rb
```diff
@@ -98,12 +98,6 @@
         it_behaves_like 'does not trigger', :sync
       end
 
-      context 'when connector is never synced' do
-        let(:last_synced) { nil }
-
-        it_behaves_like 'triggers', :sync
-      end
-
```
artem-shelkovnikov, line 106: I think it's nice to have a test for this case - if connector never synced and schedule says "every hour" - when does the trigger happen if it's 10:30? If it's 10:55? If it's 10:02?

From looking at the code I can't answer this question and it would be great to express answer to this question with a test/couple tests.

artem-shelkovnikov (commented):
Overall it looks good, but I've left a comment about a test.

In short I read the changeset and don't understand any more how the logic will work when connector never synced.

spec/core/scheduler_spec.rb
wangch079: When `last_synced = nil`, we will default it to now, and use it to find the next run, which should always be in the future.

But in the test, I don't want to test `Fugit::Cron#next_time`, instead I only want to mock it. 

In this case, what I can test is, when `last_synced = nil`, and mock `Fugit::Cron#next_time` to return a future date, the expected result is it will not trigger a job. This is the same as this example:
https://github.com/elastic/connectors-ruby/blob/59f526ff017e38da596851800bc7f4508f328552/spec/core/scheduler_spec.rb#L121-L125

That's why I didn't add an example.

artem-shelkovnikov: This is what I don't honestly understand :) 

We have a code that triggers sync when the time comes, which is:

```
next_trigger_time = cron_parser.next_time(last_synced)

# Sync if next trigger for the connector is in past
if next_trigger_time < Time.now
  # do sync
end
```

When `last_synced` will be Time.now, then this code will never trigger, how will the scheduled trigger happen?

wangch079: Hey @artem-shelkovnikov , this is a really good point.

wangch079:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.6|[](https://github.com/elastic/connectors-ruby/pull/446)|

### Questions ?

PR #434, opened 2022-11-15
Make it possible to specify maximum size/volume of ingestion queue

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/3096

Now it's possible to specify maximum size/volume of ingestion queue.

```
max_ingestion_queue_size: X - data from the connector will be sent to Elasticsearch once X items were extracted
max_ingestion_queue_bytes: Y - data from the connector will be sent to Elasticsearch once bulk operation reaches Y bytes
```

#### Pre-Review Checklist

### DIFF
--- a/lib/app/dispatcher.rb
+++ b/lib/app/dispatcher.rb
@@ -21,6 +21,8 @@ module App
     MIN_THREADS = (App::Config.dig(:thread_pool, :min_threads) || 0).to_i
     MAX_THREADS = (App::Config.dig(:thread_pool, :max_threads) || 5).to_i
     MAX_QUEUE = (App::Config.dig(:thread_pool, :max_queue) || 100).to_i
+    MAX_INGESTION_QUEUE_SIZE = (App::Config.max_ingestion_queue_size || Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_SIZE).to_i 
+    MAX_INGESTION_QUEUE_BYTES = (App::Config.max_ingestion_queue_bytes || Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_BYTES).to_i
 
     @running = Concurrent::AtomicBoolean.new(false)
 
@@ -90,6 +92,26 @@ module App
         end
       rescue StandardError => e
         Utility::ExceptionTracking.log_exception(e, 'The connector service failed due to unexpected error.')
+      end
+
+      def start_sync_task(connector_settings)
+        start_heartbeat_task(connector_settings)
+        pool.post do
+          Utility::Logger.info("Initiating a sync job for #{connector_settings.formatted}...")
+          Core::ElasticConnectorActions.ensure_content_index_exists(connector_settings.index_name)
+          job_runner = Core::SyncJobRunner.new(
+            connector_settings,
+            MAX_INGESTION_QUEUE_SIZE,
+            MAX_INGESTION_QUEUE_BYTES
+          )
+          job_runner.execute
+        rescue Core::JobAlreadyRunningError
+          Utility::Logger.info("Sync job for #{connector_settings.formatted} is already running, skipping.")
+        rescue Core::ConnectorVersionChangedError => e
+          Utility::Logger.info("Could not start the job because #{connector_settings.formatted} has been updated externally. Message: #{e.message}")
+        rescue StandardError => e
+          Utility::ExceptionTracking.log_exception(e, "Sync job for #{connector_settings.formatted} failed due to unexpected error.")
+        end
       end
 
       def start_heartbeat_task(connector_settings)
--- a/lib/app/dispatcher.rb
+++ b/lib/app/dispatcher.rb
@@ -21,8 +21,6 @@ module App
     MIN_THREADS = (App::Config.dig(:thread_pool, :min_threads) || 0).to_i
     MAX_THREADS = (App::Config.dig(:thread_pool, :max_threads) || 5).to_i
     MAX_QUEUE = (App::Config.dig(:thread_pool, :max_queue) || 100).to_i
-    MAX_INGESTION_QUEUE_SIZE = (App::Config.max_ingestion_queue_size || Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_SIZE).to_i 
-    MAX_INGESTION_QUEUE_BYTES = (App::Config.max_ingestion_queue_bytes || Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_BYTES).to_i
 
     @running = Concurrent::AtomicBoolean.new(false)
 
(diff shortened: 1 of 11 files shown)


vidok:
I would use named arguments here https://github.com/elastic/connectors-ruby/pull/434/files#diff-ff8974a0d605a7d8f2dc0e25b718febce7372e71dbf57aaf02a9ae1de9e0457dR26 but it's not a blocker. 

LGTM!

vidok (approved):

mchernyavskaya (approved):
LGTM

lib/app/dispatcher.rb
```diff
@@ -133,6 +133,8 @@ def start_consumer!
           min_threads: MIN_THREADS,
           max_threads: MAX_THREADS,
           max_queue: MAX_QUEUE,
+          max_ingestion_queue_size: (App::Config.max_ingestion_queue_size || Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_SIZE).to_i,
+          max_ingestion_queue_bytes: (App::Config.max_ingestion_queue_bytes || Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_BYTES).to_i,
```
wangch079, line 137: I would suggest moving them to the top, together with other constants, e.g. `MAX_QUEUE`.

lib/utility/bulk_queue.rb
```diff
@@ -6,12 +6,14 @@
...
 
+require 'utility/constants'
+
 module Utility
   class BulkQueue
     class QueueOverflowError < StandardError; end
 
     # 500 items or 5MB
-    def initialize(operation_count_threshold = 500, size_threshold = 5 * 1024 * 1024)
+    def initialize(operation_count_threshold = Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_LENGTH, size_threshold = Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_BYTES)
```
wangch079, line 16: Typo, but I update it in my latest PR

lib/core/ingestion/es_sink.rb
```diff
@@ -21,7 +21,7 @@
 module Core
   module Ingestion
     class EsSink
-      def initialize(index_name, request_pipeline, bulk_queue = Utility::BulkQueue.new, max_allowed_document_size = 5 * 1024 * 1024)
+      def initialize(index_name, request_pipeline, bulk_queue = Utility::BulkQueue.new, max_allowed_document_size = Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_BYTES)
```
wangch079, line 24: Not a big deal. Do you think we should hide `BulkQueeue` as the implementation detail of `EsSink`?

```suggestion
      def initialize(index_name, request_pipeline, max_ingestion_queue_size = Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_SIZE, max_allowed_document_size = Utility::Constants::DEFAULT_MAX_INGESTION_QUEUE_BYTES)
        @bulk_queue = Utility::BulkQueue.new(max_ingestion_queue_size, max_ingestion_queue_bytes)
```

wangch079 (commented):
Sorry for the late review

lib/app/dispatcher.rb
artem-shelkovnikov: A fun thing with it - it makes it impossible to mock it properly because the assignment happens when the file is included, thus I cannot have a "natural" test that sets `App::Config.max_ingestion_queue_size` first to nil, then to a value, do you know a way around it?

lib/utility/bulk_queue.rb
artem-shelkovnikov: Oh wow, I'll fix it in a separate PR to not batch it with your other changes

lib/core/ingestion/es_sink.rb
artem-shelkovnikov: Yeah it's possible, maybe it's better the way you described it indeed. I also feel like moving `Utility::BulkQueue` to `Ingestion::BulkQueue` or rearranging the classes somehow better.

lib/app/dispatcher.rb
wangch079: Will `stub_const` work?

lib/core/ingestion/es_sink.rb
wangch079: A future improvement. Not necessarily now as we already passed FF.

lib/app/dispatcher.rb
wangch079: `stub_const` will just mock the constant, so it will ignore whatever happens in `App::Config`. 

By having `App::Config.poll_interval` be nil or non-nil, you are essentially testing the ruby operator `||`.

artem-shelkovnikov: I'm trying to test a logical feature that says "if poll_interval is provided in config, then it's respected and used in scheduler properly, if it's not provided, then a different default value is used".

I don't particularly like it this way here, IMO we need to move the defaulting logic to configuration class and I can remove these tests from here

wangch079: Like this:
```ruby
context 'constant' do
  let(:poll_interval) { 120 }
  before(:each) do
    allow(App::Config).to receive(:poll_interval).and_return(poll_interval)
  end

  it 'uses configured poll interval' do
    expect(described_class.const_get('POLL_INTERVAL')).to eq(poll_interval)
  end
  
  context 'when it\`s not configured' do
    let(:poll_interval) { nil }

    it 'uses the default poll interval' do
      expect(described_class.const_get('POLL_INTERVAL')).to eq(3)
    end
  end
end
```

Again, I don't really think it's necessary.

>if poll_interval is provided in config, then it's respected and used in scheduler properly, if it's not provided, then a different default value is used

IMO, when you test `scheduler`, you shouldn't care whether the `poll_interval` is coming from configured value or default.

artem-shelkovnikov: I've tried similar approach to the one you described in the snippet, and the problem is `POLL_INTERVAL` is assigned once for the whole run-time of the tests in the batch, only when the class `App::Dispatcher` is loaded.

Therefore only one test of ones that you mentioned will work - depending on which one will be executed first OR if another file already loaded `App::Dispatcher` it will depend on a setup of the test that loaded it.

I don't see a way around it without redesigning the module and defaulting logic.

wangch079: > the problem is `POLL_INTERVAL` is assigned once for the whole run-time of the tests in the batch, only when the class `App::Dispatcher` is loaded.

OK, I never tested the snippet above, so I'm not sure if it works or not. My point is we really don't need to test `App::Config.poll_interval || 3`. Not a big deal, we can always refactor later.

PR #435, opened 2022-11-15
Add error monitor logic

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/2271

Adding `ErrorMonitor` class that helps monitor errors and if too many errors happen then raise an error.

Use case:

Connector picks up documents, transforms them to a serialized version for ingestion to Elasticsearch. If something happens during serialization, we currently terminate sync. Error monitor allows to not terminate the sync unless too many errors happen or errors happen too often. This way we can ignore some small errors that happen really rarely and still deliver data to Elasticsearch.

In future same mechanism can be used on Elasticsearch client side to also tolerate some errors when uploading data to Elasticsearch - now it's out of scope of the PR.

#### Pre-Review Checklist

## Release Note
MongoDB connector is able to tolerate rare errors during serialization and still upload documents that do not fail serialization to Elasticsearch.


lib/connectors/mongodb/connector.rb
```diff
@@ -110,7 +110,9 @@ def yield_documents
 
             view = cursor.skip(skip).limit(PAGE_SIZE)
             view.each do |document|
-              yield serialize(document)
+              yield_with_handling_transient_errors do
```
seanstory, line 113: I think this block should encompass the full block of the `each do |document|` so we capture any errors that somehow occur in the below math lines (or other logic that eventually gets added there)

lib/connectors/transient_error_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+require 'utility/logger'
+require 'utility/exception_tracking'
+require 'utility/error_monitor'
+
+module Connectors
+  class TransientErrorHelper
```
seanstory, line 12: Nit: "Transient" may be a misnomer

I probably should have called this out before, but these errors may not be transient. It might be that a specific document is just "bad." That document may _always_ fail, and with the same error. It's just not indicative of a sync-wide problem. "Sufferable" or "Bearable" or "Minor" might be better terms than "Transient".

lib/connectors/transient_error_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+      @error_monitor.note_error(e, :id => e.id)
+    end
+
+    private
+
+    def identifying_error_message(identifier)
+      identifier.present? ? " of '#{identifier}'" : ''
+    end
+
+    def fatal_exception_classes
```
seanstory, line 37: This used to be overridable at the connector level. Now the user will need to subclass the error helper AND override the connector's `initialize` method in order to change it. What if instead, `fatal_exception_classes` was a method in `Base::Connector` that could be used as a parameter for this class?

lib/utility.rb
```diff
@@ -11,6 +11,7 @@
 require 'utility/elasticsearch/index/mappings'
 require 'utility/elasticsearch/index/text_analysis_settings'
 require 'utility/environment'
+require 'utility/error_monitor'
```
seanstory, line 14: this needs to be added to the connectors_utility.gemspec

While you're at it, can you add an ALL CAPS comment to this file that tells editors that any extra additions here need to be added to the gemspec (this ruined my evening last night, so it's top of mind. :) )

lib/utility/error_monitor.rb
```diff
@@ -0,0 +1,107 @@
...
+      @error_queue = []
+    end
+
+    def note_success
+      @consecutive_error_count = 0
+      @success_count += 1
+      increment_window_index
+    end
+
+    def note_error(error, id: Time.now.to_i)
```
seanstory, line 56: - This may need an import for `Time`
- Should we use DateTime instead?

spec/connectors/transient_error_helper_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+
+    context 'when an error happens' do
+      let(:error) { StandardError.new }
+      let(:unique_error_id) { 'hey im an error' }
+
+      before(:each) do
+        allow(Utility::ExceptionTracking).to receive(:augment_exception).with(error) # this method actually populates id of the error
+        allow(error).to receive(:id).and_return(unique_error_id)
```
seanstory, line 41: should this be an `expect` instead of an `allow`?

lib/connectors/base/connector.rb
```diff
@@ -58,6 +62,10 @@ def initialize(configuration: {}, job_description: {})
 
       def yield_documents; end
 
+      def yield_with_handling_transient_errors(identifier: nil, &block)
```
seanstory, line 65: Should this PR introduce using this in the example connector too?

lib/connectors/mongodb/connector.rb
```diff
@@ -110,7 +110,9 @@ def yield_documents
...
+                yield serialize(document)
```
seanstory, line 114: Opinion: It might be good to have a unit/integration for the actual connector.

seanstory (commented):
Nice work. I'd forgotten how dense the Monitor code was, and I am sorry now for whoever had to review it the first time 😅

spec/connectors/transient_error_helper_spec.rb
artem-shelkovnikov: It's a setup, so `allow` :) 

I'm mocking `augment_exception` above and this `augment_exception` adds unique id to this error, so this block mocks the behaviour

lib/connectors/base/connector.rb
artem-shelkovnikov: Good call, I want to introduce a sample connector that uses it in a different PR.

lib/connectors/mongodb/connector.rb
artem-shelkovnikov: Good call, will add!

lib/connectors/mongodb/connector.rb
artem-shelkovnikov: It might not work cause of `break` statement, gonna try it out, but makes sense!

lib/connectors/transient_error_helper.rb
artem-shelkovnikov: Another possible option: Tolerable.

lib/connectors/transient_error_helper.rb
artem-shelkovnikov: We can postpone refactoring cause only Mongo uses it. When we refactor (can do now too) there are 2 options:
1. inline transient_error_helper or whatever it's called into Base::Connector
2. Make users of connector create an instance of transient_error_helper and pass error classes into initializer.
3. Something else I could not think about?

artem-shelkovnikov: From Sean:
3. is a combo 1+2. Connector defines a function "fatal_exception_classes" and base connector calls it when initializing an instance of transient_error_helper.

lib/utility.rb
artem-shelkovnikov: OMG yes

lib/utility/error_monitor.rb
artem-shelkovnikov: yes and yes, thx

spec/connectors/transient_error_helper_spec.rb
artem-shelkovnikov: Addressed in line 53:

```ruby
expect(error_monitor).to receive(:note_error).with(error, { :id => unique_error_id })
```

artem-shelkovnikov:
@seanstory as discussed asking for your help on this PR, plz halp!

lib/utility/error_monitor.rb
seanstory: Since this was copied from Workplace Search, I'll leave it at `Time`, and just add the import statement.

PR #436, opened 2022-11-15
Add generic (and MongoDB specific) and extensible advanced snippet validation support

timgrein:
## Closes (partially) https://github.com/elastic/enterprise-search-team/issues/3167 ##

This PR introduces a general structure for how advanced snippet validation can be implemented. 

Important notes:
* validate_filtering is now implemented by the base connector:
  * The specific connectors only need to plug in a validator class and the framework handles the rest (clear extension point without the need to reimplement validate_filtering all the time)
  * A general base validator class was added, which can be extended to implement validation logic
  * Another base validator class was added (exposes an API to validate against a schema)
* We have to introduce hyphenizing filters/options for the MongoDB ruby client (not included in this PR)
* MongoDB query/filter validation is not included in this PR as this implies the need of parsing a query, which is a very involved task (@seanstory suggested we open an issue to MongoDB

Possible follow-ups:
* Fields constraints should include a custom error message, if the constraint was violated to indicate a clear action to the user
* Custom type matcher functions should also be able to handle custom error messages to give a clear action to the user, if a custom type didn't match
* The schema validator should not only handle top-level Hashes, but also Arrays

#### Pre-Review Checklist
- [x] Tested the changes locally (only had the time for some very basic checks!)

## Release Note

Introduce advanced snippet validation support to the framework


spec/support/shared_examples/filtering.rb
```diff
@@ -0,0 +1,55 @@
...
+    validation_result = subject.is_snippet_valid?
+
+    expect(validation_result[:state]).to eq(Core::Filtering::ValidationStatus::INVALID)
+    expect(validation_result[:errors]).to_not be_empty
+
+    expected_errors ||= []
+
+    unless expected_errors.empty?
+      # TODO: validate errors
+    end
```
seanstory, line 53: doesn't seem to do anything

spec/connectors/base/advanced_snippet_against_schema_validator_spec.rb
```diff
@@ -0,0 +1,556 @@
...
+          }
+        }
+      }
+
+      before(:each) do
+        stub_const('Connectors::Base::AdvancedSnippetAgainstSchemaValidator::MAX_RECURSION_DEPTH', 1)
+      end
+
+      it 'stops logs warning' do
+        expect(Utility::Logger).to receive(:warn)
```
seanstory, line 120: assert message

seanstory (commented):
This is way more comprehensive than I'd initially expected we could have. Awesome work.

seanstory (approved):

PR #440, opened 2022-11-17
Add more tests for fields constraints and unify two method calls

timgrein:
This PR adds some more tests for fields constraints:
- Multiple fields constraints present
- Single constraint wrapped in an array
- Single constraint not wrapped in an array

Also unified the calls to `fields_constraint_violation` and `fields_constraints_violated?`.

#### Pre-Review Checklist

### DIFF
--- a/spec/connectors/base/advanced_snippet_against_schema_validator_spec.rb
+++ b/spec/connectors/base/advanced_snippet_against_schema_validator_spec.rb
@@ -22,7 +22,7 @@ describe Connectors::Base::AdvancedSnippetAgainstSchemaValidator do
   it_behaves_like 'an advanced snippet validator'
 
   describe '#is_snippet_valid?' do
-    context 'fields constraint is present' do
+    context 'fields constraints are present' do
       let(:constraints) {
         [false]
       }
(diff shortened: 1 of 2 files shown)


spec/connectors/base/advanced_snippet_against_schema_validator_spec.rb
```diff
@@ -76,6 +93,41 @@
...
       end
+
+      context 'two fields are allowed and their key must be equal to their value' do
+        let(:constraints) {
+          [
+            ->(fields) { fields.size == 2 },
+            lambda { |fields|
+              fields.each { |field_name, field_value|
+                return false unless field_name.to_s == field_value
+              }
```
artem-shelkovnikov, line 104: Curious, why different lambda syntax here?

timgrein: RuboCop want this syntax for multi-line lambdas

artem-shelkovnikov: cc @vidok something we can improve on too :)

artem-shelkovnikov (approved):

PR #441, opened 2022-11-17
Add tests to registry_spec

timgrein:
Add some missing tests to `registry_spec.rb` to have more coverage.

#### Pre-Review Checklist

### DIFF
--- a/spec/connectors/registry_spec.rb
+++ b/spec/connectors/registry_spec.rb
@@ -72,10 +72,12 @@ describe Connectors::Factory do
   end
 
   describe '#registered_connectors' do
+    let(:registered_connectors) {
+      %w[a-connector b-connector c-connector]
+    }
+
     before(:each) do
-      subject.register('b-connector', MyConnector)
-      subject.register('c-connector', MyConnector)
-      subject.register('a-connector', MyConnector)
+      registered_connectors.each { |connector| subject.register(connector, MyConnector) }
     end
 
     it 'returns registered connectors' do
--- a/spec/connectors/registry_spec.rb
+++ b/spec/connectors/registry_spec.rb
@@ -53,6 +53,7 @@ describe Connectors::Factory do
         connector_instance = subject.connector('my-connector', configuration, job_description)
 
         expect(connector_instance).to_not be_nil
+        expect(connector_instance.is_a?(MyConnector)).to eq(true)
       end
     end
 
--- a/spec/connectors/registry_spec.rb
+++ b/spec/connectors/registry_spec.rb
@@ -38,34 +38,38 @@ describe Connectors::Factory do
   end
 
   describe '#connector_class' do
-    it 'returns registered class' do
-      expect(subject.connector_class(registered_connector)).to eq MyConnector
+    context 'when called against previously registered service type' do
+      it 'returns registered class' do
+        expect(subject.connector_class(registered_connector)).to eq MyConnector
+      end
     end
   end
 
   describe '#registered?' do
-    context 'my-connector is registered' do
+    context 'when called against previously registered service type' do
       it 'should return that my-connector is registered' do
         expect(subject.registered?(registered_connector)).to be_truthy
       end
+    end
 
-      it 'should return that another-connector is not registered' do
+    context 'when called against non-registered service type' do
+      it 'should return that non-registered service type is not registered' do
         expect(subject.registered?(unregistered_connector)).to be_falsey
       end
     end
   end
 
   describe '#connector' do
-    context 'my-connector is registered' do
-      it 'should return a my-connector instance' do
+    context 'when called against previously registered service type' do
+      it 'should return the corresponding connector instance' do
         connector_instance = subject.connector(registered_connector, configuration, job_description)
 
         expect(connector_instance).to be_a(MyConnector)
       end
     end
 
-    context 'another-connector is not registered' do
-      it 'should raise an exception, that another-connector is not registered' do
+    context 'when called against non-registered service type' do
+      it 'should raise an exception, that non-registered connector is not registered' do
         expect { subject.connector(unregistered_connector, configuration, job_description) }.to raise_exception
       end
     end
--- a/spec/connectors/registry_spec.rb
+++ b/spec/connectors/registry_spec.rb
@@ -29,7 +29,7 @@ describe Connectors::Factory do
   end
 
   describe '#connector_class' do
-    it 'let us register a connector under a name' do
+    it 'returns registered class' do
       connector_class = subject.connector_class('my-connector')
       expect(connector_class == (MyConnector)).to eq(true)
     end
--- a/spec/connectors/registry_spec.rb
+++ b/spec/connectors/registry_spec.rb
@@ -30,8 +30,7 @@ describe Connectors::Factory do
 
   describe '#connector_class' do
     it 'returns registered class' do
-      connector_class = subject.connector_class('my-connector')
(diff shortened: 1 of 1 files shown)


spec/connectors/registry_spec.rb
```diff
@@ -9,21 +9,74 @@
...
+
+    it 'returns registered connectors sorted' do
+      registered_connectors = subject.registered_connectors
 
-      factory.register('sharepoint', MyConnector)
-      connector_class = factory.connector_class('sharepoint')
-      expect(connector_class.new.works).to eq 'works'
+      expect(registered_connectors[0]).to eq('a-connector')
+      expect(registered_connectors[1]).to eq('b-connector')
+      expect(registered_connectors[2]).to eq('c-connector')
```
artem-shelkovnikov, line 79: Might be better to just do `expect(registered_connectors).to include('a-connector')` to not make the test flakey if ordering of registration changes.

I'd probably also move `a-connector`, `b-connector` and `c-connector` to variables in test setup.

spec/connectors/registry_spec.rb
```diff
@@ -9,21 +9,74 @@
...
+    context 'my-connector is registered' do
+      it 'should return a my-connector instance' do
+        connector_instance = subject.connector('my-connector', configuration, job_description)
+
+        expect(connector_instance).to_not be_nil
+        expect(connector_instance.is_a?(MyConnector)).to eq(true)
+      end
+    end
+
+    context 'another-connector is not registered' do
```
artem-shelkovnikov, line 60:
```suggestion
    context 'sharepoint connector has not been registered' do
```

Idk `another-connector` looks like a leftover from a different test. Maybe worth to change `sharepoint` to `another-connector` instead.

spec/connectors/registry_spec.rb
```diff
@@ -9,21 +9,74 @@
...
+    end
+  end
+
+  describe '#connector' do
+    context 'my-connector is registered' do
+      it 'should return a my-connector instance' do
+        connector_instance = subject.connector('my-connector', configuration, job_description)
+
+        expect(connector_instance).to_not be_nil
+        expect(connector_instance.is_a?(MyConnector)).to eq(true)
```
artem-shelkovnikov, line 56: Can be probably done in a cleaner manner with type matcher: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/type-matchers

spec/connectors/registry_spec.rb
```diff
@@ -9,21 +9,74 @@
...
+  end
+
+  describe '#registered?' do
+    context 'my-connector is registered' do
+      it 'should return that sharepoint connector is registered' do
+        expect(subject.registered?('my-connector')).to eq(true)
+      end
+
+      it 'should return that another-connector is not registered' do
+        expect(subject.registered?('another-connector')).to eq(false)
```
artem-shelkovnikov, line 45: I would probably move `another-connector` to a variable `unregistered_connector_type` or something similar and reuse in all test, same with `let(:registered_type) { 'my-connector' }`

artem-shelkovnikov (commented):
Nice! Left a couple comments, but love to see more test coverage!

spec/connectors/registry_spec.rb
```diff
@@ -9,21 +9,74 @@
...
+      # nothing needed
+    end
+  end
+
+  before(:each) do
+    subject.register('my-connector', MyConnector)
   end
 
   describe '#connector_class' do
     it 'let us register a connector under a name' do
```
wangch079, line 32:
```suggestion
    it 'returns registered class' do
```

spec/connectors/registry_spec.rb
```diff
@@ -9,21 +9,74 @@
...
-      class MyConnector
-        def works
-          'works'
-        end
+      connector_class = subject.connector_class('my-connector')
+      expect(connector_class == (MyConnector)).to eq(true)
```
wangch079, line 34:
```suggestion
      expect(subject.connector_class('my-connector')).to eq MyConnector
```

spec/connectors/registry_spec.rb
```diff
@@ -9,21 +9,74 @@
...
+    end
+  end
+
+  describe '#registered?' do
+    context 'my-connector is registered' do
+      it 'should return that sharepoint connector is registered' do
+        expect(subject.registered?('my-connector')).to eq(true)
+      end
```
wangch079, line 42:
```suggestion
      it 'should return that my-connector is registered' do
        expect(subject.registered?('my-connector')).to be_truty
      end
```

timgrein: it's probably `be_truthy` right?

timgrein: changed `be_truty` to `be_truthy`

spec/connectors/registry_spec.rb
timgrein: Ah nice, really have to look more into the built-in-matchers 👍

spec/connectors/registry_spec.rb
timgrein: The message is correct, but I've fixed it by using the `unregistered_connector` field now

spec/connectors/registry_spec.rb
timgrein: 👍

spec/connectors/registry_spec.rb
timgrein: done 👍

spec/connectors/registry_spec.rb
```diff
@@ -9,21 +9,79 @@
...
+
+    before(:each) do
+      registered_connectors.each { |connector| subject.register(connector, MyConnector) }
+    end
 
-      factory.register('sharepoint', MyConnector)
-      connector_class = factory.connector_class('sharepoint')
-      expect(connector_class.new.works).to eq 'works'
+    it 'returns registered connectors' do
+      expect(subject.registered_connectors).to include('a-connector', 'b-connector', 'c-connector')
```
artem-shelkovnikov, line 84:
```suggestion
      expect(subject.registered_connectors).to include(*registered_connectors)
```

I think array splat operator usage here is a good way to make test simpler and robust :)

spec/connectors/registry_spec.rb
```diff
@@ -9,21 +9,79 @@
...
+  }
+
+  let(:unregistered_connector) {
+    'another-connector'
+  }
+
+  class MyConnector
+    def initialize(configuration: {}, job_description: {})
+      # nothing needed
+    end
```
artem-shelkovnikov, line 33:
```suggestion
    def initialize(configuration: {}, job_description: {}); end
```

This can be further simplified with empty method syntax - up to you to decide whether you want to use it here or not

spec/connectors/registry_spec.rb
```diff
@@ -9,21 +9,79 @@
...
-        def works
-          'works'
-        end
+    it 'returns registered class' do
+      expect(subject.connector_class(registered_connector)).to eq MyConnector
+    end
+  end
+
+  describe '#registered?' do
+    context 'my-connector is registered' do
```
artem-shelkovnikov, line 47: I would probably rename this style of sections throughout the file to something like:

```
context 'when called against previously registered service type'
```

It's also a good practice to start statements inside `context` with "when": "when connector is registered", "when variable is nil", etc.

artem-shelkovnikov (approved):
Looks good, left a couple more small comments to improve on the PR even further :)

PR #442, opened 2022-11-18
Fix a problem with error monitor and max_error_ratio being incorrectly handled

artem-shelkovnikov:
Found a problem with error monitor class: instead of raising an error when number of errors in the window exceeds max_error_ratio, currently it just raises an error when number of errors that happened is equal to `window_size * max_error_ratio`.

Bug happened because the window of errors never considered successful documents as a part of equation - successful document was not reported into the window ever.

#### Pre-Review Checklist

### DIFF
--- a/lib/utility/error_monitor.rb
+++ b/lib/utility/error_monitor.rb
@@ -93,7 +93,7 @@ module Utility
     end
 
     def num_errors_in_window
-      @window_errors.count { |e| e == true } .to_f
+      @window_errors.count(true).to_f
     end
 
     def increment_window_index
--- a/lib/utility/error_monitor.rb
+++ b/lib/utility/error_monitor.rb
@@ -51,8 +51,7 @@ module Utility
     def note_success
       @consecutive_error_count = 0
       @success_count += 1
-      @window_errors[@window_index] = false
-      increment_window_index
+      track_window_error(false)
     end
 
     def note_error(error, id: Time.now.to_i)
@@ -61,10 +60,9 @@ module Utility
       Utility::Logger.debug("Message id: #{id} - #{error_message}\n#{stack_trace}")
       @total_error_count += 1
       @consecutive_error_count += 1
-      @window_errors[@window_index] = true
       @error_queue << DocumentError.new(error.class.name, error_message, stack_trace, id)
       @error_queue = @error_queue.drop(1) if @error_queue.size > @error_queue_size
-      increment_window_index
+      track_window_error(true)
       @last_error = error
 
       raise_if_necessary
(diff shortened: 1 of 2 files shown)


lib/utility/error_monitor.rb
```diff
@@ -92,10 +93,31 @@ def raise_if_necessary
     end
 
     def num_errors_in_window
-      @window_errors.count(&:itself).to_f
+      @window_errors.count { |e| e == true } .to_f
```
wangch079, line 96:
```suggestion
      @window_errors.count(true).to_f
```

lib/utility/error_monitor.rb
```diff
@@ -51,6 +51,7 @@ def initialize(
     def note_success
       @consecutive_error_count = 0
       @success_count += 1
+      @window_errors[@window_index] = false
```
wangch079, line 54: It's better to encapsulate set window error and increment window index into one method.
```ruby
def set_window_error(error)
  @window_errors[@window_index % @window_size] = error
  @window_index = (@window_index + 1) % @window_size
end
```

lib/utility/error_monitor.rb
```diff
@@ -92,10 +91,32 @@ def raise_if_necessary
...
+      # [ true, false, true,   false, true  ]
+      #   ^^^^
+      #   5 % 5 == 0
+      #
+      # Eigth document is successful:
+      # [ true, false, false,  false, true  ]
+      #                ^^^^^
+      #                7 % 5 == 2
+      # And so on.
+      @window_errors[@window_index] = is_error
```
wangch079, line 119: I will be more defensive here, but not a blocker.

wangch079 (approved):
LGTM

lib/utility/error_monitor.rb
artem-shelkovnikov: I copied the code from Enterprise Search and tried to leave it as close to original as possible, but in the end I'd rewrite this class fully - it's a bit sloppy and as research showed it's prone to bugs.

seanstory: oof, that's a little hurtful. I was quite proud of this implementation. Of course, didn't catch this bug of overwriting past errors in the window to successes 🙈 . Nice fix, and great comment to explain the logic.

artem-shelkovnikov: Sorry @seanstory, when writing my comment I did not think that it could be hurtful, but I see it now.

I had some small problems when reading the code when the bug happened and figuring out the logic and it made me think that some parts of the code can be encapsulated better - particularly the rate calculating logic - but the code works in every known scenario now, so it's a working component that we can use. Can it be improved? Of course, any piece of code can :)

PR #467, opened 2022-11-24
Feature: Add simple rule validation support and implement basic generic validation

timgrein:
## Related to https://github.com/elastic/enterprise-search-team/issues/3174 ##

This change introduces simple rule validation support to the framework (will be backported to 8.6).

Connector implementations can plug-in dedicated validation classes for different stages (pre-processing, post-processing or for all stages). The framework will take care of the rest (instantiating the correct classes, passing the advanced snippet or the simple rules and checking, if pre-processing simple rules are active)

The following validations are implemented:
  - Every rule is validated against a rule schema to make sure the correct fields and values are present (https://github.com/elastic/enterprise-search-team/issues/3174)
  - Conflicting rules with regards to their policy (INCLUDE and EXCLUDE for the exact same fields, values, rules are not allowed) (https://github.com/elastic/enterprise-search-team/issues/3375)
  - Match all regex is not allowed (https://github.com/elastic/enterprise-search-team/issues/3372)
  - Empty fields and values are not allowed (https://github.com/elastic/enterprise-search-team/issues/3373)

I've also moved everything into a dedicated "filtering" directory so we can isolate changes in the future and also throw away stuff, if we decide to take another road for filtering (and its validation).

#### Pre-Review Checklist


lib/core/filtering/hash_against_schema_validator.rb
```diff
@@ -7,53 +7,43 @@
...
-
-      private
-
-      def validate_against_schema(config_schema, advanced_snippet, recursion_depth = 0)
+      def validate_against_schema(schema = @schema, payload = @payload, recursion_depth = 0)
         # Prevent unintentional/intentional SystemStackErrors/crashes
         return unexpected_error if exceeded_recursion_depth?(recursion_depth)
 
-        return valid_snippet if config_schema.nil? || config_schema.empty?
+        return valid_snippet if schema.nil? || schema.empty?
```
seanstory, line 27:
```suggestion
        return valid_snippet unless schema.present?
```

lib/core/filtering/simple_rules/validation/no_conflicting_policies_rules_validator.rb
```diff
@@ -0,0 +1,47 @@
...
+        class NoConflictingPoliciesRulesValidator < Core::Filtering::SimpleRules::Validation::SimpleRulesValidator
+
+          def are_rules_valid
+            rule_field_value_to_policy = {}
+
+            @rules.each do |simple_rule|
+              rule_field_value = simple_rule.slice('rule', 'field', 'value')
+              policy = simple_rule['policy']
+
+              return conflicting_rules(rule_field_value) if rule_field_value_to_policy.key?(rule_field_value) && rule_field_value_to_policy[rule_field_value]['policy'] != policy
```
seanstory, line 24:
```suggestion
              return conflicting_rules(rule_field_value) if rule_field_value_to_policy.key?(rule_field_value)
```

Should we ever have the same `rule_field_value`, even if the policy is the same? I think the combo of those 3 should be unique, right?

lib/core/filtering/simple_rules/validation/no_conflicting_policies_rules_validator.rb
```diff
@@ -0,0 +1,47 @@
...
+            { :state => Core::Filtering::ValidationStatus::VALID, :errors => [] }
+          end
+
+          private
+
+          def conflicting_rules(rule_type_value)
+            {
+              :state => Core::Filtering::ValidationStatus::INVALID,
+              :errors => [
+                :ids => [SIMPLE_RULES_ID],
```
seanstory, line 38: This should actually be the IDs of the two rules that conflicted. Like if you had:
```
{
  {
    "id": "foo",
    "policy":"include",
    "field":"bar",
    "rule":"equals",
    "value":"baz",
    "order":1
  },
  {
    "id": "qux",
    "policy":"exclude",
    "field":"bar",
    "rule":"equals",
    "value":"baz",
    "order":2
  }
]
```

you should get `:ids => ['foo', 'qux']`

seanstory: @sphilipse I believe you'd mentioned that Kibana cannot handle these errors in 8.6, so changing this behavior is probably not a high priority right now. But maybe worth filing an issue to come back to this in early in 8.7.

lib/core/filtering/simple_rules/validation/simple_rules_schema.rb
```diff
@@ -0,0 +1,66 @@
...
+# frozen_string_literal: true
+
+require 'core/filtering/simple_rules/simple_rule'
+
+module Core
+  module Filtering
+    module SimpleRules
+      module Validation
+        ALLOWED_VALUE_TYPES = ->(rule_value) { rule_value.is_a?(String) || rule_value.is_a?(Integer) || rule_value.is_a?(TrueClass) || rule_value.is_a?(FalseClass) }
+        MATCH_ALL_REGEX_NOT_ALLOWED = ->(simple_rule) { !(simple_rule['rule'] == Core::Filtering::SimpleRule::Rule::REGEX && (simple_rule['value'] == '(.*)' || simple_rule['value'] == '.*')) }
```
seanstory, line 15: will this complain about the default rule?

spec/core/filtering/advanced_snippet/advanced_snippet_validator_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+# frozen_string_literal: true
+
+require 'core/filtering/advanced_snippet/advanced_snippet_validator'
+
+describe Core::Filtering::AdvancedSnippet::AdvancedSnippetValidator do
+  subject { described_class.new({}) }
+
+  describe '#is_snippet_valid' do
+    it 'should raise an exception' do
+      expect { subject.is_snippet_valid }.to raise_exception
```
seanstory, line 15: Should this check the exception class and/or message?

spec/connectors/mongodb/mongo_advanced_snippet_against_schema_validator_spec.rb
```diff
@@ -16,7 +16,28 @@
...
+    it '' do
+      validation_result = subject.is_snippet_valid
+
+      expect(validation_result[:state]).to eq(Core::Filtering::ValidationStatus::INVALID)
+      expect(validation_result[:errors]).to_not be_empty
+      expect(validation_result[:errors]).to be_an(Array)
+      expect(validation_result[:errors][0][:ids]).to eq(['advanced_snippet'])
+      expect(validation_result[:errors][0][:messages]).to be_an(Array)
+    end
+  end
```
seanstory, line 38: should these be in a shared_examples file? Looks like they're used outside of just this spec

seanstory (approved):
several nits, but largely LGTM. Anything you don't want to fix now, let's just make a note of to come back to later.

Great work and design!

tarekziade (approved):
LGTM from an architectural PoV -- we had a design review with Tim

My key asks for this feature were 
- to make sure the filtering code is decoupled from the rest
- to fallback to no filtering in case of an issue at filtering time or validation time
- to have explicit logging when the filtering validation or execution happen, so we can trace back what happened

DianaJourdan:
@timgrein @tarekziade are the key asks smth we want to do before merging this PR?

lib/core/filtering/simple_rules/validation/no_conflicting_policies_rules_validator.rb
timgrein: Good point, will create an issue for that

lib/core/filtering/simple_rules/validation/simple_rules_schema.rb
timgrein: It didn't during testing. I guess the default rule is not sent for validation? Is the ID of the default rule always the same? If yes I could add a check for that.

timgrein:
@DianaJourdan 1. and 3. is already addressed. 2. also in some way, but we probably have to reiterate to know exactly how we handle issues during filtering/validation time in a unified way.

spec/connectors/mongodb/mongo_advanced_snippet_against_schema_validator_spec.rb
timgrein: Double checked, they're not used outside of this spec. There is something similar for checking if a whole filter is valid, maybe you mean that?

spec/core/filtering/advanced_snippet/advanced_snippet_validator_spec.rb
timgrein: Addressed 👍 (checks now the message)

PR #469, opened 2022-11-28
Fix: Sync rules not showing up for native connectors

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3386 ##

This PR fixes the bug that sync rules do not appear for native connectors. This happens, because the native connectors are created in status `NEEDS_CONFIGURATION`, which didn't trigger the configuration task (which adds the features to the corresponding connector entry in .elastic-connectors). Therefore I've moved the logic for checking if a connector needs configuration to the connector settings so we can reuse it.

I've also changed the logic, which handles when to set a connector to `CONFIGURED`. This only happened, when all configurable fields had defaults set, but this should also happen if all fields were set by the user. 

#### Pre-Review Checklist

### DIFF
--- a/lib/core/configuration.rb
+++ b/lib/core/configuration.rb
@@ -17,33 +17,55 @@ module Core
     class << self
 
       def update(connector_settings, service_type = nil)
-        if connector_settings.connector_status == Connectors::ConnectorStatus::CREATED
+        if connector_settings.needs_configuration?
           connector_class = Connectors::REGISTRY.connector_class(connector_settings.service_type || service_type)
+
           unless connector_class
             Utility::Logger.error("Couldn't find connector for service type #{connector_settings.service_type || service_type}")
             return
           end
-          configuration = connector_class.configurable_fields_indifferent_access
+
           features = connector_class.kibana_features.each_with_object({}) { |feature, hsh| hsh[feature] = true }
+          configurable_fields = connector_class.configurable_fields_indifferent_access
           doc = {
-            :configuration => configuration,
             :features => features
           }
 
           doc[:service_type] = service_type if service_type && connector_settings.needs_service_type?
 
-          # We want to set connector to CONFIGURED status if all configurable fields have default values
-          new_connector_status = if configuration.values.all? { |setting| setting[:value].present? }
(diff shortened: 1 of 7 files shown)


timgrein:
FYI: This still leads to a weird UX, where the "Sync rules" tab appears after 3-5 seconds after connector creation (depending how fast the first configuration task is scheduled and executed). I think it's fine for 8.6 to have it like this (better than the alternative of adding the features via the dev console), but we should probably reiterate how we handle those feature flags to prevent these lags in the UI.

mchernyavskaya (approved):
LGTM

mchernyavskaya:
@timgrein this still requires making correspondent changes to `ent-search` right? Otherwise, native connectors won't benefit from the fix.
The process of updating gems is described [here](https://docs.google.com/document/d/10KJOIhe4sauDul8iWeV9Cn-_3uPWa76qG8SwYk6BCAA/edit#bookmark=id.feh6z04d6cjq).

artem-shelkovnikov:
@mchernyavskaya I also remember @tarekziade mentioning that this process was automated, maybe it'll work now without additional work needed?

lib/core/configuration.rb
```diff
@@ -17,34 +17,56 @@ class Configuration
     class << self
 
       def update(connector_settings, service_type = nil)
-        if connector_settings.connector_status == Connectors::ConnectorStatus::CREATED
+        if connector_settings.needs_configuration?
```
seanstory, line 20: I dont' think this lines up with @wangch079 's [suggestion for how to address this](https://github.com/elastic/connectors-ruby/pull/408#discussion_r1017799108). See also [this slack thread](https://elastic.slack.com/archives/C01795T48LQ/p1669655686117569?thread_ts=1668763646.497289&cid=C01795T48LQ) with @sphilipse  and I.

I'm apprehensive of this change, because it changes the protocol by changing which states are the responsibility of the Connector Service to handle. Similar changes would also need to be made to the Python framework. I don't think we want to do that right now without a decent amount of visibility from from the whole team, and opportunity to comment.

PR #473, opened 2022-11-28
Fix: MongoDB aggregation pipelines not working

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3428 ##

This PR introduces a quick fix to allow the usage of aggregation pipelines in 8.6. We didn't have the time to check, whether we run into the same memory leak problems as for find. We have to probably reiterate on that in 8.7. Also we should use a common abstraction for the document yielding (at least for MongoDB) to cleanup yield_documents.

#### Pre-Review Checklist

### DIFF
--- a/connectors_service.gemspec
+++ b/connectors_service.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
     if dep.latest_version?
       s.add_dependency dep.name
     else
-      s.add_dependency dep.name, dep.requirement
+      s.add_dependency dep.name, dep.requirement.as_list
     end
   end
 end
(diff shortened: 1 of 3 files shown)


artem-shelkovnikov (approved):
Being mindful of time left till release of 8.6 I put my approval on this PR.

It's not perfect but makes aggregation pipelines work, we'll need to clean up the code later.

Comment on the logic:
1. It's great to have a comment in `when AGGREGATE` that explains the problem that's solved here
2. Add a test for this use case that just checks that AGGREGATE ignores skip/take

None of these comments is blocking, feel free to merge it when you feel like it

PR #476, opened 2022-11-28
Fix: snake_case MongoDB filtering

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3441 ##

This PR introduces filtering transformation capabilities to the framework. It also implements snake casing for MongoDB.

#### Pre-Review Checklist

### DIFF
--- a/lib/connectors/mongodb/mongo_filter_snake_case_transformer.rb
+++ b/lib/connectors/mongodb/mongo_filter_snake_case_transformer.rb
@@ -1,0 +1,46 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+require 'core/filtering/transform/filter_transformer'
+require 'active_support'
+
+module Connectors
+  module MongoDB
+    class MongoFilterSnakeCaseTransformer < Core::Filtering::Transform::FilterTransformer
+
+      def initialize(filter = {}, transformation = ->(f) { snake_case_filter(f) })
+        super
+      end
+
+      private
+
+      def snake_case_filter(filter, transformed_filter = {})
+        filter.each do |key, value|
+          snake_case_key = key.underscore
+
+          value = value.is_a?(Hash) ? snake_case_filter(value, {}) : value
+
+          if value.is_a?(Array)
+            new_entries = []
+
+            value.each do |entry|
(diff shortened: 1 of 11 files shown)


lib/core/filtering/transform/filter_transformer.rb
```diff
@@ -0,0 +1,26 @@
...
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+#
+
+# frozen_string_literal: true
+
+module Core
+  module Filtering
+    module Transform
+      class FilterTransformer
```
seanstory, line 12: This transformer pattern feels a little overkill to me. Feels like it would be a lot less work to just change the casing in https://github.com/elastic/connectors-ruby/blob/main/lib/connectors/mongodb/mongo_advanced_snippet_schema.rb#L13-L18

Just want to triple-check that @danajuratoni feels that its very  important to have the users use `camelCase` options as documented in the core mongo docs ([example](https://www.mongodb.com/docs/manual/reference/method/cursor.allowDiskUse/)) vs using `snake_case` options as documented in the ruby mongo client ([example](https://www.mongodb.com/docs/ruby-driver/current/api/Mongo/Collection.html#aggregate-instance_method))?

I would have thought it would be better to couple ourselves to what the ruby gem syntax is, since we can only support whatever is supported through the ruby gem anyway. That way we can say that the Advanced snippet could be copy-pasted directly into an `irb` session for the user to troubleshoot their own query.

seanstory (approved):
The code looks good (one small comment). The strategy feels questionable to me, but as long as @danajuratoni feels this behavior is required, I think this is a reasonable approach.

lib/core/filtering/transform/filter_transformer.rb
danajuratoni: We should follow the official MongoDB documentation and should not tie ourselves to a technical implementation. A customer using native MongoDB connectors should not care if they're implemented in ruby, nor should their advanced rules be impacted if say, tomorrow we decide to promote a Java MongoDB connector to replace the ruby connector as native.

lib/connectors/mongodb/mongo_filter_snake_case_transformer.rb
```diff
@@ -0,0 +1,46 @@
...
+# frozen_string_literal: true
+
+require 'core/filtering/transform/filter_transformer'
+require 'active_support'
+
+module Connectors
+  module MongoDB
+    class MongoFilterSnakeCaseTransformer < Core::Filtering::Transform::FilterTransformer
+
+      def initialize(filter = {}, transformation = ->(f) { snake_case_filter(f) })
```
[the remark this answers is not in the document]
timgrein: Good point, done 👍

PR #480, opened 2022-11-29
bug: pre-procesing simple rules turns int into str #3348

mchernyavskaya:
## Closes https://github.com/elastic/enterprise-search-team/issues/3348

The reason for this bug was, we don't set value type in simple rules, so it's always saved as a string. But in comparisons, string doesn't work against numerics. So if we run a query like this, it won't find anything:

```
db.listingsAndReviews.countDocuments({"bedrooms": {$gte: "2"}})
0
```

This, however, gives a desired result:

```
db.listingsAndReviews.countDocuments({"bedrooms": {$gte: 2.0}})
```

So, the code is fixed to match.

#### Pre-Review Checklist


lib/connectors/mongodb/mongo_rules_parser.rb
```diff
@@ -59,18 +59,20 @@ def parse_equals(rule)
       end
 
       def parse_greater_than(rule)
+        coerced_value = try_parse_float(rule.value)
```
DianaJourdan, line 62: Isn't there a chance to use gt and lt operators with other values than float? What about date type or using the numeric value as a string?

seanstory: ++. We should try integers, dates, and booleans.

mchernyavskaya: I opted for float because if the value in the document is an int, it's going to work as well:

```
db.listingsAndReviews.countDocuments({"bedrooms": {$gte: 2.0}})
```

Dates are unfortunately more difficult, since there's so many possible formats. It's difficult to say for sure that the value should be converted to a date. Check this out: 

```
irb(main):020:0> Date.parse('20')
=> #<Date: 2022-11-20 ((2459904j,0s,0n),+0s,2299161j)>
```

The number 20 is parsed as a valid date. But that's not all.

```
=> #<Date: 2022-11-20 ((2459904j,0s,0n),+0s,2299161j)>
irb(main):021:0> Date.parse('5')
       10: from /Users/marynacherniavska/.rbenv/versions/jruby-9.3.3.0/bin/irb:23:in `<main>'
        9: from org/jruby/RubyKernel.java:1052:in `load'
        8: from /Users/marynacherniavska/.rbenv/versions/jruby-9.3.3.0/lib/ruby/gems/shared/gems/irb-1.0.0/exe/irb:11:in `<main>'
        6: from org/jruby/RubyKernel.java:1237:in `catch'
        5: from org/jruby/RubyKernel.java:1507:in `loop'
        4: from org/jruby/RubyKernel.java:1091:in `eval'
        3: from (irb):21:in `evaluate'
        2: from /Users/marynacherniavska/.rbenv/versions/jruby-9.3.3.0/lib/ruby/stdlib/date.rb:488:in `parse'
        1: from /Users/marynacherniavska/.rbenv/versions/jruby-9.3.3.0/lib/ruby/stdlib/date.rb:446:in `new_by_frags'
ArgumentError (invalid date)
```

And this is also not all.

```
irb(main):023:0> Date.parse('05')
=> #<Date: 2022-11-05 ((2459889j,0s,0n),+0s,2299161j)>
```

I am not sure I want to introduce this complexity right now. I think we should punt on dates, or maybe we want to introduce the conversion for selected formats only, but then, which formats? We need to define them, which blows the scope.

I'd prefer to fix at least the simple case now.

lib/core/filtering/simple_rules/simple_rule.rb
```diff
@@ -138,6 +138,18 @@ def to_h
         @rule_hash
       end
 
+      def try_coerce_value
+        coerced = to_float(value)
+        begin
+          coerced = to_date(value) if coerced.is_a?(String)
+        rescue ArgumentError
+          coerced = to_bool(value) if coerced.is_a?(String)
```
seanstory, line 146: I don't think we could get here if `value = "false"`

seanstory: hmm, though you have a test asserting this explicitly, so I must be wrong?

seanstory (approved):
My brain isn't following, but the tests look good. 🤷

lib/core/filtering/simple_rules/simple_rule.rb
mchernyavskaya: Yes, it works for `false`.
1. `to_float` returns a string as it's not float but the implementation doesn't throw
2. `to_date` throws from the `rescue` clause because of the int conversion
3. `to_bool` is called on row 146 and works as expected `'false' => false`

mchernyavskaya:
Since our latest fixes didn't go to BC4 anyway, I think this one can go into BC5 with the rest. WDYT @DianaJourdan @danajuratoni ?

PR #483, opened 2022-11-29
Specifying the backport version and giving directions

seanstory:
Looks like manual backporting wasn't well defined for connectors-ruby. Copied setup from ent-search.

See: https://elastic.slack.com/archives/C03GAPTQ2JF/p1669730253080489


timgrein (approved):
LGTM. Hopefully this works now :-)

PR #508, opened 2022-12-28
Update CONNECTOR_PROTOCOL.md

artem-shelkovnikov:
Found when working on connectors-python repository.

There's expectation that connector id is stored within Jobs index inside `connector.id`, but protocol says it should be `connector.connector_id` - this seems incorrect.

Plus' moved `id` on top of `connector` object - good to have ids as first fields of objects.


sphilipse (approved):
nice catch!

PR #511, opened 2023-01-03
[Filtering] Respect filtering feature flags

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3658 ##

This PR introduces the capability to the framework to not schedule filtering validation tasks if all filtering features are disabled. It also skips this validation and filtering with basic rules in a sync job if all filtering features are disabled. If filtering with basic rules is disabled but filtering with the advanced rules is enabled it stills executes the validation in the sync job without applying basic rule filtering.

#### Pre-Review Checklist


lib/core/connector_settings.rb
```diff
@@ -64,6 +64,22 @@ def [](property_name)
       @elasticsearch_response[:_source][property_name]
     end
 
+    def features
+      self[:features]
```
wangch079, line 68:
```suggestion
      self[:features] || {}
```

lib/core/connector_settings.rb
```diff
@@ -64,6 +64,22 @@ def [](property_name)
...
+    end
+
+    # being defensive here to avoid regressions, will be subject to change in the near future
+    def filtering_rule_feature_enabled?
+      features && !features.empty? && features[:filtering_rules]
```
wangch079, line 73:
```suggestion
      features[:filtering_rules]
```

or 

```suggestion
      !!features[:filtering_rules]
```

lib/core/connector_settings.rb
```diff
@@ -64,6 +64,22 @@ def [](property_name)
...
+    end
+    def filtering_advanced_config_feature_enabled?
+      features && !features.empty? && features[:filtering_advanced_config]
```
wangch079, line 76:
```suggestion
      features[:filtering_advanced_config]
```

or 

```suggestion
      !!features[:filtering_advanced_config]
```

lib/core/scheduler.rb
```diff
@@ -148,6 +148,21 @@ def configuration_triggered?(connector_settings)
     end
 
     def filtering_validation_triggered?(connector_settings)
+      features = connector_settings.features
+
+      if features.nil? || features.empty?
+        # Regression bug
+        Utility::Logger.error("#{connector_settings.formatted} features are empty. Skip filtering validation.")
```
wangch079, line 155: This seems not necessary unless you want to log the error.

What does it imply if it's empty? A migration failed?

lib/core/sync_job_runner.rb
```diff
@@ -94,19 +90,31 @@ def do_sync!
...
-          post_process_result = post_processing_engine.process(document)
-          if post_process_result.is_include?
-            @sink.ingest(document)
-            incoming_ids << document['id']
+        # We want to validate advanced filtering rules even if basic rules are disabled
+        if @connector_settings.any_filtering_feature_enabled?
+          Utility::Logger.info("Checking active filtering for sync job #{@job_id} for connector #{@connector_id}.")
+          validate_filtering(@job.filtering)
+          Utility::Logger.debug("Active filtering for sync job #{@job_id} for connector #{@connector_id} is valid.")
+        end
```
wangch079, line 98: Why move this block here? If there is any invalid rule, `validate_filtering` will raise an exception and stop the sync? So it makes sense to execute this before `ElasticConnectorActions.fetch_document_ids`?

lib/core/sync_job_runner.rb
```diff
@@ -94,19 +90,31 @@ def do_sync!
...
           end
+        else
+          Utility::Logger.info('Yielding documents without basic rule filtering...')
 
-          periodically do
-            check_job
-            @job.update_metadata(@sink.ingestion_stats, @connector_instance.metadata)
+          yield_docs do |document|
+            @sink.ingest(document)
+            incoming_ids << document['id']
```
wangch079, line 117: Github doesn't support suggestions on deleted lines. I just feel this `if..else` contains duplicate code. I think we can have something like

```ruby
post_processing_engine = @connector_settings.filtering_rule_feature_enabled? ? Core::Filtering::PostProcessEngine.new(@job.filtering) : nil

yield_docs do |document|
   continue if post_processing_engine && !post_processing_engine.process(document).is_include?
   @sink.ingest(document)
   incoming_ids << document['id']
 end

```

lib/core/scheduler.rb
timgrein: For example a migration failed or someone tempered with the features manually

lib/core/sync_job_runner.rb
timgrein: Good point! That was unnecessary to move it down.

lib/core/sync_job_runner.rb
timgrein: Looks a lot cleaner, adjusted it!

lib/core/connector_settings.rb
timgrein: Both suggestions break the tests when features were tempered with. That was the reason why I was so defensive here and checked for multiple cases.

lib/core/scheduler.rb
timgrein: Maybe we could make the message more explicit like: "... features are empty. This is likely to be an index related error. Please check your index for connector id '...'.)" I just think it would be good to be explicit that this is an error and something which shouldn't happen (as it already happened before).

timgrein: But not a super strong opinion, I can also take it out, if you would prefer it.

wangch079: I do want to take it out. My point is, we should make our code robust, i.e. it won't crash because some of the system indices (`.elastic-connectors`, `.elastic-connector-sync-jobs`) are tempered. 

But I think it's overkill if we check the existence and validity of each field of system indices before use.

timgrein: With the current implementation we won't crash as they'll just be empty if the features are null in the index, right? 

```
def features
      self[:features] || {}
end
```

But if you think it's overkill to check the existence/validity you would advocate for taking out the explicit guarding/logging, right? Sorry, if I'm not understanding fully what's the action to take 😄

lib/core/scheduler.rb
```diff
@@ -148,6 +148,21 @@ def configuration_triggered?(connector_settings)
...
+
+        return false
+      end
```
wangch079, line 158:
```suggestion
```

My suggestion is to remove this whole block.

wangch079 (approved):

PR #513, opened 2023-01-10
Add custom scheduling trigger

navarone-feekery:
#### Pre-Review Checklist


lib/connectors/crawler/scheduler.rb
```diff
@@ -22,11 +22,50 @@ def connector_settings
...
+          connector_settings.each do |cs|
+            # crawler only supports :sync
+            if sync_triggered?(cs)
+              yield cs, :sync, nil
+              next
+            end
+
+            custom_schedule_triggered, schedule_key = custom_schedule_triggered?(cs)
+            if custom_schedule_triggered
+              yield cs, :sync, schedule_key
```
navarone-feekery, line 36: I also considered the following, but I wasn't sure if it was wise to change the structure of the connector settings:

```ruby
yield cs.merge(scheduling_key: scheduling_key), :sync
```

lib/core/scheduler.rb
```diff
@@ -83,47 +80,7 @@ def sync_triggered?(connector_settings)
...
-      end
-
-      next_trigger_time = cron_parser.next_time(Time.parse(last_synced))
-
-      # Sync if next trigger for the connector is in past
-      if next_trigger_time < Time.now
-        Utility::Logger.info("#{connector_settings.formatted.capitalize} sync is triggered by cron schedule #{current_schedule}.")
+      if schedule_triggered?(connector_settings.scheduling_settings, connector_settings.formatted, connector_settings.last_synced)
         return true
       end
```
timgrein, line 86: I think you can directly return schedule_triggered?, right?

```suggestion
     schedule_triggered?(connector_settings.scheduling_settings, connector_settings.formatted, connector_settings.last_synced)
```

lib/core/scheduler.rb
```diff
@@ -189,5 +146,67 @@ def connector_registered?(service_type)
         false
```
timgrein, line 146:
```suggestion
```

spec/connectors/crawler/crawler_scheduler_spec.rb
```diff
@@ -29,4 +41,123 @@
...
+      let(:scheduling_settings) do
+        {
+          :enabled => sync_enabled,
+          :interval => sync_interval
+        }
+      end
+
+      let(:weekly_enabled) { false }
+      let(:weekly_interval) { '0 0 * * 1 ?' }
+      let(:weekly_last_synced) { 1.week.ago.beginning_of_day.to_s }
```
timgrein, line 71: That looks neat  🎉

timgrein (commented):
Just a small comment :)

lib/connectors/crawler/scheduler.rb
```diff
@@ -22,11 +22,48 @@ def connector_settings
...
+      def when_triggered
+        loop do
+          connector_settings.each do |cs|
+            # crawler only supports :sync
+            if sync_triggered?(cs)
+              yield cs, :sync, nil
+              next
+            end
+
+            custom_schedule_triggered, schedule_key = custom_schedule_triggered?(cs)
```
vidok, line 34: It looks like `custom_schedule_triggered` is not needed here. I would simplify it a bit

```Ruby
schedule_key = triggered_schedule_key(cs)
yield cs, :sync, schedule_key if schedule_key
```

Also, methods ending with `?` should return only boolean value they are called `boolean methods`.

lib/connectors/crawler/scheduler.rb
vidok: I can't see the comment you are replying to but I totally agree that it's wise **not to change** the structure of the connector settings. We don't have a concept of "virtual" fields so it can lead to unexpected behavior

lib/connectors/crawler/scheduler.rb
```diff
@@ -22,11 +22,48 @@ def connector_settings
...
+      end
+
       private
 
       def connector_registered?(service_type)
         service_type == 'elastic-crawler'
       end
+
+      def custom_schedule_triggered?(cs)
+        cs.custom_scheduling_settings.each do |key, custom_scheduling|
```
vidok, line 58: Could you please leave a comment that only the first found scheduling will be triggered?

lib/core/scheduler.rb
```diff
@@ -189,5 +146,67 @@ def connector_registered?(service_type)
...
       end
     end
+
+    def schedule_triggered?(scheduling_settings, identifier, last_synced)
```
vidok, line 150: It looks like `last_synced` is not needed here because it's already presented in `scheduling_settings`

lib/connectors/crawler/scheduler.rb
```diff
@@ -22,11 +22,48 @@ def connector_settings
...
+          connector_settings.each do |cs|
+            # crawler only supports :sync
+            if sync_triggered?(cs)
+              yield cs, :sync, nil
+              next
+            end
+
+            custom_schedule_triggered, schedule_key = custom_schedule_triggered?(cs)
+            if custom_schedule_triggered
+              yield cs, :sync, schedule_key
```
vidok, line 36: Do you also plan to propagate `schedule_key` to the job's payload?

lib/connectors/crawler/scheduler.rb
navarone-feekery: I wasn't replying, just pre-emptively explaining 😄 
But good to know I was on the right track!

lib/core/scheduler.rb
navarone-feekery: I can! Totally missed this, thanks

lib/core/scheduler.rb
navarone-feekery: The `last_synced` value is placed a little differently between `scheduling` and `custom_scheduling`. The `scheduling_settings` object here will be a little different depending on where it originated from.

```
# scheduling
{
  enabled: <boolean>,
  interval: <string>
}

# custom_scheduling
{
  enabled: <boolean>,
  interval: <string>,
  name: <string>,
  last_synced: <string>,
  configuration_overrides: <object>
}
```

This comment made me realise I left `last_synced` out of the `custom_scheduling` mappings update so I can see where the confusion would come from. I'll fix that in a separate PR.

lib/connectors/crawler/scheduler.rb
navarone-feekery: Do you mean to the `.elastic-connectors-sync-jobs` doc for this sync job? If so, yes.

lib/core/scheduler.rb
navarone-feekery: I've updated the connector protocol in this PR to include `custom_scheduling.*.last_synced`

lib/connectors/crawler/scheduler.rb
```diff
@@ -22,11 +22,50 @@ def connector_settings
...
+        loop do
+          connector_settings.each do |cs|
+            # crawler only supports :sync
+            if sync_triggered?(cs)
+              yield cs, :sync, nil
+              next
+            end
+
+            schedule_key = custom_schedule_triggered(cs)
+            puts('***')
```
vidok, line 35: I guess you will remove these `puts` calls. 

this reminds me of a super nice and helpful [blog post ](https://tenderlovemaking.com/2016/02/05/i-am-a-puts-debuggerer.html)

navarone-feekery: AHH I'll just go bury myself now

lib/core/scheduler.rb
```diff
@@ -195,5 +148,67 @@ def connector_registered?(service_type)
...
+      cron_parser = Fugit::Cron.parse(current_schedule)
+
+      # Don't sync if the scheduling interval is non-parsable
+      unless cron_parser
+        Utility::Logger.error("Unable to parse sync schedule for #{identifier}: expression #{current_schedule} is not a valid Quartz Cron definition.")
+        return false
+      end
+
+      # We want to sync when sync never actually happened
+      if last_synced.nil? || last_synced.empty?
```
vidok, line 183: Does this mean that when I configure a new custom schedule it will be triggered during the next service iteration because `last_syncted` is `nil`? 

if yes, I don't think customers expect this behavior.

navarone-feekery: Default schedules work this way so I don't think it's unprecedented behaviour.

If we don't trigger this when `last_synced: nil` then there's no way to determine if the next sync is ready because `last_synced` won't get generated by the first sync.
If we had a UI or creation flow we could force a default of `last_synced: Time.now` but that won't be available in 8.7.
🤔

Would you be comfortable with this if it was well-documented?

lib/core/scheduler.rb
```diff
@@ -83,51 +80,7 @@ def sync_triggered?(connector_settings)
...
-      next_trigger_time = cron_parser.next_time(Time.parse(last_synced))
-
-      # Sync if next trigger for the connector is in past
-      if next_trigger_time < Time.now
-        Utility::Logger.info("#{connector_settings.formatted.capitalize} sync is triggered by cron schedule #{current_schedule}.")
-        return true
-      end
-
-      false
+      schedule_triggered?(connector_settings.scheduling_settings, connector_settings.formatted, connector_settings.last_synced)
```
vidok, line 83: What is the fundamental difference between `scheduling_settings` and `custom_scheduling`? I have a feeling that `scheduling_settings` might be just a part of `custom_scheduling` and then we can have something like this: 

```JSON
scheduling_settings: {
  key1: { ...configuration }
  key2: { ...configuration }
}
```

It will help us to avoid having 2 similar abstractions in the protocol and in the source code.

lib/core/scheduler.rb
vidok: > Would you be comfortable with this if it was well documented?

I wouldn't :). I think this is a bad customer experience. When I schedule a job to be triggered at 6:00pm I expect that it will run at 6:00pm and not right after I clicked `Save`. 

I would discuss this question with PMs.

lib/core/scheduler.rb
navarone-feekery: Originally this was what I wanted to do (move everything to within `scheduling` and from 8.7 onwards multiple schedules would be supported). It was brought up in the implementation document that it lacks backwards compatibility. So the solution was to have a "main" scheduling (what already exists) and `custom_scheduling` would belong under its own namespace.

The fundamental difference: `custom_scheduling` is a subset of `scheduling` that runs partial jobs that override top-level settings.

There's a conversation still open on the implementation document about this right now: https://docs.google.com/document/d/1F7SCENHYKpG58ZDPZRrcrqctBJhYt8tUOAxMiI_M1sU/edit#

lib/core/scheduler.rb
navarone-feekery: @vidok thanks for bringing this up as it has sparked a lot of conversation around how connector scheduling should function if the connector has never synced before.

I've just learnt that the connectors cron logic requires a `last_synced` value in order to calculate when the next sync should run. Because of that, we need to populate `last_synced` with a value in order to trigger any schedules. Until this is resolved I think we should implement custom scheduling with the same requirements that regular scheduling has.

[Relevant connectors code](https://github.com/elastic/connectors-ruby/blob/9d5be7a79ee18c794403d14b1dcfc2c574786771/lib/core/scheduler.rb#L122)
[Slack thread](https://elastic.slack.com/archives/C01795T48LQ/p1674644501487809)

navarone-feekery: There's also an active bug issue for tracking this here: https://github.com/elastic/enterprise-search-team/issues/3366

navarone-feekery:
@vidok is there anything else in this PR that you think could be changed?

lib/core/scheduler.rb
vidok: I left a comment in the doc. Let's proceed with the current solution but could you please create an issue with the tech debt label?

vidok:
> @vidok is there anything else in this PR that you think could be changed?

I left one comment but no changes are required from my side  🚀

vidok (approved):

lib/core/scheduler.rb
navarone-feekery: @vidok good idea, I've created the issue here: https://github.com/elastic/enterprise-search-team/issues/3808

PR #515, opened 2023-01-10
Fix makefile for arm64

seanstory:
See the slack thread: https://elastic.slack.com/archives/CGRRUP4L9/p1673383147442159?thread_ts=1673380391.857499&cid=CGRRUP4L9

See the rbenv issue where this fix was found: https://github.com/rbenv/ruby-build/issues/1691#issuecomment-752802171

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -199,6 +199,7 @@ GEM
       hashdiff (>= 0.4.0, < 2.0.0)
 
 PLATFORMS
+  arm64-darwin-22
   universal-darwin-21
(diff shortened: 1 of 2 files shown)


kderusso (approved):
Thanks Sean! This runs successfully on my machine. 👍

tarekziade:
interesting, looks like a macOs only issue, the aarch64 here https://github.com/elastic/ent-search/blob/main/dist/shared/cloud/coservices/install-services.sh.erb#L56 builds fine..

PR #526, opened 2023-01-31
Add id to connector indices + dynamic: false

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/3729

Connected to issue https://github.com/elastic/enterprise-search-team/issues/3729

Updating connectors protocol to reflect the changes in the PR:

- adding "dynamic: false" to both indices

### DIFF
--- a/docs/CONNECTOR_PROTOCOL.md
+++ b/docs/CONNECTOR_PROTOCOL.md
@@ -134,7 +134,6 @@ This is our main communication index, used to communicate the connector's config
   },
   "dynamic": false,
   "properties" : {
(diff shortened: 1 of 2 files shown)


wangch079 (approved):
You may also want to add it here:

https://github.com/elastic/connectors-ruby/blob/fa054cad0a2f324e9ac27befffa28d535302180d/lib/core/elastic_connector_actions.rb#L334

and here

https://github.com/elastic/connectors-ruby/blob/fa054cad0a2f324e9ac27befffa28d535302180d/lib/core/elastic_connector_actions.rb#L455

PR #527, opened 2023-02-02
Update connector protocol

navarone-feekery:
## Related to: https://github.com/elastic/enterprise-search-team/issues/3545

- Fix whitespacing
- Fix incorrectly described `custom_scheduling.[key].configuration_overrides`, which should be an object.
- Added `preferences` to `.elastic-connectors`
  - This is a configurable field from Kibana so I've copied the structure of `configuration` so it can be used similarly in the future

#### Pre-Review Checklist

## Related Pull Requests

* Ent-search (migration): https://github.com/elastic/ent-search/pull/7292
* Kibana (index creation): https://github.com/elastic/kibana/pull/150165

### DIFF
--- a/docs/CONNECTOR_PROTOCOL.md
+++ b/docs/CONNECTOR_PROTOCOL.md
@@ -99,15 +99,10 @@ This is our main communication index, used to communicate the connector's config
     reduce_whitespace: boolean;         -> Whether the `request_pipeline` should squish redundant whitespace
     run_ml_inference: boolean;          -> Whether the `request_pipeline` should run the ML Inference pipeline
   };
-  preferences: {
-    [key]: {
-      label: string   -> The label to be displayed for the field in Kibana
-      value: string,  -> The value of the field configured in Kibana
-    }
-  };                  -> Dynamic preferences for the connector
+  preferences: object;  -> Contains dynamic field:value pairs for preferences for the connector
   scheduling: {
-    enabled: boolean; -> Whether job schedule is enabled
-    interval: string; -> Quartz Cron syntax
+    enabled: boolean;   -> Whether job schedule is enabled
+    interval: string;   -> Quartz Cron syntax
   };
   service_type: string; -> Service type of the connector
   status: string;       -> Connector status Enum, see below
--- a/docs/CONNECTOR_PROTOCOL.md
+++ b/docs/CONNECTOR_PROTOCOL.md
@@ -99,7 +99,9 @@ This is our main communication index, used to communicate the connector's config
     reduce_whitespace: boolean;         -> Whether the `request_pipeline` should squish redundant whitespace
     run_ml_inference: boolean;          -> Whether the `request_pipeline` should run the ML Inference pipeline
   };
-  preferences: object;  -> Contains dynamic field:value pairs for preferences for the connector
+  preferences: {
+    extract_full_html: boolean; -> Whether or not to fully extract HTML (Crawler)
+  };                            -> Contains dynamic field:value pairs for preferences for the connector
   scheduling: {
     enabled: boolean;   -> Whether job schedule is enabled
     interval: string;   -> Quartz Cron syntax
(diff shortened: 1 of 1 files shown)


navarone-feekery:
Removing the `label/value` sections as per feedback on the kibana PR: https://github.com/elastic/kibana/pull/150165#issuecomment-1413739817

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -97,16 +92,19 @@ This is our main communication index, used to communicate the connector's config
...
-    run_ml_inference: boolean; -> Whether the `request_pipeline` should run the ML Inference pipeline
-  }
+    extract_binary_content: boolean;    -> Whether the `request_pipeline` should handle binary data
+    name: string;                       ->  Ingest pipeline to utilize on indexing data to Elasticsearch
+    reduce_whitespace: boolean;         -> Whether the `request_pipeline` should squish redundant whitespace
+    run_ml_inference: boolean;          -> Whether the `request_pipeline` should run the ML Inference pipeline
+  };
+  preferences: {
+    extract_full_html: boolean; -> Whether or not to fully extract HTML (Crawler)
+  };                            -> Contains dynamic field:value pairs for preferences for the connector
```
artem-shelkovnikov, line 104: I think I've missed the discussion - why cannot it be a configurable field?

docs/CONNECTOR_PROTOCOL.md
```diff
@@ -235,6 +233,11 @@ This is our main communication index, used to communicate the connector's config
         "run_ml_inference" : { "type" : "boolean" }
       }
     },
+    "preferences" : {
+      "properties" : {
+        "extract_full_html" : { "type" : "boolean" }pr
```
artem-shelkovnikov, line 238:
```suggestion
        "extract_full_html" : { "type" : "boolean" }
```

docs/CONNECTOR_PROTOCOL.md
navarone-feekery: @artem-shelkovnikov I think the big difference is a configuration is necessary to run, but this is just an opt-in to fetch a specific data type.

@sphilipse I remember discussing this with you but my notes just say "don't use configuration" -- do you remember why we decided that?

artem-shelkovnikov: Since you initialise and use the configuration, you can do the optional/required on your side of course, I just feel that having two objects that represent same concepts (e.g. alteration of behaviour of individual connector) might bite us back

sphilipse: Because configuration fields show up in the UI, and are meant for service-type specific user-set configurations (like, say, username and password). This setting is independent from the service type, and shows up in a completely different part of the UI.

sphilipse: Actually not independent from the service type, I guess, 'cause this'll only be used by crawlers. But still: as a type of configuration this shows up in a very different place, and conceptually is different from the configuration fields (which are more "how do I connect to my data source").

sphilipse: The counterpoint to this would be that we don't actually use the configuration blob for the crawler, and the crawler already has a bunch of bespoke UIs, so maybe it's okay to put it in there anyway 🤔

navarone-feekery: We do actually use the configuration blob for crawler. We add `{ "nextSyncConfig": { "label": "nextSyncConfig", "value": null } }` after running a crawl. In theory you could add configuration for the next crawl in place of `null` but we aren't using it in practice, it's a holdover from when we were trying to make crawler a connector.

Which leads me to a thought -- we don't really use `configuration` for crawler now, but if it became a connector in the future, we would. I also think there may be use cases for `preferences` on other connectors. For example if your DB has JSON objects in it, whether to store that in a document as JSON or string.

artem-shelkovnikov: > The counterpoint to this would be that we don't actually use the configuration blob for the crawler, and the crawler already has a bunch of bespoke UIs, so maybe it's okay to put it in there anyway 🤔

That was my feeling - we already have configurable_fields + features on the connectors side, but crawler is a very special connector with special UI that's not supported and won't be probably supported in generic connectors. I'm feeling like we should be really careful in this particular case.

navarone-feekery: > won't be probably supported in generic connectors

This is true, Crawler will always be an awkward half-connector. It invalidates my point.
If we don't have a need for `preferences` for other connectors then maybe `configuration` is the way to go.

navarone-feekery:
We jumped on a call and decided that using `configuration` was preferable for this.

PR #528, opened 2023-02-02
Don't delete content indices used by existing connectors

wangch079:
This PR ensures that the content indices of existing connectors are not deleted, when cleaning up orphaned jobs.

#### Pre-Review Checklist
- [x] Considered corresponding documentation changes
- [x] Contributed any configuration settings changes to the configuration reference

### DIFF
--- a/lib/core/jobs/consumer.rb
+++ b/lib/core/jobs/consumer.rb
@@ -37,7 +37,7 @@ module Core
       end
 
       def subscribe!(index_name:)
-        Utility::Logger.info("Starting a new consumer for #{@index_name} index")
+        Utility::Logger.info("Starting a new consumer for #{index_name} index")
 
         @index_name = index_name
         start_timer_task!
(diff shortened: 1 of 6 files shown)


lib/core/jobs/consumer.rb
```diff
@@ -37,7 +37,7 @@ def initialize(scheduler:,
       end
 
       def subscribe!(index_name:)
-        Utility::Logger.info("Starting a new consumer for #{@index_name} index")
+        Utility::Logger.info("Starting a new consumer for #{index_name} index")
 
         @index_name = index_name
```
navarone-feekery, line 42: Is it okay to call `@index_name` before it has been assigned the value from the method argument? At a glance it looks like the memoised `@index_name` could be `nil` or contain an old value.

navarone-feekery: AH I was reading the diff wrong, maybe I'm colourblind :facepalm:

navarone-feekery (approved):
LGTM

lib/core/jobs/consumer.rb
wangch079: It won't complain but it does print empty string, which I don't think is what it intend to do.

PR #529, opened 2023-02-02
Reset last_sync_error when claiming a job

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/3384

This PR resets `last_sync_error`, and update `status` to `connected` when claiming a job

#### Pre-Review Checklist
- [x] Considered corresponding documentation changes
- [x] Contributed any configuration settings changes to the configuration reference

### DIFF
--- a/lib/core/elastic_connector_actions.rb
+++ b/lib/core/elastic_connector_actions.rb
@@ -157,12 +157,18 @@ module Core
         )
       end
 
-      def update_connector_last_sync_status(connector_id, last_sync_status)
+      def update_connector_sync_start(connector_id)
         doc = connector_with_concurrency_control(connector_id)
+
+        body = {
+            last_sync_status: Connectors::SyncStatus::IN_PROGRESS,
+            last_sync_error: nil,
+            status: Connectors::ConnectorStatus::CONNECTED
+        }
 
         update_connector_fields(
           connector_id,
-          { last_sync_status: last_sync_status },
+          body,
           doc[:seq_no],
           doc[:primary_term]
         )
--- a/lib/core/sync_job_runner.rb
+++ b/lib/core/sync_job_runner.rb
@@ -202,7 +202,7 @@ module Core
       end
 
       begin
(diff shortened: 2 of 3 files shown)


lib/core/elastic_connector_actions.rb
```diff
@@ -157,12 +157,18 @@ def update_connector_sync_now(connector_id, sync_now)
         )
       end
 
-      def update_connector_last_sync_status(connector_id, last_sync_status)
+      def update_connector_sync_start(connector_id)
```
navarone-feekery, line 160: Note: changing the name of this method will break Crawler if/when we update the utility gem (specifically [this line](https://github.com/elastic/ent-search/blob/ab343921d317d4c8857f392ab712108dd7ef41af/shared_togo/lib/shared_togo/crawler2/scheduler.rb#L60)). Once this is merged we should make a PR in ent-search to make sure it doesn't surprise anyone later.

navarone-feekery (approved):
LGTM, just a comment about impact on ent-search

lib/core/elastic_connector_actions.rb
```diff
@@ -157,12 +157,18 @@ def update_connector_sync_now(connector_id, sync_now)
...
         doc = connector_with_concurrency_control(connector_id)
 
+        body = {
+            last_sync_status: Connectors::SyncStatus::IN_PROGRESS,
+            last_sync_error: nil,
+            status: Connectors::ConnectorStatus::CONNECTED
```
artem-shelkovnikov, line 166: Why update status here?

lib/core/elastic_connector_actions.rb
wangch079: Good catch. Will do.

lib/core/elastic_connector_actions.rb
wangch079: A new job gets claimed and starts execution. we should mark the job as `connected`, otherwise it may stay `error` until job completes.

artem-shelkovnikov: Ah I see, overlooked it.

PR #530, opened 2023-02-08
Implement nested features

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/3653

This PR implements new connector features structure:

```
{
"features": {
  "sync_rules": {
    "basic" : { "enabled" : true },
    "advanced": {"enabled":true }
  }
}
```

While this structure is also implemented and used, old structure is still maintained and used for backwards compatibility.

#### Pre-Review Checklist

### DIFF
--- a/docs/CONNECTOR_PROTOCOL.md
+++ b/docs/CONNECTOR_PROTOCOL.md
@@ -55,8 +55,11 @@ This is our main communication index, used to communicate the connector's config
   description: string;  -> The description of the connector
   error: string;        -> Optional error message
   features: {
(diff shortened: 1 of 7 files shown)


seanstory (approved):

seanstory:
Also needs to update the CONNECTOR_PROTOCOL.md file with this new structure

artem-shelkovnikov:
Good point @seanstory, updated!

I'm not sure about naming though - I've called nested keys "feature"."subfeature", but is this naming actually correct? cc @timgrein @sphilipse

seanstory:
I think that works just fine. We can update it later if there's confusion.

PR #553, opened 2023-05-30
Modify the structure of scheduling

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/4628

This PR modifies the structure of `scheduling`, to accommodate different types of scheduling.

#### Pre-Review Checklist

### DIFF
--- a/lib/app/console_app.rb
+++ b/lib/app/console_app.rb
@@ -89,7 +89,7 @@ module App
     def enable_scheduling
       return unless connector_registered?
 
-      previous_schedule = Core::ConnectorSettings.fetch_by_id(connector_id)&.scheduling_settings&.fetch(:interval, nil)
+      previous_schedule = Core::ConnectorSettings.fetch_by_id(connector_id)&.full_sync_scheduling&.fetch(:interval, nil)
       if previous_schedule.present?
         puts "Please enter a valid crontab expression for scheduling. Previous schedule was: #{previous_schedule}."
       else
--- a/lib/core/connector_settings.rb
+++ b/lib/core/connector_settings.rb
@@ -103,7 +103,11 @@ module Core
     end
 
     def scheduling_settings
-      self[:scheduling]
+      self[:scheduling] || {}
+    end
+
+    def full_sync_scheduling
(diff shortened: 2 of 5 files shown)


artem-shelkovnikov (approved):
LGTM - might be worth to add logic that also checks if any other type of schedule is there and shows error/warning that schedule other than full is not supported

wangch079:
> LGTM - might be worth to add logic that also checks if any other type of schedule is there and shows error/warning that schedule other than full is not supported

We don't have this kind of validation in python either. We can't prevent users from adding something funny as we don't enforce strict mapping, but I think it's fine because we only care about `full`, `incremental` and `permissions`.

artem-shelkovnikov:
> We don't have this kind of validation in python either. We can't prevent users from adding something funny as we don't enforce strict mapping, but I think it's fine because we only care about full, incremental and permissions.

What I mean is that Ruby won't support even incremental/permissions, so if someone decides to use MongoDB in Python and then switches to Ruby it'll be a legit case that Ruby does not support and might be worth to communicate to user.

wangch079:
>so if someone decides to use MongoDB in Python and then switches to Ruby it'll be a legit case that Ruby does not support and might be worth to communicate to user.

🤔 Mmm... is it even a valid use case? we are not supporting Ruby connectors anymore. We made this PR to support Crawlers, not connectors.

artem-shelkovnikov:
> 🤔 Mmm... is it even a valid use case? we are not supporting Ruby connectors anymore. We made this PR to support Crawlers, not connectors.

Not sure, just wanted to double check with you how you feel about it

wangch079:
>Not sure, just wanted to double check with you how you feel about it

I think the only reason we make some changes to `connectors-ruby`, is either 1) there's severe vulnerability or 2) there's breaking changes to crawler(I'm not even sure about point 1).

If the scheduling structure change doesn't affect crawler, I don't think I will make this PR, even though it means `connectors-ruby` won't work with Elasticsearch and Enterprise Search 8.9. Because we are not releasing any new versions (surprisingly to see we still have 8.8 branch)

PR #559, opened 2023-07-17
Ignore 404 for meta pipeline values

sphilipse:
This stops connectors from throwing when the connectors index hasn't been created but we need to know the default pipeline settings.


VERSION
```diff
@@ -1 +1 @@
-8.10.0.0
+8.10.0.0.pre.dev
```
seanstory, line 1: the `pre.dev` comes from running `make release`. Feel free to bump this to 8.10.0.1, release this change (for ent-search) and then bump to 8.10.0.2 for `main` (SNAPSHOT) is inferred.

sphilipse: done

seanstory (approved):

PR #563, opened 2023-08-15
Bump rack and activesupport

navarone-feekery:
Resolves dependabot alerts: https://github.com/elastic/connectors-ruby/security/dependabot


vidok (approved):

navarone-feekery:
@vidok this needs re-review as I had to fix the Gemfile.lock file

vidok (approved):

PR #564, opened 2023-08-16
Bump stack versions

navarone-feekery:
Related to [8.10 RM tasks](https://github.com/elastic/enterprise-search-team/issues/5427)

Bump stack versions required after cutting 8.10 branch.


.backportrc.json
```diff
@@ -1,13 +1,14 @@
 {
   "targetBranchChoices": [
     { "name": "main", "checked": true },
+    "8.10",
     "8.9",
     "8.8",
```
wangch079, line 6: As there won't be any minor release for 8.8, I think we can remove this line. Be careful of the trailing comma as well.

navarone-feekery: Done!

wangch079 (approved):

PR #580, opened 2023-09-21
Update VERSION for 8.10.2 release

seanstory:


navarone-feekery:
@seanstory as we are no longer releasing connectors ruby, should we close this PR?

seanstory:
yeah good point

PR #591, opened 2024-12-04
Upgrade fugit and mime-types

wangch079:
This PR upgrades dependency `fugit` and `mime-types`


wangch079:
I faced a similar issue as [this one](https://github.com/ruby-debug/debase/issues/92#issuecomment-2013704819) when I ran

```bash
make install
```

and I get it pass by running

```bash
gem install debase -v0.2.5.beta2 -- --with-cflags="-Wno-incompatible-function-pointer-types"
```

I'm not planning to invest more time in it.

artem-shelkovnikov (approved):

PR #597, opened 2025-01-07
Fix for specific time scheduled crawls skipping if too many crawlers

mattnowzari:
## Closes https://github.com/elastic/search-team/issues/8926

A fix for specific time scheduled crawls being skipped if too many crawlers exist.

Great explanation by @navarone-feekery [here](https://github.com/elastic/search-team/issues/8926) that includes an RCA, how to reproduce, and three solutions to tackle this problem, two of which are addressed in this PR.

> 2. The Time.now [here](https://github.com/elastic/connectors-ruby/blob/7fef71378f766cf792b7b9445f87a8f9b6eb32ca/lib/core/scheduler.rb#L182) should probably be set at the beginning of the iteration over connector settings
> 3. We should add debug logging to connectors-ruby [here](https://github.com/elastic/connectors-ruby/blob/7fef71378f766cf792b7b9445f87a8f9b6eb32ca/lib/core/scheduler.rb#L189-L190) for when a schedule is skipped because it was deemed not available to sync. Having this debug log would have saved me a few hours in SDH.

#### Pre-Review Checklist
- [ ] Considered corresponding documentation changes
- [ ] Contributed any configuration settings changes to the configuration reference


VERSION
```diff
@@ -1 +1 @@
-8.16.0.0
+8.16.0.1
```
navarone-feekery, line 1: Was this auto incremented? If not, I don't think it's necessary to bump as we manage this when doing a release

mattnowzari: This was auto-incremented I believe, I'll manually bump it back down

spec/connectors/crawler/crawler_scheduler_spec.rb
```diff
@@ -88,14 +89,15 @@
       let(:weekly_next_trigger_time) { 1.day.from_now }
       let(:monthly_next_trigger_time) { 1.day.from_now }
 
+      let(:time_at_poll_start) { Timecop.freeze(Time.now) }
```
navarone-feekery, line 92: I think having this initialized outside of any context blocks means all tests will share the same start time, which means that if there's some disruption to the test it's possible some may fall outside of the expected poll interval, so there's risk the tests will be flaky. I think this would be safer to initialize inside each context block (or in a before block, or anything else that works)

lib/core/scheduler.rb
```diff
@@ -179,12 +179,15 @@ def schedule_triggered?(scheduling_settings, identifier)
...
+      next_trigger_time = cron_parser.next_time(time_at_poll_start)
       # Sync if next trigger happens before the next poll
-      if next_trigger_time <= Time.now + @poll_interval
+      poll_window = time_at_poll_start + @poll_interval
+      if next_trigger_time <= poll_window
         Utility::Logger.info("#{identifier.capitalize} sync is triggered by cron schedule #{current_schedule}.")
         return true
+      else
+        # log that a sync was not triggered, share the next trigger time and when poll interval was meant to end
+        Utility::Logger.debug("Sync for #{identifier.capitalize} not triggered, as #{next_trigger_time} occurs after the current interval: #{poll_window}.")
```
navarone-feekery, line 190: Nit/discussion: I wonder which would be more valuable to the user, the `poll_window` value, or a description of the `time_at_poll_start` and their configured `poll_interval` 🤔 
Maybe all of this information can be included, if the log isn't getting too long. What are your thoughts?

```suggestion
        Utility::Logger.debug("Sync for #{identifier.capitalize} not triggered, as #{next_trigger_time} occurs after the current poll window: #{poll_window}. Poll window began at #{time_at_poll_start}, poll interval is #{@poll_interval}" seconds.)
```

navarone-feekery (commented):
Looking good! Just left a few possible test cases that could be covered.

spec/connectors/crawler/crawler_scheduler_spec.rb
```diff
@@ -156,6 +162,54 @@
...
+        let(:next_trigger_time) { time_at_poll_start + poll_interval }
+        let(:weekly_next_trigger_time) { time_at_poll_start + poll_interval }
+        let(:monthly_next_trigger_time) { time_at_poll_start + poll_interval }
+
+        before(:each) do
+          allow(cron_parser).to receive(:next_time).with(time_at_poll_start).and_return(next_trigger_time, weekly_next_trigger_time, monthly_next_trigger_time)
+        end
+
+        it_behaves_like 'triggers', nil
+      end
```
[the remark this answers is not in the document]
mattnowzari: I hadn't thought of these, I agree these are good tests to add as well, will address these in the next commit!

spec/connectors/crawler/crawler_scheduler_spec.rb
mattnowzari: What's funny is that I actually had them initialized in each context block initially, then thought "nah that isn't DRY enough" without thinking of the potential consequences! Will address this as well.

lib/core/scheduler.rb
mattnowzari: That's a really good idea - I agree adding the value of `poll_interval` makes a lot of sense here as well. I'll play around with the verbiage to see how we can incorporate the info we want in a concise manner.

If it gets too wordy I could possibly break this out into 2 debug lines maybe?

seanstory:
Nit: your PR title is `Closes <issue>` but there's going to need to be a follow-up ent-search PR to expose this setting in enterprise-search.yml, right? So you probably don't want the issue to close as soon as this PR merges. Instead, you may want to use `Related to <issue>` or `Part of <issue>` instead

spec/connectors/crawler/crawler_scheduler_spec.rb
seanstory: I _think_ that, since rspec `lets` are lazy, this actually isn't initialized until inside a block when first referenced?

navarone-feekery: I might be wrong, but I thought the let is indeed lazy, but once it's called its retained for its entire scope. In this case that scope is shared across several tests, so I _think_ it's the same value for all of them.

mattnowzari: To Nav's point - everything I am reading about `let` indicates they are lazily evaluated, however it'd mean that it would get set on whatever `Timecop.freeze(Time.now) `was at its first evaluation, I _think_

mattnowzari: I dug deeper into this, and it turns out that the way it is currently set (with the singlular `let(:time_at_poll_interval)` at the top-level context block), it will get evaluated every time the `before(:each) do` block executes, because :time_at_poll_start is being evaluated on lines 99 and 100 below.
Thus, each test starts with a newly-evaluated Time.now value.

Here are the results of ye olde "add print statements everywhere and see whats goin on" experiment (I cleaned up some unrelated output for clarity). I added a `sleep 2` to enforce a noticeable passage of time between tests

```
time_at_poll_start in  before(:each) do  block: 2025-01-09 10:52:53 -0500
time_at_poll_start inside method schedule_triggered?(): 2025-01-09 10:52:53 -0500

time_at_poll_start in  before(:each) do  block: 2025-01-09 10:52:55 -0500
time_at_poll_start inside method schedule_triggered?(): 2025-01-09 10:52:55 -0500

time_at_poll_start in  before(:each) do  block: 2025-01-09 10:52:57 -0500
time_at_poll_start inside method schedule_triggered?(): 2025-01-09 10:52:57 -0500

time_at_poll_start in  before(:each) do  block: 2025-01-09 10:52:59 -0500
time_at_poll_start inside method schedule_triggered?(): 2025-01-09 10:52:59 -0500
```

All this to say, I still don't mind having `let(:time_at_poll_start)` being defined within each context block, as it might mean less mental overhead when some future person reads these tests. Curious to know your guys' thoughts?

seanstory: I think it's good as is, personally. It is more DRY ("ruby-onic"?) :)

lib/core/scheduler.rb
mattnowzari: So this morning when I responded to this comment my brain didn't compute the fact that you had already suggested a change @navarone-feekery  - sorry for the brainfart! I've incorporated your wording as it's far better than what I was going to come up with 🤣

navarone-feekery: I don't think anyone is picky about whether or not you click "accept changes" or you manually add the changes. I'm not chasing github points or anything, it's just a nice format to suggest a change. How the change goes in is inconsequential, so it's not something to worry about :)

spec/connectors/crawler/crawler_scheduler_spec.rb
navarone-feekery: TIL, thanks for investigating this!

spec/connectors/crawler/crawler_scheduler_spec.rb
```diff
@@ -156,6 +162,70 @@
...
+
+        let(:next_trigger_time) { time_at_poll_start + poll_interval + 10 }
+        let(:weekly_next_trigger_time) { time_at_poll_start + poll_interval - 10 }
+
+        before(:each) do
+          allow(cron_parser).to receive(:next_time).with(time_at_poll_start).and_return(next_trigger_time, weekly_next_trigger_time)
+          expect(Utility::Logger).to receive(:debug).exactly(1).times.with(instance_of(String))
+        end
+
+        it_behaves_like 'triggers', :weekly_key
```
navarone-feekery, line 195: Just an rspec point/nit: I don't think an `expect` should be in a `before` block if we're following principles of [four-phase testing](https://thoughtbot.com/blog/four-phase-test). The `before` block is part of the test's setup phase. We want to use `expect` during the verify stage, which is within the `it` block.

I can see how it looks awkward to add more work to the `it` block when this spec is using shared examples with `it_behaves_like`. It's possible to wrap the shared example in another `it` block. This will execute everything within as if it was one:

```suggestion
        before(:each) do
          allow(cron_parser).to receive(:next_time).with(time_at_poll_start).and_return(next_trigger_time, weekly_next_trigger_time)
        end

        it 'triggers the first schedule and skips the second' do
            expect(Utility::Logger).to receive(:debug).exactly(1).times.with(instance_of(String))
            it_behaves_like 'triggers', :weekly_key
        end

```

spec/connectors/crawler/crawler_scheduler_spec.rb
```diff
@@ -156,6 +162,70 @@
...
+        let(:weekly_enabled) { true }
+        let(:monthly_enabled) { true }
+
+        let(:next_trigger_time) { time_at_poll_start + poll_interval + 10 }
+        let(:weekly_next_trigger_time) { time_at_poll_start + poll_interval + 10 }
+        let(:monthly_next_trigger_time) { time_at_poll_start + poll_interval + 10 }
+
+        before(:each) do
+          allow(cron_parser).to receive(:next_time).with(time_at_poll_start).and_return(next_trigger_time, weekly_next_trigger_time, monthly_next_trigger_time)
+          expect(Utility::Logger).to receive(:debug).exactly(3).times.with(instance_of(String)) # expect three debug messages because three schedules are not being triggered
```
navarone-feekery, line 177: I think it's more important to verify that the debug log we receive is the expected one. Because there's three we can use match and a simple regex, e.g. `match(/some_pattern/)`. I don't think we need the full string matched, just the beginning or end so we're sure it's the expected log.

Also (total nit) if the comment goes above the line it's much shorter and will be kind to IDEs.

```suggestion
          # expect three debug messages because three schedules are not being triggered
          expect(Utility::Logger).to receive(:debug).exactly(3).times.with(match(/^Sync for \w+ not triggered as .*/))
```

navarone-feekery (commented):
Thanks for implementing the features! I just have some ruby-specific comments, but the implementation looks really good :D

spec/connectors/crawler/crawler_scheduler_spec.rb
mattnowzari: Good callout! I had to add a tiny bit more to the regex (in the form of an OR) to get it to match on slight variation in the `identifier` name the test kicks out, but otherwise 100% agree here.

`^Sync for (\w+.*)|( - \w+) not triggered as .*`

```
Sync for Formatted not triggered as 2025-01-10 09:39:44....etc
Sync for Formatted - weekly not triggered as 2025-01-10 09:39:44...etc
```

spec/connectors/crawler/crawler_scheduler_spec.rb
mattnowzari: So I'm running into a bit of a road block with this one - when I attempt to put an 'it_behaves_like' inside of an 'it' block, it gives me the following:

`RSpec::Core::ExampleGroup::WrongScopeError: `it_behaves_like` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).`

I've tried the inverse of wrapping an `it`block with an `it_behaves_like do` block, but then the test doesn't behave as expected (expect won't receive a debug message)

Logically what you've suggested makes sense to me, but RSpec does not seem too happy. Thoughts?

navarone-feekery (approved):
:shipit:

